PR #1, opened 2019-08-12
Add a basic fallback to disputable assertions when unanimous assertions fail

hkalodner:


hkalodner:
Changed have been approved internally

PR #18, opened 2019-08-23
Added error handling for invalid machine file.

tpobryan:


packages/arb-validator/testmachine/machine.go
```diff
@@ -36,11 +36,13 @@ type Machine struct {
...
-	gm, err := goloader.LoadMachineFromFile(codeFile, warnMode)
+	gm, gmerr := goloader.LoadMachineFromFile(codeFile, warnMode)
+	cm, cmerr := cmachine.New(codeFile)
+
 	return &Machine{
-		cmachine.New(codeFile),
+		cm,
 		gm,
-	}, err
+	}, fmt.Errorf("Go machine error: %v \ncpp machine error: %v", gmerr, cmerr)
```
hkalodner: This will cause an error to always be returned which will cause this to fail. Instead there should only be an error if `gmerr` or `cmerr` are not nil

PR #71, opened 2019-09-10
Arb test

tpobryan:


packages/arb-provider-go/test/connection_test.go
```diff
@@ -0,0 +1,280 @@
...
+func TestFib(t *testing.T) {
+	session, err := RunValidators(t)
+	if err != nil {
+		t.Errorf("Validator setup error %v", err)
+	}
+
+	t.Run("TestFibResult", func(t *testing.T) {
+		fibsize := 15
+		fibnum := 11
+		fmt.Println("generating fib")
```
hkalodner: Use t.Log or t.Logf instead of fmt.Print or log.Print variants when inside tests.

packages/arb-provider-go/test/connection_test.go
```diff
@@ -0,0 +1,280 @@
...
+			t.Errorf("GetFib error - expected %v got %v", big.NewInt(int64(144)), fibval)
+		}
+		log.Printf("Fibonacci value number %v = %v", fibnum, fibval)
+
+	})
+	t.Run("TestEvent", func(t *testing.T) {
+		eventChan := StartEventListeners(session.Contract)
+
+		fibsize := 15
+		go func() {
```
hkalodner: I don't think this should need to be in a different thread.

packages/arb-provider-go/test/connection_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+		fmt.Println("waiting for event")
+	Loop:
+		for ev := range eventChan {
+			switch event := ev.(type) {
+			case *FibonacciTestEvent:
+				fmt.Printf("Received fibonacci test event %v", event.Number)
+				break Loop
+			case ListenerError:
+				t.Errorf("errorEvent %v %v", event.ListenerName, event.Err)
```
hkalodner: This error and the default error should break out of the loop along with causing the error or else there's a chance of infinite loop.

packages/arb-provider-go/test/eventListeners.go
```diff
@@ -0,0 +1,68 @@
...
+
+	startGenerateFibListener(fibonacci, ch)
+	//startSubmitYourDeckHashListener(poker, ch)
+	//startEncryptAndProveCommandListener(poker, ch)
+	//startDealFaceDownCardListener(poker, ch)
+	//startDealFaceUpCardListener(poker, ch)
+	//startCallForBetActionListener(poker, ch)
+	//startPresentBestHandListener(poker, ch)
+	//startDeclareWinningsListener(poker, ch)
+	//startReportZKPListener(poker, ch)
```
hkalodner: This comments should be removed and startGenerateFibListener should probably just be inlined since there's only one function. It might be even easier to just kill this file entirely and move it into the test since there's not much code here.

packages/arb-provider-go/test/eventListeners.go
```diff
@@ -0,0 +1,68 @@
...
+		evCh := make(chan *FibonacciTestEvent, 2)
+		start := uint64(0)
+		watch := &bind.WatchOpts{
+			Context: context.Background(),
+			Start:   &start,
+		}
+		sub, err := fibonacci.WatchTestEvent(watch, evCh)
+		if err != nil {
+			log.Println(err)
+			panic("listener startup error")
```
hkalodner: This should lead to an early return and a test error, not a panic

packages/arb-test/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
hkalodner: This and the rest of the packages/arb-test files should be removed. The code necessary for Fibonacci is currently just in https://github.com/OffchainLabs/demo-app-fibonacci and could be moved to arbitrum/demos if we want to. It shouldn't be left as a top level packages.

packages/arb-validator/coordinator/rpcserver.go
```diff
@@ -40,7 +41,12 @@ func NewRPCServer(
 	connectionInfo ethbridge.ArbAddresses,
 	ethURL string,
 ) *RPCServer {
-	return &RPCServer{NewServer(machine, key, validators, connectionInfo, ethURL)}
+	return StartRPCServerVM(NewCoordinator(machine, key, validators, connectionInfo, ethURL))
```
hkalodner: Can you explain the change from `NewServer()` to `StartRPCServerVM(NewCoordinator())`?

packages/arb-validator/coordinator/server.go
```diff
@@ -57,6 +57,17 @@ func NewServer(
 	connectionInfo ethbridge.ArbAddresses,
 	ethURL string,
 ) *Server {
+	return CreateVM(NewCoordinator(machine, key, validators, connectionInfo, ethURL))
```
hkalodner: I'm on board with refactoring this chunks since it was a lot of code doing pretty different things, but do these new functions have to be public?

packages/arb-provider-go/test/connection_test.go
```diff
@@ -0,0 +1,280 @@
...
+func setupValidators(coordinatorKey string, followerKey string, t *testing.T) error {
+
+	seed := time.Now().UnixNano()
+	// seed := int64(1559616168133477000)
+	fmt.Println("seed", seed)
+	brand.Seed(seed)
+
+	jsonFile, err := os.Open("bridge_eth_addresses.json")
+	if err != nil {
+		t.Errorf("bridge_eth_addresses error %v", err)
```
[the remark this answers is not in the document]
tpobryan: Agree and will change this instance. The question remains how to handle the error condition from ListenAndServe (line 153). As it is a separate thread, it cannot simply return an error thus we need to set the test error. Is this an acceptable inconsistency or should we convert all error conditions to set the test error?

packages/arb-validator/cmd/coordinatorServer/main.go
```diff
@@ -133,6 +133,7 @@ func main() {
 
 	// Validator creation
 	server := coordinator.NewRPCServer(machine, key, validators, connectionInfo, ethURL)
+	coordinator.StartRPCServerVM(server)
```
hkalodner: I think this would be better as `server.CreateVM()`

PR #80, opened 2019-10-05
Implement lazy hashing in AVM

ewfelten:
Rather than computing the hash of a tuple when we create (or modify) the tuple, just set a flag saying that the tuple's hash hasn't been computed yet. The hash gets computed only when it is needed.

This change is made is both the CPP and Go versions of the AVM.

This will be faster than the previous code, because some tuples will be discarded before their hash is ever computed--so we will compute fewer hashes overall.


hkalodner:
Closing since this is superseded by #131

PR #82, opened 2019-10-17
Challenge test fixed

tpobryan:
Refactoring to get arb-validator challenge test to work with automated testing.


packages/arb-validator/test/challenge_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/offchainlabs/arbitrum/packages/arb-util/value"
+	"github.com/offchainlabs/arbitrum/packages/arb-validator/ethvalidator"
+	"github.com/offchainlabs/arbitrum/packages/arb-validator/loader"
+	"github.com/offchainlabs/arbitrum/packages/arb-validator/valmessage"
+)
+
+func TestChallenge(t *testing.T) {
+
+	bridge_eth_addresses := "../../arb-provider-go/test/bridge_eth_addresses.json"
+	contract := "../../arb-provider-go/test/contract.ao"
```
hkalodner: arb-validator shouldn't have a dependence on arb-provider-go. For now it should just have a separate copy of `bridge_eth_addresses.json` and `contract.ao`

packages/arb-validator/cmd/challengeTest/main.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
```
hkalodner: This file should be removed now that we have an actual test

packages/arb-validator/test/challenge_test.go
```diff
@@ -0,0 +1,250 @@
...
+			} else {
+				log.Println("****************************")
+				log.Println("recoverable error continuing")
+				log.Println(message)
+				log.Println("****************************")
+			}
+		case message := <-challengerErrMonitorChan:
+			if !message.Recoverable {
+				t.Error("challenger unexpected unrecoverable error")
+				log.Println("***************************")
```
hkalodner: Use `t.Log()` instead of `log.Println()` inside tests

packages/arb-validator/test/challenge_test.go
```diff
@@ -0,0 +1,250 @@
...
+	receipt, err := coordinator.Val.SendEthMessage(
+		context.Background(),
+		tup,
+		big.NewInt(0),
+	)
+	if err != nil {
+		log.Fatalln("Send error", err)
+	}
+	if receipt.Status == 0 {
+		log.Fatalln("Follower could not send message")
```
hkalodner: Use `t.Fatal()` instead of `log.Fatalln()` inside tests

packages/arb-validator/validator/validator.go
```diff
@@ -327,14 +327,19 @@ func (validator *Validator) RequestDisputableAssertion(length uint64) (<-chan bo
 	return resultChan, errChan
 }
 
+func (validator *Validator) ValidatorClosing() {
```
hkalodner: This should be private

packages/arb-validator/validator/validator.go
```diff
@@ -364,12 +382,25 @@ func (validator *Validator) Run(ctx context.Context, recvChan <-chan ethbridge.N
 			case ethbridge.VMEvent:
 				err := validator.eventUpdate(ev, notification.Header)
 				if err != nil {
-					log.Println("Error processing event update", err)
+					//log.Printf("*****Validator %v: error - %v\n", validator.Name, err)
+					if errstat, ok := err.(*bridge.Error); ok {
```
hkalodner: If this is the only type of error that can be returned from `eventUpdate` it'd probably be better to change the signature and avoid this cast

packages/arb-validator/validator/validator.go
```diff
@@ -345,7 +350,20 @@ func (validator *Validator) Run(ctx context.Context, recvChan <-chan ethbridge.N
 				validator.latestHeader = newHeader
 				err := validator.timeUpdate()
 				if err != nil {
-					log.Println("Error processing time update", err)
+					//log.Printf("Validator %v: Error processing time update - %v\n", validator.Name, err)
+					if errstat, ok := err.(*bridge.Error); ok {
```
hkalodner: If this is the only type of error that can be returned from `timeUpdate` it'd probably be better to change the signature and avoid this cast

tpobryan:
Requested changes have been addressed. The latest commit fixes issue #83 giving slightly more urgency to merging this.

PR #86, opened 2019-10-29
fix cpp opcode 'isZero' number of parameters and bug where assertion …

tpobryan:
fix cpp opcode 'isZero' number of parameters and bug where assertion step count was incremented when machine was blocked - both cpp and go AVMs.


hkalodner:
This fixes #87 and #88

Issue #90, opened 2019-11-01
OneStepProof error in tget 


tpobryan:
OneStepProof.sol - doesn't catch index out of bounds if index = tuple size

Issue #89, opened 2019-11-01
arb-avm-cpp stack cleanup on error incorrect


tpobryan:
When error condition hit in opcode processing, stack is not cleaned up properly before jumping to error handler.

PR #91, opened 2019-11-01
fixed issues 89 and 90

tpobryan:


packages/arb-avm-cpp/avm/src/machine.cpp
```diff
@@ -324,6 +322,8 @@ void Machine::runOne() {
         }
     }
 
+    auto startStackSize = m.stack.stacksize();
```
hkalodner: This is the size after the instruction has run which is incorrect. It should be the size before the instruction has been run, but after the immediate value (if it exists) has been added to the stack.

hkalodner:
This fixes #89, fixes #90

PR #109, opened 2019-12-04
[WIP] Token Interoperability

hkalodner:
Main Features:

- Store code point for contract along with storage and balance. This makes it easy to clone a new contract with the same code, but fresh balance and storage. This will enable both token contract cloning for this feature pipeline as well as adding public ArbSys support for contract cloning.

- Add contract templates for ERC20 and ERC721 contracts with custom Arbitrum functionality added. The custom functions are
    - `function adminMint(address account, uint256 amount) public` - This function is only callable by the VM OS and will add coins to a user's account after they have been deposited.
    - `function withdraw(uint256 amount) public` - This function is callable by any user and will trigger a withdrawal of the given number of tokens

Other Changes:

 - Simplify metadata removal (https://solidity.readthedocs.io/en/v0.5.13/metadata.html). We had been searching in the parsed instructions for the specific prefix bytes. It is much easier to just grab "its length [which] is added in a two-byte big-endian encoding".

- Enabled GASLIMIT opcode with constant limit of 10000000000. This opcode was being generated in token contract byte code and blocked compilation. I believe this may be a result of data being stored in the end of the code as described in https://ethereum.stackexchange.com/questions/15050/extra-byte-in-the-thedao-v1-bytecode/15289#15289


hkalodner:
Superseded by #126

PR #112, opened 2019-12-11
Initial implementation of ArbGas

ewfelten:
VM now tracks the total ArbGas consumption when building an assertion.

Disputable assertions include a (claimed) amount of ArbGas consumed. Bisection must divide the ArbGas consumption between sub-assertions. One-step proofs must use the correct amount of ArbGas. All of this is provided by validators and checked by the EthBridge.

For now, every instruction costs 1 unit of ArbGas.

Haven't yet implemented a facility for dapps to limit their ArbGas usage.


packages/arb-avm-cpp/avm/include/avm/opcodes.hpp
```diff
@@ -292,4 +292,67 @@ const std::map<OpCode, std::vector<bool>> InstructionAuxStackPops = {
     {OpCode::HALT, {}},
     {OpCode::DEBUG, {}}};
 
+const std::map<OpCode, uint64_t> InstructionArbGasCost = {
```
hkalodner: This should be std::unordered_map instead of std::map

packages/arb-avm-go/vm/instructions.go
```diff
@@ -122,37 +123,38 @@ func (w BlockedError) Error() string {
 	return "VMBlockederror"
 }
 
-func RunInstruction(m *Machine, op value.Operation) (StackMods, machine.BlockReason) {
+func RunInstruction(m *Machine, op value.Operation) (StackMods, uint64, machine.BlockReason) {
```
hkalodner: It doesn't look like the gas return value is used anywhere. It might be better to remove this.

PR #115, opened 2019-12-17
checkpoint refactor

minhptruong:


ewfelten:
This fails one of the CircleCI tests.  Looks like a build problem.

packages/arb-avm-cpp/avm_values/src/vmValueParser.cpp
```diff
@@ -39,9 +39,9 @@ const char* getContractData(const std::string& contract_filename) {
     return buf;
 }
 
-InitialVmValues getInitialVmValues(const std::string& contract_filename) {
+InitialVmValues getInitialVmValues(const std::string& contract_filename,
+                                   TuplePool& pool) {
```
hkalodner: This uses a reference, but the header declaration uses a pointer

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -32,7 +32,9 @@ CCheckpointStorage* createCheckpointStorage(const char* db_path,
     auto string_filename = std::string(db_path);
     auto string_contract_path = std::string(contract_path);
 
-    auto state = getInitialVmValues(string_contract_path);
+    TuplePool pool;
+
+    auto state = getInitialVmValues(string_contract_path, pool);
```
hkalodner: This leads to a dangling ref to pool

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -150,13 +152,13 @@ int saveData(CCheckpointStorage* storage_ptr,
...
-    auto results = transaction->saveValue(key_vector, value_vector);
+    auto results = transaction->saveData(key_vector, value_vector);
 
     if (results.status.ok()) {
         auto status = transaction->commit();
         return status.ok();
     } else {
-        return results.status.ok();
+        false;
     }
```
hkalodner: I think the false is missing a `return` 😁. Also I think it'd be nicer to structure this as
```cpp
if (!results.status.ok()) {
    return false;
}
auto status = transaction->commit();
return status.ok();
```
This also applies in the other locations

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -184,12 +186,12 @@ int deleteData(CCheckpointStorage* storage_ptr, const char* key) {
...
 
-    auto results = transaction->deleteValue(key_vector);
+    auto results = transaction->deleteData(key_vector);
 
     if (results.status.ok()) {
         auto status = transaction->commit();
         return status.ok();
     } else {
-        return results.status.ok();
+        false;
```
hkalodner: `return`

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -132,7 +134,7 @@ int deleteValue(CCheckpointStorage* storage_ptr, void* hash_key) {
         auto status = deleter.commitTransaction();
         return status.ok();
     } else {
-        return results.status.ok();
+        false;
```
hkalodner: `return`

packages/arb-avm-cpp/data_storage/include/data_storage/checkpoint/checkpointstorage.hpp
```diff
@@ -33,9 +33,10 @@ class TransactionDB;
 class CheckpointStorage {
    private:
     std::unique_ptr<DataStorage> datastorage;
-    const InitialVmValues& initial_state;
+    const InitialVmValues initial_state;
 
    public:
+    TuplePool* pool;
```
hkalodner: I think this should actually own the pool, `std::unique_ptr<TuplePool>`

packages/arb-avm-cpp/data_storage/src/checkpoint/checkpointstorage.cpp
```diff
@@ -25,9 +25,9 @@
 
 CheckpointStorage::CheckpointStorage(const std::string& db_path,
                                      const InitialVmValues& initial_state_)
-    : initial_state(initial_state_) {
-    auto db = new DataStorage(db_path);
-    datastorage = std::unique_ptr<DataStorage>(db);
+    : datastorage(std::make_unique<DataStorage>(db_path)),
+      initial_state(initial_state_) {
+    pool = new TuplePool();
```
hkalodner: unique_ptr!

packages/arb-avm-cpp/avm_values/include/avm_values/vmValueParser.hpp
```diff
@@ -24,9 +24,10 @@ struct InitialVmValues {
     bool valid_state;
     std::vector<CodePoint> code;
     value staticVal;
+    std::shared_ptr<TuplePool> pool;
```
hkalodner: I don't think this needs to store the pool

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
```diff
@@ -58,7 +58,9 @@ struct MachineState {
     BlockReason blockReason;
 
     MachineState();
-    MachineState(const std::vector<CodePoint>& code_, const value& static_val_);
+    MachineState(const std::vector<CodePoint>& code_,
+                 const value& static_val_,
+                 std::shared_ptr<TuplePool>& pool_);
```
hkalodner: Pass this `shared_ptr` by value instead of reference to show that this is going to make a copy

packages/arb-avm-cpp/avm/src/machinestate/machinestate.cpp
```diff
@@ -38,11 +38,10 @@ MachineState::MachineState()
...
 MachineState::MachineState(const std::vector<CodePoint>& code_,
-                           const value& static_val_)
-    : pool(std::make_unique<TuplePool>()),
-      pendingInbox(pool.get()),
-      context({0, 0}),
-      inbox(pool.get()) {
+                           const value& static_val_,
+                           std::shared_ptr<TuplePool>& pool_)
+    : pendingInbox(pool_.get()), context({0, 0}), inbox(pool_.get()) {
+    pool = pool_;
```
hkalodner: After changing `pool_` to pass by value, set `pool(std::move(pool_))` in the initializer list

packages/arb-avm-cpp/avm_values/include/avm_values/vmValueParser.hpp
```diff
@@ -24,9 +24,10 @@ struct InitialVmValues {
...
 };
 
-InitialVmValues getInitialVmValues(const std::string& contract_filename,
-                                   TuplePool* pool);
+InitialVmValues parseInitialVmValues(const std::string& contract_filename,
+                                     std::shared_ptr<TuplePool>& pool);
```
hkalodner: This can just pass a raw reference to `TuplePool`

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -55,7 +51,7 @@ CMachine* getInitialMachine(const CCheckpointStorage* storage_ptr) {
     auto state = storage->getInitialVmValues();
 
     if (state.valid_state) {
-        MachineState machine_state(state.code, state.staticVal);
+        MachineState machine_state(state.code, state.staticVal, state.pool);
```
hkalodner: Get `pool` out of `storage` instead of `state`

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -130,68 +127,75 @@ int deleteValue(CCheckpointStorage* storage_ptr, void* hash_key) {
...
+    auto key_ptr = reinterpret_cast<const char*>(key);
+    auto data_ptr = reinterpret_cast<const char*>(data);
+
+    auto key_str = std::string(key_ptr);
+    auto data_str = std::string(data_ptr);
+
     auto key_vector =
         std::vector<unsigned char>(key_str.begin(), key_str.end());
+    auto data_vector =
+        std::vector<unsigned char>(data_str.begin(), data_str.end());
```
hkalodner: This is super unsafe. The `std::string` constructor taking a `const char*` assumes that the pointer is to a c string which is null terminated. The go calls this with the pointer to the beginning of a byte slice which is not null terminated. Instead you should also pass the key and value lengths and then construct the vectors directly from the pointers.
ex: `auto key_vector = std::vector<unsigned char>(key, key + key_length);`

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -130,68 +127,75 @@ int deleteValue(CCheckpointStorage* storage_ptr, void* hash_key) {
...
 }
 
-char* getData(const CCheckpointStorage* storage_ptr, const char* key) {
+ByteSlice getData(const CCheckpointStorage* storage_ptr, const void* key) {
     auto storage = static_cast<const CheckpointStorage*>(storage_ptr);
     auto transaction = storage->makeConstTransaction();
 
-    auto key_str = std::string(key);
+    auto key_ptr = reinterpret_cast<const char*>(key);
+    auto key_str = std::string(key_ptr);
```
hkalodner: Unsafe, see above

packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp
```diff
@@ -130,68 +127,75 @@ int deleteValue(CCheckpointStorage* storage_ptr, void* hash_key) {
...
 }
 
-int deleteData(CCheckpointStorage* storage_ptr, const char* key) {
+int deleteData(CCheckpointStorage* storage_ptr, const void* key) {
     auto storage = static_cast<CheckpointStorage*>(storage_ptr);
     auto transaction = storage->makeTransaction();
 
-    auto key_str = std::string(key);
+    auto key_ptr = reinterpret_cast<const char*>(key);
+    auto key_str = std::string(key_ptr);
```
hkalodner: Unsafe, see above

packages/arb-avm-cpp/avm/src/machinestate/machinestate.cpp
```diff
@@ -39,9 +39,11 @@ MachineState::MachineState()
...
 MachineState::MachineState(const std::vector<CodePoint>& code_,
                            const value& static_val_,
-                           std::shared_ptr<TuplePool>& pool_)
-    : pendingInbox(pool_.get()), context({0, 0}), inbox(pool_.get()) {
-    pool = pool_;
+                           std::shared_ptr<TuplePool> pool_)
+    : pool(std::move(pool_)),
+      pendingInbox(pool_.get()),
+      context({0, 0}),
+      inbox(pool_.get()) {
```
hkalodner: `pool_` has been moved into `pool` so it is now invalid. Use `pool.get()` instead

PR #120, opened 2020-01-07
TransactionReceipt implementation

julientregoat:
While writing tests for our dApp, I noticed that I was unable to call `TransactionReceipt`. It turns out we were not fully implementing the `DeployBackend` [interface described in geth.](https://godoc.org/github.com/ethereum/go-ethereum/accounts/abi/bind#DeployBackend). Implementing this interface also gives us compatibility with the `WaitDeployed` and `WaitMined` functions present in the geth `bind` package, furthering our Ethereum interoperability.

Currently, I had to zero out a few properties that don't necessarily translate from Arbitrum to Ethereum. I followed the examples left in `txtracker.go` and `provider.ts` to mimic the current 'translation' as much as possible.  However, there are some properties that weren't covered in those files.

- `Bloom` - we don't currently have a bloom filter system implemented for our logs yet, zeroed
- `PostState` - from what I can tell, this would translate to a hash of the state of the VM after this transaction was executed. I see various hashes in `txtracker.go` but I'm not sure which would be appropriate; `txInfo.PartialHash` seemed like a possibility. Or maybe we can compute the hash using `assertionInfo.BeforeHash`? 

I have not tested this yet, hence the draft state, but wanted to open it up for discussion.

Also
 - left a TODO for something noted by @hkalodner - `sendTransaction` [should return as soon as the tx is pending](https://godoc.org/github.com/ethereum/go-ethereum/accounts/abi/bind#ContractTransactor). Currently, it'll loop until the tx confirms or encountering an error while calling `GetMessageResult`. I'm not sure of the original implementation reasoning, but `TransactionReceipt` + `WaitDeployed` / `WaitMined` can cover this functionality in the expected way. 
 - Fixed a typo in an error message (`validatorProxy.go`)


packages/arb-provider-go/connection.go
```diff
@@ -371,3 +372,86 @@ func (sub *subscription) Unsubscribe() {
...
+// TransactionReceipt returns the receipt of a transaction by transaction hash.
+// Note that the receipt is not available for pending transactions.
+func (conn *ArbConnection) TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
+	result, ok, err := conn.proxy.GetMessageResult(txHash.Bytes())
+	if err != nil {
+		log.Println("TransactionReceipt error:", err)
+		return nil, err
+	} else if !ok {
+		// transaction was not found - throw an error?
+		return nil, nil
```
hkalodner: It looks like ethclient returns `nil, ethereum.NotFound` in this case

packages/arb-provider-go/connection.go
```diff
@@ -371,3 +372,86 @@ func (sub *subscription) Unsubscribe() {
...
+			}
+		}
+	}
+
+	var newContractAddr common.Address
+	if ethMsg.Data.CallData.ContractID != nil {
+		newContractAddr = common.BytesToAddress(ethMsg.Data.CallData.ContractID.Bytes()[12:])
+	} else {
+		newContractAddr = common.BytesToAddress([]byte{0})
+	}
```
hkalodner: We don't currently support contract creation through the eth transaction mechanism

julientregoat:
Changes made. 

Re: documenting the mapping, I can draft a `Arbitrum Ethereum Interoperability Support.md`? Not sure if that's the best name, it doesn't exactly flow off the tongue!

hkalodner:
I think a document would be great. It could start out pretty simple, but it would be helpful for making sure we keep all provider implementations in sync.

hkalodner:
I'm gonna merge these fixes in since we want to improve the `arb-provider-go` and this seems like a strict improvement over the status quo.

PR #124, opened 2020-01-09
Rollup mock eth

tpobryan:
Changed ethclient to arbclient interface everywhere outside of ethbridge, cleanup of some copyright notices, and stubbed mockbridge.


tpobryan:
Cleaned up. Now you make decision if ethbridge or mockbridge when you create the client. This could be wrapped with an option too choose like we did with cpp vs. go VM if we want later.

PR #128, opened 2020-01-15
Add RollupCreated event

julientregoat:
and update typechain defs. there was a very small change to the `GlobalPendingInbox` bytecode, the source of which is indeterminate to me as I did not make any changes. the definitions were built off of built JSON from contracts on the most recent version of the `rollup` branch so it should be accurate in that sense.


hkalodner:
It might be that the ABI for global pending inbox hadn't been regenerated since the previous edit to it. Overall I mostly just ignore diffs on generated files in pull requests

PR #132, opened 2020-01-22
Fix cors header problem with geth

hkalodner:


julientregoat:
not sure if this actually fixed it @hkalodner - metamask is still unable to connect

PR #137, opened 2020-01-24
Ethhash2 instruction

ewfelten:
This creates a new AVM instruction, Ethhash2, which takes two int from the stack, concatenates their contents, and computes the Ethereum hash of the result.  

The hashing library in AVM standard library is updated to use this where possible.


packages/arb-compiler-evm/arb-test.py
```diff
@@ -499,6 +520,13 @@ def test_tuple(vm):
 # print(vm.code)
 with open("../arb-validator/proofmachine/opcodetesthash.ao", "wb") as f:
     arb.marshall.marshall_vm(vm, f)
+code = arb.compile_block(test_ethhash2)
+vm = arb.compile_program(arb.ast.BlockStatement([]), code)
+vm.static = 4
+print("ethhash2 ", len(vm.code), " codepoints")
```
hkalodner: Remove extraneous print statement

ewfelten: Current code does this print for every test block.

hkalodner: It turns out this mirrors the rest of the file so the change is unneeded.

PR #141, opened 2020-01-26
Evil rollup manager, for testing, and some tests

ewfelten:
Submitting a PR for this now because this includes some code useful for testing, as well as a refactor of the checkpointing API.


packages/arb-bridge-eth/contracts/vm/Snapshot.sol
```diff
@@ -0,0 +1,142 @@
+/*
```
hkalodner: This file should be eliminated

PR #149, opened 2020-01-28
Time functionality

### LINKED ISSUE #143: Update handling of time in EVM programs
This has several pieces.  
- [ ] Ensure that assertion time bounds are not more than K blocks apart.  
- [ ] Make that K a parameter of the chain, with advice to developers about how to set it.  K should generally be a few blocks.
- [ ] Update message/time library in VM-OS to handle time correctly.   Library should track the max of (a) timestamps on messages given to VM, and (b) values returned from previous library calls that asked for the time. Calls to get time should return max of that tracked value and the current time lower bound.
- [ ] Have compiler generate call to message/time library for when EVM program asks for time. 
- [ ] Write documentation for devs on how to use time in rollup.  This stuff is basically the same for all rollup systems (whether they admit it or not), but we need to explain it clearly.

ewfelten:
Fixes #143


docs/Time_in_Arbitrum.md
```diff
@@ -0,0 +1,48 @@
...
+
+Most developers will be fine using ``block.number``.
+If you don't mind a little bit of time lag, you can keep on using ``block.number`` just like you would on Ethereum.
+
+### Setting deadlines for user response
+An exception is a use case like an auction, where the contract emits an event, and you want to give people out there in userland N block times to respond.
+To do that, you'll want to "start the clock ticking" by using code like this:
+
+    deadline = ArbSys(address(100)).timeUpperBound() + N;
+    emit StartAuction(..., deadline+N);
```
hkalodner: This shouldn't have the "+N" I think

docs/Time_in_Arbitrum.md
```diff
@@ -0,0 +1,48 @@
...
+Generally, some time will elapse between the time when a message is put into the pending inbox (and timestamped) and the time when the contract processes the message and carries out the transaction requested by the message.
+
+If your Solidity contract, running on Arbitrum Rollup, accesses ``block.number``, this will give you the ArbChain's current time. 
+If you want to know when the current message (the one that requested the transaction you're currently running) was timestamped, you can get it in Solidity by writing:
+
+    ArbSys(address(100)).currentMessageTime()
+
+Because ArbChains operate off-chain, an ArbChain's time might differ a little bit from the current Ethereum block time.
+The ArbChain's clock will never get ahead of the Ethereum block time, but the ArbChain's clock might "run a bit slow". 
+In particular, it might be as many as 20 blocks behind the true Ethereum time.
```
hkalodner: Are we settling on 20, or is this going to be parameterized and 20 is an example?

docs/Time_in_Arbitrum.md
```diff
@@ -0,0 +1,48 @@
...
+These messages are put into the chain's *pending inbox*.
+Every message is timestamped with the time it was put into the pending inbox.
+
+Messages in an ArbChain's pending inbox are processed in order.
+Generally, some time will elapse between the time when a message is put into the pending inbox (and timestamped) and the time when the contract processes the message and carries out the transaction requested by the message.
+
+If your Solidity contract, running on Arbitrum Rollup, accesses ``block.number``, this will give you the ArbChain's current time. 
+If you want to know when the current message (the one that requested the transaction you're currently running) was timestamped, you can get it in Solidity by writing:
+
+    ArbSys(address(100)).currentMessageTime()
```
hkalodner: Somewhere we should separately explain ArbSys calls since the syntax is weird

packages/arb-compiler-evm/contract-templates/contracts/ArbSys.sol
```diff
@@ -28,4 +28,10 @@ interface ArbSys {
 
     // Send given amount of Eth to dest with from sender.
     function withdrawEth(address dest, uint256 amount) external;
+
+    // Return time when current message was posted on-chain
+    function currentMessageTime() external returns(uint);
```
hkalodner: I think `messageTime` or `timeDelivered` might be better here. The word current is a bit weird. It's true that it's the current message we're processing, but it's kind of redundant in Ethereum's model of always being in a transaction context

hkalodner: Also `timeDelivered` has the advantage of matching `timeUpperBound`. It'd be nice to have a uniform naming scheme

docs/Time_in_Arbitrum.md
ewfelten: We can change this later, but I wanted to put in a concrete value, and 20 seems reasonable.  Happy to change it if others think another number is better.

docs/Time_in_Arbitrum.md
ewfelten: Maybe another documentation page about ArbSys?

packages/arb-compiler-evm/contract-templates/contracts/ArbSys.sol
ewfelten: I prefer not to use "delivered" because it's not when the message was delivered to the contract, it's when the message was submitted to the pending inbox.  We could go with messageTime.  Or maybe messageSubmissionTime?

docs/Time_in_Arbitrum.md
ewfelten: Fixed this one.

docs/Time_in_Arbitrum.md
ewfelten: I'm creating an issue to document ArbSys.

packages/arb-compiler-evm/contract-templates/contracts/ArbSys.sol
ewfelten: Actually, this is a non-mutating operation, so it could be invoked in the context of a non-mutating call.  Not clear what it should do in that case.  (And this might affect how we name it.)

docs/Time_in_Arbitrum.md
hkalodner: Sounds good to me. 20 sounds good to me as a place to start.

packages/arb-compiler-evm/contract-templates/contracts/ArbSys.sol
hkalodner: In a non-mutating call currently we just use the current time as the time the message is marked with. I also use a timebounds where the upper and lower bound are just the current time.

ewfelten: So this, as currently implemented, will return the current time if invoked by/during a call?

hkalodner: Yep. All three time values should be equal for a call. We can do that since we know exactly when it's executed. At least that's what makes sense to me currently

PR #154, opened 2020-01-29
Evil rollup server testing

ewfelten:
The evilRollupServer behaves just like the normal rollupServer, except that the evil one sometimes has false views about the hashes of Machines.  These cause it to get into execution challenges (on both sides of the challenge) against honest validators, and to fight those challenges to the bitter end.


packages/arb-validator/rollup/nodeGraph.go
```diff
@@ -236,18 +236,26 @@ func (chain *NodeGraph) CommonAncestor(n1, n2 *Node) *Node {
 	return n1.prev
 }
 
-func (chain *NodeGraph) generateNodePruneInfo() []pruneParams {
+func (chain *NodeGraph) generateNodePruneInfo(stakers *StakerSet) []pruneParams {
```
hkalodner: Why does this function take `stakers` rather than using  `chain.stakers`?

packages/arb-validator/cmd/rollupServer/rollupServer.go
```diff
@@ -195,13 +195,19 @@ func validateRollupChain() error {
 	}
 
 	ctx := context.Background()
-	manager, err := rollupmanager.CreateManager(ctx, address, true, client, checkpointing.NewRollupCheckpointerImplFactory(
+	manager, err := rollupmanager.CreateManager(
```
hkalodner: This api should probably be cleanup up for the common case. Maybe introduce a CreateManager and CreateManagerAdvanced or something like that. We want a common case API that's minimized for inclusion by other users.

packages/arb-validator/rollup/nodeGraph.go
ewfelten: Because chain is a NodeGraph, which doesn't have a stakers field.  (stakers is a field of StakedNodeGraph, which is a "superclass" of NodeGraph.)

hkalodner: Ah right. So then probably it should be moved into `StakedNodeGraph`

ewfelten: Did that.

packages/arb-validator/cmd/rollupServer/rollupServer.go
ewfelten: Did that.

PR #167, opened 2020-02-02
Golang linting [WIP]

hkalodner:
* Use golangci-lint to clean up go packages
* Use clang-tidy to clean up arb-avm-cpp


hkalodner:
It's not worth the large amount of code changes. I'm gonna separately put in some of the individual fixes

PR #168, opened 2020-02-02
Blocktime commandline switch

ewfelten:
Add a commandline switch for blocktime to rollupServer and evilRollupServer (when in validate mode).


packages/arb-util/common/timeticks.go
```diff
@@ -21,9 +21,7 @@ import (
 	"time"
 )
 
-const (
-	AverageSecondsPerBlock = 2
-)
+var AverageSecondsPerBlock = int64(2)
```
hkalodner: It looks like this should also be used for the _timeConversionFactor which is currently hardcoded to 13. The duration calculation also has a divided by 13 in it which seems like it shouldn't been there.

AverageSecondsPerBlock should also be of type `time.Duration` rather than a raw int64. The will also allow for sub second granularity which could be useful for some deployments.

ewfelten: I'll change this to be a time.Duration.

On the other issue, the logic has been that there are 13000 ticks per block, so the wall-clock duration of a tick might vary.  If the number of ticks per block varies, we'll have to modify the EthBridge to know that.  I think it's better to keep the EthBridge-side conversation factor constant, and vary the ticks-to-Duration conversion on the client.

hkalodner: Ah right. Makes sense

ewfelten: I did some work to better rationalize the handling of times on the client side.  This is still not ideal in many respects.  The use of three different time units (blocks, ticks, Duration) is annoying, but maybe necessary because the three are really doing different work.

That said, it probably makes sense to have the number of ticks per block be a number more "natural" than 13000.

ewfelten: The reason we have ticks at all is so that deadlines can be denominated in fractions of a blocktime.  (Because deadlines are computed based on past deadlines, if we just rounded up in every case, the roundoff errors could accumulate and slow down the max VM speed.)

hkalodner:
Replaced by #178

PR #169, opened 2020-02-02
Confirm and Prune multiple

hkalodner:
* Replace `ConfirmValid` and `ConfirmInvalid` with a single `Confirm` that can confirm multiple nodes
* Replace `PruneLeaf` with `PruneLeaves` that can prune multiple nodes
* Fix circleci code coverage


packages/arb-bridge-eth/contracts/vm/NodeGraph.sol
```diff
@@ -107,23 +107,45 @@ contract NodeGraph is ChallengeType {
...
+            latestConfirmedProofLengths.length == pruneCount,
+            "input length mistmatch"
         );
-        delete leaves[leaf];
-
-        emit RollupPruned(leaf);
+        uint256 prevLeafOffset = 0;
+        uint256 prevConfOffset = 0;
+        for (uint256 i = 0; i < pruneCount; i++) {
+            bytes32 from = fromNodes[i];
```
ewfelten: I think we need to add `require(leafProofLengths[i] > 0 && latestConfirmedProofLengths[i] > 0)` here.

packages/arb-bridge-eth/contracts/vm/NodeGraph.sol
```diff
@@ -107,23 +107,45 @@ contract NodeGraph is ChallengeType {
         bytes32 logsAccHash;
     }
 
-    function pruneLeaf(
-        bytes32 from,
-        bytes32[] calldata leafProof,
-        bytes32[] calldata latestConfirmedProof
+    function pruneLeaves(
+        bytes32[] memory fromNodes,
```
ewfelten: Maybe better to have a single `fromNode` and a single proof from there to `latestConfirmed`, then multiple path proofs to prunable leaves.  The single `fromNode` would just be the oldest ancestor of all of the leaves we are pruning.

packages/arb-validator/mockbridge/arbRollup.go
```diff
@@ -196,60 +196,7 @@ func (vm *ArbRollup) MakeAssertion(
...
-	//	protoHash,
-	//	stakerAddresses,
-	//	stakerProofs,
-	//	stakerProofOffsets,
-	//)
-	//if err != nil {
-	//	return err
-	//}
-	//return vm.waitForReceipt(ctx, tx, "ConfirmInvalid")
+func (vm *ArbRollup) Confirm(ctx context.Context, opp *valprotocol.ConfirmOpportunity) error {
```
ewfelten: This looks like it's stubbed out.  Is that intended?

packages/arb-validator/rollup/listener.go
```diff
@@ -43,9 +48,8 @@ type ChainListener interface {
 	MessageDelivered(context.Context, *ChainObserver, arbbridge.MessageDeliveredEvent)
 
 	AssertionPrepared(context.Context, *ChainObserver, *preparedAssertion)
-	ValidNodeConfirmable(context.Context, *ChainObserver, *confirmValidOpportunity)
-	InvalidNodeConfirmable(context.Context, *ChainObserver, *confirmInvalidOpportunity)
-	PrunableLeafs(context.Context, *ChainObserver, []pruneParams)
+	NodesConfirmable(context.Context, *ChainObserver, *valprotocol.ConfirmOpportunity)
+	PrunableLeafs(context.Context, *ChainObserver, []valprotocol.PruneParams)
```
ewfelten: The naming here is inconsistent.  NodesConfirmable is noun-adjective.  PrunableLeafs is adjective-noun.  Should be consistent.

packages/arb-validator/rollup/node.go
```diff
@@ -311,6 +328,21 @@ func GeneratePathProof(from, to *Node) []common.Hash {
...
+	if from == to {
+		return []common.Hash{}
+	}
+	sub := GeneratePathProof(from, to.prev)
+	if sub == nil {
+		return nil
+	}
+	return append(sub, to.innerHash)
+}
+
```
ewfelten: This looks very similar to GeneratePathProof.  Suggest adding a comment explaining the difference.

packages/arb-validator/rollup/stakedNodeGraph.go
```diff
@@ -163,110 +168,140 @@ func (chain *StakedNodeGraph) generateNodePruneInfo(stakers *StakerSet) []pruneP
...
+
+			confirmable := sng.isConfirmableNode(
+				node,
+				currentTime,
+				stakerAddrs,
+			)
+			if confirmable {
+				var confOpp valprotocol.ConfirmNodeOpportunity
+				if node.linkType == valprotocol.ValidChildType {
+					if node.assertion == nil {
```
ewfelten: Suggest a comment here explaining why we're not confirming the node because we don't know the assertion that made it.

And: Wasn't the assertion posted on-chain?  So how do we not know it?

packages/arb-validator/rollup/stakedNodeGraph.go
```diff
@@ -163,110 +168,140 @@ func (chain *StakedNodeGraph) generateNodePruneInfo(stakers *StakerSet) []pruneP
...
-		offsets = append(offsets, new(big.Int).SetUint64(uint64(len(proof))))
+		stakeCount++
 	}
-	if len(offsets) == 1 {
-		return nil, nil
+	if stakeCount == 0 {
+		return false
 	}
-	return proof, offsets
+	return true
```
ewfelten: Suggest using `return stakeCount > 0`

packages/arb-bridge-eth/contracts/vm/NodeGraph.sol
ewfelten: Fixed this. Will push soon.

packages/arb-bridge-eth/contracts/vm/NodeGraph.sol
ewfelten: Not fixing this now.  Creating an issue for this instead.

packages/arb-validator/mockbridge/arbRollup.go
hkalodner: The real mockbridge stuff is all on a separate branch and will have to independently be ported to this

packages/arb-validator/rollup/listener.go
hkalodner: Fixed

packages/arb-validator/rollup/node.go
hkalodner: This was unused so I'm just killing it. I don't remember why it was originally included

packages/arb-validator/rollup/stakedNodeGraph.go
hkalodner: Added comment.
```go
// We need to know the contents of the actual assertion to confirm it
// We've only seen the hash accumulator of the messages before whereas this requires the full values
```

packages/arb-validator/rollup/stakedNodeGraph.go
hkalodner: Done

PR #178, opened 2020-02-04
Milliblock ticks

ewfelten:
Change a "tick" to be 1/1000 of a block.


packages/arb-util/common/timeticks.go
```diff
@@ -21,28 +21,18 @@ import (
...
-}
-
 func TimeFromBlockNum(blockNum *TimeBlocks) TimeTicks {
-	return TimeTicks{new(big.Int).Mul(_timeConversionFactor, blockNum.AsInt())}
+	return TimeTicks{new(big.Int).Mul(big.NewInt(TicksPerBlock), blockNum.AsInt())}
 }
 
 func TimeFromSeconds(seconds int64) TimeTicks {
-	return TimeTicks{new(big.Int).Mul(_timeTicksPerSecond, big.NewInt(seconds))}
+	return TimeTicks{big.NewInt(int64(time.Duration(seconds*TicksPerBlock) * time.Second / _durationPerBlock))}
```
hkalodner: These should probably be titles "Ticks" instead of "Time" for clarity

ewfelten: Done.

PR #184, opened 2020-02-05
Augur

minhptruong:


hkalodner:
Rather than pushing all of augur into our main repo, it'd be better to fork augur into an OffchainLabs  copy and make the changes there

PR #185, opened 2020-02-05
Add ArbFactory support to arb-provider-ethers

hkalodner:
Fix ABI generation and add more ABI's to arb-provider-ethers


packages/arb-provider-ethers/src/lib/value.ts
```diff
@@ -30,9 +30,9 @@ export enum OpCode {
...
 // Valid opcode ranges (inclusive)
 const OP_CODE_RANGES: Array<[number, number]> = [
-    [0x01, 0x0a],
+    [0x00, 0x0a],
     [0x10, 0x1b],
-    [0x20, 0x21],
+    [0x20, 0x22],
     [0x30, 0x3d],
     [0x40, 0x44],
     [0x50, 0x52],
```
ewfelten: This disagrees with the AVM spec in two places:  AVM spec says that 0x00 and 0x75 are not valid instructions.  So I think the relevant lines should be [0x01, 0x0a] and (lower down) [0x70, 0x74].

hkalodner: The reason for the top one is that the default error pc uses a value of 0 currently. We could change that, but it would be much more work thank just adjusting this. 0x75 is in there since we have a DEBUG opcode that isn’t part of the official spec, but gets used for internal testing and debugging.

hkalodner: I’d argue that leaving this as is would be best, but I’m also open to making adjustments.

ewfelten: OK, I'm convinced.

ewfelten:
Conflicts need to be resolved.

hkalodner:
@ewfelten, this should be good to go after the circleci checks complete

PR #186, opened 2020-02-05
Gas price commandline option

ewfelten:


packages/arb-validator/cmd/rollupServer/rollupServer.go
```diff
@@ -159,6 +161,7 @@ func validateRollupChain() error {
 
 	// Rollup creation
 	auth := bind.NewKeyedTransactor(key)
+	auth.GasPrice = big.NewInt(*gasPrice)
```
hkalodner: The GasPrice field is denominated in Wei, but this uses a value in Gwei

ewfelten: Fixed. Now take commandline option as a float64 in Gwei, which gets converted to Wei.

packages/arb-validator/cmd/rollupServer/rollupServer.go
```diff
@@ -159,6 +161,7 @@ func validateRollupChain() error {
 
 	// Rollup creation
 	auth := bind.NewKeyedTransactor(key)
+	auth.GasPrice = big.NewInt(int64(1e9*(*gasPrice))
```
hkalodner: Approving with the note that this line is only safe since we can be confident that 1e9*gasPrice will be significantly less that uint64 max.

ewfelten: Pushed commit that checks for overflow

PR #188, opened 2020-02-05
fix branch issues preventing arb-provider-ethers from compiling

julientregoat:
I hit some issues when trying to run `update-abi` and `build` on this branch.

- couldn't find `prettier/@typescript-eslint` package anywhere so assumed it was a typo
- ensure a local version of `eslint` is used
- ensure deps are installed when calling `update-abi`
- update contract bindings


packages/arb-provider-ethers/src/lib/provider.ts
```diff
@@ -296,11 +296,11 @@ export class ArbProvider extends ethers.providers.BaseProvider {
                     return '0x100';
                 }
                 const arbInfo = ArbInfoFactory.connect(ARB_INFO_ADDRESS, this);
-                return arbInfo.getCode(params.address, { blockTag: params.blockTag });
```
hkalodner: Why did you remove the blockTag on these? It's true we ignore it when we actually make the call, but it seemed like leaving it in place would make it easier to implement the blockTag stuff in the future.

julientregoat: ah sorry, I wasn't trying to make a change that would affect anything. since the contract interface doesn't describe a second parameter, typescript doesn't compile since it sees an extra parameter. I can put this back and add a `@ts-ignore`?

PR #189, opened 2020-02-06
Create Glossary.md

ewfelten:


docs/Glossary.md
```diff
@@ -0,0 +1,20 @@
...
+* Challenge: When two stakers disagree about the correct verdict on an assertion, those stakers can be put in a challenge. The challenge is refereed by the EthBridge. Eventually one staker wins the challenge. The loser forfeits their stake. Half of the loser's stake is given to the winner, and the other half is burned. 
+* Client: A program running on a user's machine, often in the user's browser, that interacts with contracts on an Arbitrum chain and provides a user interface.
+* Compiler: A part of the Arbitrum toolchain that takes a set of Solidity contracts and compiles them into an Arbitrum executable file.
+* Confirmation: The final decision by an Arbitrum chain to accept a node as being a settled part of the chain's history. When a node is confirmed, any funds paid out by the chain to the main Ethereum chain are transferred.
+* EthBridge: A group of contracts running on the Ethereum chain, which act as a record-keeper and rule enforcer for Arbitrum chains.
+* Node: A particular point in the history of an Arbitrum Chain. A node corresponds to a sequence of assertions that have been made, and a verdict about whether each of those assertions was accepted.
```
hkalodner: The term node is potentially confusing because of the naming collision between this and the usage to describe blockchain client software ex. "Ethereum Node". We should consider whether there is a way to disambiguate this.

docs/Glossary.md
```diff
@@ -0,0 +1,20 @@
...
+* Pending inbox: Holds a sequence of messages sent by clients to the contracts on an Arbitrum Chain. The pending inbox is managed by the on-chain EthBridge. Every message is timestamped with the Ethereum block number when it was put into the pending inbox.
+* Staker: A party who deposits a stake, in Eth, to vouch for a particular node in an Arbitrum Chain. A party who stakes on a false node can expect to lose their stake. An honest staker can recover their stake once the node they are staked on has been confirmed.
+* Validator: A program, run on some user's local computer, that keeps track of the state of an Arbitrum Chain.
```
hkalodner: We should consider describing "Validator" as a role rather than as a program. That role is implemented using software, but it's somewhat more general than that I think. I guess this difference comes out in the description of staker vs validator.

ewfelten: Changed this.  Now a validator is "a party who..."

docs/Glossary.md
ewfelten: Agree that "node" is not ideal.  Logically, it's something like a proposed history of the chain.  "Chain state" isn't quite right, because the state of the chain includes a bunch of these nodes.  

In one version of the original spec, I called the boxes "verdicts".  The idea being that it represents a judgment about which assertions have been accpted.

Thoughts?

ewfelten: I'm going to go with "chain state".

Issue #201, opened 2020-02-09
Validators don't assert after kill-restart


ewfelten:
After a validator is killed and restarted, it no longer makes assertions.  This occurs in both the old and new checkpointers, so it is probably a problem in the validator-side logic.

To replicate, with local geth chain running:
$ yarn install:validator
$ ./scripts/setup_rollup.py --geth packages/arb-compiler-evm/infiniteloop.ao 2
$ rollupServer validate validator-states/validator0 ws://127.0.0.1:7546 0x377477e6aceb3df1740ce348439ee865a55b1a14
[wait a while, then control-C the rollupServer]
[re-run the rollupServer with the same command line]

Everything will look normal, with stakers advancing their stakes, nodes getting confirmed, and dead nodes getting pruned. But the validator won't try to compute assertions.

The same thing happens if you use arb_deploy instead of running rollupServer directly.

Issue #200, opened 2020-02-08
Checkpointer crashes when reorg reverses an earlier reorg


ewfelten:
The checkpointer tries to clean up block when they are mooted by a reorg. This becomes a problem if a reorg unmoots a previously mooted block, setting up a race between the cleanup process and users of the unmooted block.

The fix is to change the checkpointer's data structures so it doesn't try to keep a linked list of checkpoints, and it only deletes a checkpoint when it is aged out. 

This bug triggers a crash of the validator due to dereferencing nil. Workaround is to delete the checkpoint_db and then restart the validator.

PR #202, opened 2020-02-09
Checkpointer data structure fix

ewfelten:


ewfelten:
Fixes #200 
Fixes #201

packages/arb-validator/rollupmanager/manager.go
```diff
@@ -128,15 +128,23 @@ func CreateManagerAdvanced(
...
+					func(chainObserverBytes []byte, restoreCtx checkpointing.RestoreContext) {
+						chainObserverBuf := &rollup.ChainObserverBuf{}
+						if err := proto.Unmarshal(chainObserverBytes, chainObserverBuf); err != nil {
+							log.Fatal(err)
+						}
+						var err error
+						chain, err = chainObserverBuf.UnmarshalFromCheckpoint(runCtx, restoreCtx, checkpointer)
+						if err != nil {
+							log.Fatal(err)
+						}
```
hkalodner: It would probably be cleaner to have the callback function return an error that get's propagated through rather than failing here

packages/arb-validator/rollupmanager/manager.go
```diff
@@ -128,15 +128,23 @@ func CreateManagerAdvanced(
...
+					clnt,
+					rollupAddr,
+					updateOpinion,
+					func(chainObserverBytes []byte, restoreCtx checkpointing.RestoreContext) {
+						chainObserverBuf := &rollup.ChainObserverBuf{}
+						if err := proto.Unmarshal(chainObserverBytes, chainObserverBuf); err != nil {
+							log.Fatal(err)
+						}
+						var err error
+						chain, err = chainObserverBuf.UnmarshalFromCheckpoint(runCtx, restoreCtx, checkpointer)
```
hkalodner: Just to be clear, does this callback function run synchronously with the RestoreLatestState function? It should be that the callback has already finished running when the outer function returns for this code to be safe. That property should be clearly stated in the code since callbacks tend to be asynchronous rather than synchronous.

packages/arb-validator/checkpointing/indexedCheckpointer.go
```diff
@@ -0,0 +1,477 @@
...
+		new(big.Int).Set(maxReorgHeight),
+		nil,
+		nil,
+	}
+	go ret.writeDaemon()
+	go ret.cleanupDaemon()
+	return ret
+}
+
+func (cp *IndexedCheckpointer) New(_ context.Context) RollupCheckpointer {
```
hkalodner: This method is pretty strange looking. Is there a reason for it?

packages/arb-validator/checkpointing/indexedCheckpointer.go
```diff
@@ -0,0 +1,477 @@
...
+		log.Fatal(err)
+	}
+	return append([]byte{2}, bytesBuf...)
+}
+
+func (cp *IndexedCheckpointer) RestoreLatestState(
+	ctx context.Context,
+	clnt arbbridge.ArbClient,
+	_ common.Address,
+	_ bool,
```
hkalodner: Are these parameters used anywhere? If not they should be removed.

packages/arb-validator/checkpointing/indexedCheckpointer.go
```diff
@@ -0,0 +1,477 @@
...
+				ckpWithMan := &CheckpointWithManifest{}
+				if err := proto.Unmarshal(val, ckpWithMan); err != nil {
+					return err
+				}
+				callback(ckpWithMan.Contents, cp.newRestoreContextLocked())
+				return nil
+			}
+		}
+	}
+	callback(nil, nil)
```
hkalodner: Does the callback know what to do when nil is passed as both params? This is fairly nonstandard looking. What does it mean if the code reaches this point?

packages/arb-validator/rollupmanager/manager.go
ewfelten: Yup, fixed this.

packages/arb-validator/rollupmanager/manager.go
ewfelten: Yes, it runs synchronously.  I changed the parameter name to "unmarshalFunc" (rather than "callback") and noted in a comment on the interface that the callback is synchronous.

packages/arb-validator/checkpointing/indexedCheckpointer.go
ewfelten: Yes.  This code has to satisfy an interface that uses a factory pattern.  The interface does this so it can kill the checkpointer on a reorg, then call the factory again to make a new checkpointer to use going forward.

But this code doesn't do a kill/restart on reorg. This code is reorg-aware, so it just make a single instance once and returns the same one every time the factory is called.  And rather than having a separate factory object, I just made a single object that acts as its own factory, returning itself.

Added a comment to the source code explaining this.

packages/arb-validator/checkpointing/indexedCheckpointer.go
ewfelten: Eliminated these params, here and in the interface that this is implementing.  (No implementation of the interface used them.)

packages/arb-validator/checkpointing/indexedCheckpointer.go
ewfelten: Replaced this with a log.Fatal, because this should never happen.

PR #206, opened 2020-02-10
Fix integration bug with new checkpointer

ewfelten:


packages/arb-validator/rolluptest/evilRollupCheckpointer.go
```diff
@@ -51,15 +51,15 @@ func NewEvilRollupCheckpointerFactory(
 }
 
 type evilRollupCheckpointer struct {
-	cp *checkpointing.RollupCheckpointerImpl
+	cp checkpointing.RollupCheckpointer
```
hkalodner: Is there a reason this isn't just a `checkpointing.RestoreContext` to avoid the typecasting?

ewfelten: We need this to be both a RollupCheckpointer and a RestoreContext.  If I made it a RestoreContext there would be even more typecasting.

hkalodner: Ah

PR #209, opened 2020-02-10
Limit workahead

edfelten:
This limits how far in the future an assertion's deadline can be. Assertions that would violate the limit are rejected by the EthBridge.


ewfelten:
It's not simple for the client to predict whether an assertion will fail this check.  Probably the cleanest way would be for the chain object to keep track of the deadline on the latest assertion seen, and then the code that submits assertions would add the verification time for a new assertion to that, and wait until that deadline was less than 2 grace periods in the future.  The assertion submit code would need to be able to block until the desired block showed up.  So, possible but not simple.

hkalodner:
Can't we just create assertions and then discard them if they're invalid like we do with the timebounds check in the opinion thread?

ewfelten:
The client who makes an assertion doesn't know what deadline that assertion is going to get. The deadline is assigned by the EthBridge, and depends on block.number at the time that the assertion is recognized on-chain.  The client can derive a lower bound on the workahead amount for a new assertion, but that depends on both (1) the actual deadline assigned by the EthBridge to the latest assertion, and (2) the current block number. It's awkward to make both of those available to the code that would submit an assertion.

Also, if an assertion will be rejected at the current time, it might be acceptable later once block.number has advanced. So the client will want to wait for block.number to increase, then check again for whether it can submit.  Again, possible but a hassle.

edfelten:
I think we'll need an 80% solution here.  The client can keep track of the timeout on the last assertion that was made online. If that is more than, say, 1.8 challenge periods in the future, then the client should just sleep for a while instead of trying to generate more assertions. That should be easy enough to do, and will allow clients to work ahead pretty far without hitting the limit.

edfelten:
Closing this as no longer needed. Protocol updates will address the problem instead.

PR #211, opened 2020-02-11
Update chain param setting doc for consistency with launcher UI

ewfelten:


docs/Chain_parameters.md
```diff
@@ -11,36 +11,40 @@ Note: If you run the validator (`arb-validator` or `arb_deploy.py`) without any
...
-    but it can also slow down confirmation of transactions.
--   speed limit: This is the maximum amount of ArbGas consumption that can occur per second of real time.
-    Making this bigger lets the contracts in your ArbChain run faster,
-    but making it too big risks overwhelming validators who can't keep up.
-
-## Recommended parameters for production testing
+-   speed limit: This controls how fast computations on the chain can go, to make sure that every validator can keep up with 
+    computations. This is essentially the speed of the slowest validator you expect to have.  The speed is scaled so that 1.0
+    is the speed of a typical developer laptop. For example, if you are using half of a laptop's capacity to run the
+    validator, you should set the speed limit to 0.5.
```
hkalodner: Eventually we'll need to do something more precise, but this works for now.

docs/Chain_parameters.md
```diff
@@ -11,36 +11,40 @@ Note: If you run the validator (`arb-validator` or `arb_deploy.py`) without any
...
 
 For executing in a production-like setting, we suggest the following parameters:
 
--   stake requirement: 2% of total value at stake in your ArbChain
+-   stake requirement: 1 Eth
 -   grace period: 180 minutes
--   max execution steps: 1,000,000,000
--   speed limit: 100,000,000 ArbGas per second
+-   max assertion size: 50 seconds
+-   max time bounds: 20 blocks
```
hkalodner: Speed limit appears to be missing here.

docs/Chain_parameters.md
```diff
@@ -11,36 +11,40 @@ Note: If you run the validator (`arb-validator` or `arb_deploy.py`) without any
...
-## Recommended parameters for debugging
-
 If you're debugging your dapp, you care more about fast turnaround than security, so we suggest the following parameters:
 
 -   stake requirement: 0.1 Eth
 -   grace period: 10 minutes
--   max execution steps: 1,000,000,000
--   speed limit: 20,000,000 ArbGas per second
+-   speed limit: 0.2
+-   max assertion size: 250 seconds
```
hkalodner: Wouldn't we want smaller assertions and higher speed limit for debugging?

docs/Chain_parameters.md
```diff
@@ -11,36 +11,40 @@ Note: If you run the validator (`arb-validator` or `arb_deploy.py`) without any
...
+## Recommended parameters 
 
 For executing in a production-like setting, we suggest the following parameters:
 
--   stake requirement: 2% of total value at stake in your ArbChain
+-   stake requirement: 1 Eth
 -   grace period: 180 minutes
--   max execution steps: 1,000,000,000
--   speed limit: 100,000,000 ArbGas per second
+-   max assertion size: 50 seconds
```
julientregoat: should we add speed limit: 1.0 here as well for consistency?

docs/Chain_parameters.md
```diff
@@ -11,36 +11,40 @@ Note: If you run the validator (`arb-validator` or `arb_deploy.py`) without any
...
+    running at 100% capacity, the updated state of the chain will be posted to the Ethereum chain this often.
+-   max time bounds width: This is how much time an assertion has to be accepted on-chain, before it is discarded. It is also
+    the maximum clock lag that can exist between the chain's clock (as returned by `block.number` in Solidity) and the block
+    number on the main Ethereum chain.
+
+## Recommended parameters 
 
 For executing in a production-like setting, we suggest the following parameters:
 
--   stake requirement: 2% of total value at stake in your ArbChain
```
julientregoat: is this more accurate? maybe we should keep this and just say we default to 1 in the deployment tool?

docs/Chain_parameters.md
ewfelten: Fixed.

docs/Chain_parameters.md
ewfelten: I reduced the assertion size to 15 seconds.  

I think we want a lower speed limit because we're more likely to be running multiple validators on the same machine, when we're doing local testing.

hkalodner: Makes sense

docs/Chain_parameters.md
ewfelten: I'll add this back in as commentary, that this is what we expect to recommend for production, eventually.

PR #212, opened 2020-02-11
Message batching

hkalodner:
* Single tx unbatched: 40546 gas
* 100 batched txes: 20,823 gas
* 100 batched txes without sig verification: 12,409 gas


packages/arb-bridge-eth/test/arbRollup.js
```diff
@@ -216,18 +221,26 @@ class Assertion {
   }
 
   deadline() {
-    return 13000 * this.blockNumber + grace_period_ticks;
+    return 1000 * this.blockNumber + grace_period_ticks;
```
ewfelten: I'm not sure what this is for, but it's not the formula that the EthBridge uses to compute challenge deadlines.

packages/arb-validator/ethbridge/globalInbox.go
```diff
@@ -63,6 +65,44 @@ func (con *globalInbox) SendTransactionMessage(ctx context.Context, data []byte,
 	return con.waitForReceipt(ctx, tx, "SendTransactionMessage")
 }
 
+func (con *globalInbox) DeliverTransactionBatch(
+	ctx context.Context,
+	chain common.Address,
+	transactions []message.Transaction,
+	signatures [][65]byte,
+) error {
+	con.auth.Lock()
```
ewfelten: Could move the lock acquisition down to just before the DeliverTransactionBatch call, so the preparation for the call is done without holding the lock.

hkalodner: Done

packages/arb-bridge-eth/test/arbRollup.js
hkalodner: I changed it to be closer to correct and added a comment. Maybe at some point it'll be worth doing more, but the only use case of this is getting transactions accepted in the ethbridge for gas cost analysis which this achieves.

Issue #228, opened 2020-02-13
EthBridge get docker image should seed accounts with a reasonable amount of funds


hkalodner:
Currently the geth node seeds wallets with a massive amount of money which breaks UX and makes it hard to notice changes in balance. This should be reduced to a more reasonable amount of ETH.

PR #229, opened 2020-02-13
Start Geth Local Node with reasonable balance

hkalodner:
* Funded local blockchain accounts start with 1000 Eth
* arb-deploy funds created validators with 5 ETH


hkalodner:
This fixes #228

PR #244, opened 2020-02-16
update installation docs to clone master

### LINKED ISSUE #242: Installation Error: Easy Fix?
I'm running Linux Mint 19.2 "Tina" with Cinnamon as my desktop environment.

I ran into some difficulty following the installation guide here:
https://developer.offchainlabs.com/docs/Installation/

I encountered an error when running this code from the above link:

    git clone -b v0.4.2 --depth=1 -c advice.detachedHead=false https://github.com/offchainlabs/arbitrum.git
    cd arbitrum
    yarn
    yarn build
    yarn install:deps

Specifically after running `yarn install:deps`:

    ./scripts/install-deps: line 16: pre-commit: command not found
    error Command failed with exit code 127.

After snooping around the repo I downloaded and the master branch of arbitrum, I found this difference in the scripts.

(on my filesystem) `./arbitrum/scripts/install-deps`:

    15  # Setup pre-commit
    16  pre-commit install

(on the master branch) `https://github.com/offchainlabs/arbitrum/scripts/install-deps`:

    15  # Setup pre-commit
    16  python3 -m pre_commit install

The second one works. The first one does not, which is the error I encountered following the installation guide. I found a branch matching the code I downloaded, which I'm assuming correlates to v0.4.2:
https://github.com/OffchainLabs/arbitrum/tree/version-bump/scripts/install-deps

And I suppose that's the file that needs to be changed, or else maybe the original `git clone` command on the installation page needs to be updated to a more recent version, if it has already been fixed. I hope this is helpful!

julientregoat:
update the branch used in the installation docs. for the short term, might be more effective to just clone `master` so we don't have to update the docs while we make frequent updates. alternatively, perhaps we can adjust the script to grab the latest tag and clone that

fixes #242

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -68,7 +68,7 @@ Here are the important dependencies in case you are not running on a supported O
 Download the Arbitrum Monorepo from source:
 
 ```bash
-git clone -b v0.4.2 --depth=1 -c advice.detachedHead=false https://github.com/offchainlabs/arbitrum.git
+git clone -b master --depth=1 -c advice.detachedHead=false https://github.com/offchainlabs/arbitrum.git
 cd arbitrum
 yarn
 yarn build


hkalodner:
This is an interesting question. I'd like to be able to do this, but it would require making sure that master is always stable. @edfelten, what do you think?

hkalodner:
I think we're sticking with explicitly tagging in the docs to avoid the risk of instability

PR #245, opened 2020-02-19
use BigNumber for HashOnlyValue

julientregoat:


hkalodner:
I'm guessing that this supersedes #238?

hkalodner:
Actually it looks like this only includes the HashOnlyValue change and removing node-fetch from the yarn.lock, but not from package.json

hkalodner:
#238 is now merged

julientregoat:
should be gtg

hkalodner:
Looks like one of the tests has to be upgraded to match the change

PR #248, opened 2020-02-28
Fix update-abi to put its output in the right place

edfelten:
Fix update-abi script to put its output into arb-validator-core instead of arb-validator. It should have been updated when arb-validator-core was refactored out of the old arb-validator.


hkalodner:
I'm a bit confused since I remember updating abi-gen, but it I clearly missed actually adding it to the PR or something since master is broken.

PR #276, opened 2020-04-03
add install virtualenv to installation instructions

DZGoldman:


julientregoat:
`virtualenv` should be getting installed with the `yarn install:deps` command which happens further on. was it required earlier or did it not install properly with that script?

DZGoldman:
Yep yep, I was still on the 0.4.2 version. closing

PR #293, opened 2020-04-23
Efficient Assertion Tracing

hkalodner:
- Add the newly created valid node hash to the `RollupAsserted` event so that simple clients can track the progress of that assertion.
- Add `ConfirmedValidAssertion` to `MessageResult` in `arb-provider-ethers`


hkalodner:
This change requires an EthBridge update and the branch includes an experimental version number. Before merging, this should be updated.

PR #294, opened 2020-04-23
Clean up type-only exports; update formatting + linting usage

julientregoat:
see commits for more details


packages/arb-provider-ethers/scripts/builder
```diff
@@ -1,8 +1,14 @@
 rm -r ./dist
 
-yarn run tsc
+tsc
```
hkalodner, line 3: Will this work if typescript isn't installed globally? If so do we require a global installation?

hkalodner: On second thought, I guess this is safe since `builder` is only run from the `package.json`?

packages/arb-provider-ethers/src/lib/abi/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  T extends Pick<EventDescription, 'encodeTopics'>
+> extends EventDescription {
+  encodeTopics: T['encodeTopics']
+}
+
+export interface TypedFunctionDescription<
+  T extends Pick<FunctionDescription, 'encode'>
+> extends FunctionDescription {
+  encode: T['encode']
+}
```
julientregoat, line 25: forgot to explain this. I copied this from the typechain generated `index.d.ts` file. Since that file is generated, it can be overwritten at anytime. 

When building this project with `tsc`, the `index.ts` file takes precedence over the `.d.ts` file, so the types aren't copied into the build folder (not that they would be copied over without the presence of `index.ts` - typescript doesn't emit anything for declaration only files).

this is the simplest way of having a directory index while also maintaining these types (which are depended on by the rest of the package)

packages/arb-provider-ethers/scripts/builder
julientregoat: yep, it works without a global installation, I don't have one atm. though now I'm wondering if that's the reason that we used `yarn run` in the first place - to ensure the global installation isn't used. if the presence of a global installation is an issue, maybe this could be replaced with `npm run tsc` instead for the behavior we're looking for

julientregoat:
external dependencies should only fail if they're updated though, right? the `ethProvider` change is definitely breaking but we haven't been following semver so thinking this is good?

packages/arb-provider-ethers/scripts/builder
julientregoat: just tested it, doesn't conflict with global installations either

hkalodner: Awesome

packages/arb-provider-ethers/src/lib/abi/index.ts
hkalodner: Ah, I missed this going through. That makes sense and seems like a good approach.

hkalodner:
Hmm. I think we might need to do a minor version bump. `arb-provider-web3` uses `"arb-provider-ethers": "^0.4.3"` for instance.

julientregoat:
> Hmm. I think we might need to do a minor version bump. `arb-provider-web3` uses `"arb-provider-ethers": "^0.4.3"` for instance.

yeah that would prevent any rogue npm updates. might also be good to pin all versions used internally?

hkalodner:
Hmm. I hadn't really thought about exact pinning internals before, but I think you're right. The only tricky part will be properly keeping them in sync when updating, but that should work alright.

PR #332, opened 2020-05-15
added in a higher gas limit

JesseAbram:
I noticed an issue while trying to launch our contracts. It seems like   options.allowUnlimitedContractSize = true should have solved it, however I do get block size limit when deploying

image

Setting the gas limit to 10 million which is pretty close to the mainnet limit solved it


hkalodner:
Ah actually you've angered our formatting guidelines 😁 

I think you just need to drop the semicolon from the end of the line

JesseAbram:
hahah that is actually funny, I did this originally in the library which puts the semi colons in. So I wrote it without a semi colon and was like oh I guess their linting standards are to add this, then I just copied it over lol. Anyways fixed

PR #344, opened 2020-06-04
[WIP] Add Reorg and Crash Restore support to transaction result server

hkalodner:
This PR is feature complete, but currently lacks testing. I'm submitting this so that we can jump into code review while I work on adding tests.

Major features:

- Add `RestartingFromLatestValid` to listener interface to notify listeners of a reorg (or initial launch)
- Consolidate `AdvancedCalculatedValidNode` and `AdvancedKnownAssertion` into `AdvancedKnownNode` which is called every time the opinion thread moves the latest calculated valid node
- Add reorg support to txtracker server. Previously reorgs were ignored, leading to the validator potentially serving false receipts. Now the server deletes information that came from nodes which have been removed
- Add support for returning results from past nodes. The old system only served data from nodes which had been newly added after the validator launched. Now the txserver will be able to serve receipts that occurred any time in the rollup's history.
- Optimize log query efficiency by store bloom filters along with nodes.
- Improve stake moving code in `AdvancedKnownNode` method of `ValidatorChainListener` in response to listener API change as well as noticing pre-existing issues in the code.

TODO:
- Add tests. The tx tracker itself is fairly hard to test since we currently have no facility for mocking the rollup observer and so this can only be easily tested in integration testing. However the `txtrackerdb` component is cleanly separable and thus should be testable.


hkalodner:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 101
- Added 166
           

Complexity increasing per file
==============================
- packages/arb-validator/ckptcontext/context.go  5
- packages/arb-validator/structures/node.go  6
- packages/arb-avm-cpp/cmachine/nodeStore_test.go  16
- packages/arb-validator/ckptcontext/ckptcontext.pb.go  7
- packages/arb-avm-cpp/cmachine/blockStore.go  3
- packages/arb-validator/rollupvalidator/txtracker.pb.go  7
- packages/arb-validator/nodeview/nodeView.go  3
- packages/arb-validator/rollup/rollup_test.go  1
- packages/arb-avm-cpp/cmachine/nodeStore.go  2
         

Complexity decreasing per file
==============================
+ packages/arb-validator/rollupvalidator/txtracker.go  -7
+ packages/arb-validator/rollupmanager/manager.go  -12
+ packages/arb-validator/rollup/opinionThread.go  -2
+ packages/arb-validator/checkpointing/indexedCheckpointer.go  -1
+ packages/arb-validator/rollup/rollup.go  -1
         

Clones added
============
- packages/arb-validator/rollup/stakerSet.go  2
- packages/arb-avm-cpp/data_storage/src/nodestore.cpp  9
- packages/arb-validator/ckptcontext/ckptcontext.pb.go  5
- packages/arb-validator/rollupvalidator/txtracker.pb.go  5
- packages/arb-validator/checkpointing/indexedCheckpointer_test.go  3
         

Clones removed
==============
+ packages/arb-avm-cpp/cavm/ccheckpointstorage.cpp  -6
+ packages/arb-avm-cpp/avm/src/machinestate/datastack.cpp  -1
+ packages/arb-avm-cpp/avm/src/machinestate/machinestate.cpp  -16
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/OffchainLabs/arbitrum/pullRequest?prid=5639569&bid=18468138)

packages/arb-validator/rollupmanager/manager.go
```diff
@@ -79,151 +83,107 @@ func CreateManagerAdvanced(
...
-			chain.Start(runCtx)
+			man.activeChain.Start(runCtx)
 
 			current, err := clnt.CurrentBlockId(runCtx)
 			if err != nil {
 				log.Fatal(err)
 			}
 
-			headersChan, err := clnt.SubscribeBlockHeaders(runCtx, chain.CurrentBlockId())
+			headersChan, err := clnt.SubscribeBlockHeaders(runCtx, man.activeChain.CurrentBlockId())
```
edfelten, line 137: Wondering if there is a race condition here. We called RestartFromLatestValid and then Start on man.activeChain above. Now we get CurrentBlockId() from activeChain. Is there a chance that CurrentBlockId changed in the meantime, in a way that causes us to miss some events?

edfelten: And we use the variable "current" in this code as well, as what appears to be a separate notion of what the current block is.

edfelten: If these are indeed different, recommend renaming and/or commenting for clarity.

packages/arb-validator/rollupvalidator/txtracker.go
```diff
@@ -29,72 +37,90 @@ import (
...
+	Logs    []evm.Log
+	TxIndex uint64
+	TxHash  common.Hash
 }
 
-type txRequest struct {
-	txHash     common.Hash
-	resultChan chan<- txInfo
+type nodeInfo struct {
+	TxLogs            []logsInfo
```
edfelten: Suggest adding comments to explain the difference between TxLogs and OutLogs, and which things are covered by LogsAccHashes and LogsValHashes and TransactionHashes.

packages/arb-validator/rollupvalidator/txtracker.go
```diff
@@ -110,214 +136,290 @@ func (a *assertionInfo) FindLogs(address *big.Int, topics []common.Hash) []logRe
...
 type txTracker struct {
-	txRequestIndex int
-	transactions   map[common.Hash]txInfo
-	assertionInfo  []*assertionInfo
-	assertionMap   map[common.Hash]*assertionInfo
-	accountNonces  map[common.Address]uint64
-	vmID           common.Address
-	requests       chan validatorRequest
+	sync.RWMutex
+	rollup.NoopListener
```
edfelten, line 138: Why use a NoopListener?  It looks like it doesn't do anything.

packages/arb-validator/rollupvalidator/txtracker.go
```diff
@@ -110,214 +136,290 @@ func (a *assertionInfo) FindLogs(address *big.Int, topics []common.Hash) []logRe
...
+		defer tr.Unlock()
+
+		if err := tr.txDB.confirmNode(ev.NodeHash); err != nil {
+			log.Println(err)
+			return
+		}
+	}()
+}
+
+func (tr *txTracker) processNextNode(node *structures.Node) {
```
edfelten, line 196: Is there some locking needed here?  We could have multiple instances to processNextNode running at the same time, and other stuff happening simultaneously as well.

Also concerned about race conditions between this async thread and reorgs.

packages/arb-validator/rollupvalidator/txtrackerdb.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	txdb.deleteNode(node)
+	return nil
+}
+
+func (txdb *txDB) deleteNode(node *nodeInfo) {
+	for _, txHash := range node.TransactionHashes {
+		delete(txdb.transactions, txHash)
+	}
+	delete(txdb.nodeHeightLookup, node.NodeHeight)
```
edfelten, line 107: Seems like there are race conditions here, unless protected by some external locking. Deletions here might race against activities that set the same values for the post-reorg world.

hkalodner: So the intention here is for the `txDB` to not be thread safe and to instead have the `txTracker` use a readwrite mutex lock when accessing it. On thing that might help would be to comment on each method what type of lock it needs.

packages/arb-validator/rollupvalidator/txtracker.go
hkalodner: There's a write lock that gets held by this function after it finishes processing the node which is a pure function so I think this is safe as is

packages/arb-validator/rollupvalidator/txtracker.go
hkalodner: It's because the Listener interface has a large number of different methods that must be implemented, but this class only cares about a few of them. This way we don't need to include noop implementations of all the other interface requirements.

packages/arb-validator/rollupvalidator/txtracker.go
edfelten: My bad.  There was locking there.

packages/arb-validator/rollupmanager/manager.go
```diff
@@ -39,26 +42,27 @@ const (
...
-	RollupAddress common.Address
-	client        arbbridge.ArbClient
-	listeners     []rollup.ChainListener
-
-	listenerAddChan chan rollup.ChainListener
-	actionChan      chan func(*rollup.ChainObserver)
-	ckpFac          checkpointing.RollupCheckpointerFactory
+	// listenersLock is locked when writing listeners or activeChain
+	listenersLock sync.Mutex
+	// validCallLock is locked when there is not a valid chain caught up to head
```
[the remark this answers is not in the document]
hkalodner: Added more detailed comments

packages/arb-validator/rollupmanager/manager.go
```diff
@@ -79,151 +83,107 @@ func CreateManagerAdvanced(
...
+						break headerLoop
 					}
-				case action := <-man.actionChan:
-					action(chain)
 				}
 			}
 
+			man.validCallLock.Lock()
+
+			man.listenersLock.Lock()
```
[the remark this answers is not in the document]
hkalodner: I cleaned this up

packages/arb-validator/rollupmanager/manager.go
hkalodner: Code has been refactored and variable names changed to increase understandability

packages/arb-validator/rollupvalidator/txtracker.go
hkalodner: Updated variable names and added comments with further explanation

packages/arb-validator/rollupvalidator/txtrackerdb.go
hkalodner: Added comments to make this more clear

packages/arb-validator/rollup/listener.go
```diff
@@ -38,6 +37,8 @@ const (
 )
 
 type ChainListener interface {
+	RestartingFromLatestValid(context.Context, *ChainObserver, *structures.Node)
```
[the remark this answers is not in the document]
hkalodner: Added comment

PR #373, opened 2020-06-17
Refactor checkpointer and fix memory usage

hkalodner:
- Fix explosive memory usage in validator - Issue was due to repeated copying of code when machines were cloned which is supposed to be a cheap operation
- Reorganize checkpointer in order to simplify public API and unite related pieces of code
- Don't parse assertion values in validator - We actually only need the values marshaled.


packages/arb-avm-cpp/avm/include/avm/machine.hpp
```diff
@@ -57,10 +71,9 @@ class Machine {
 
     TuplePool& getPool() { return *machine_state.pool; }
 
-    SaveResults checkpoint(CheckpointStorage& storage);
-    bool restoreCheckpoint(const CheckpointStorage& storage,
-                           const std::vector<unsigned char>& checkpoint_key);
-    DeleteResults deleteCheckpoint(CheckpointStorage& storage);
+    void marshal_value(const value& val, std::vector<unsigned char>& buf) {
```
minhptruong, line 74: curious why marshal_value is in the machine interface? seems like it you should be able to do it without a machine

hkalodner: That's basically a compromise I made. The value type has CodePointRef whereas the marshal function produces the marshaling of a CodePoint. I put marshal value in the machine interface since it's linked to the code of the machine. In the future we can examine whether or not we could propagate the CodePointRef idea to the marshaling format as well.

This is a super thin wrapper around the global marshal_value function since it just calls `::marshal_value(val, buf, machine_state.static_values->code)`

packages/arb-avm-cpp/avm/include/avm/avmfwd.hpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2019-2020, Offchain Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
```
minhptruong, line 6: maybe a better name? not super apparent what the class is supposed to be.

hkalodner: Good call. I'm gonna take a look

packages/arb-avm-cpp/data_storage/include/data_storage/checkpointstorage.hpp
```diff
@@ -33,20 +36,23 @@ class TransactionDB;
 class CheckpointStorage {
```
minhptruong, line 36: once codes become mutable, is the plan to have to make new storages and keep track of older ones as well?

packages/arb-avm-cpp/avm/include/avm/avmfwd.hpp
hkalodner: Changed this over to be called `status.hpp`

packages/arb-avm-cpp/data_storage/include/data_storage/checkpointstorage.hpp
hkalodner: That's a great question. My current thought is that we'll use shared memory and mutexes to handle things. It's made a bit easier by the fact that code is immutable, it can just be added to. Potentially we'll want machines that are adding to their code to commit to the shared store when they get cloned or checkpointed. This is definitely something important to figure out.

Given that the approach this replaces is basically infeasible memory-wise I think it makes sense to do this now and then figure out how to handle mutable code when we get there.

PR #437, opened 2020-07-16
Update README with setup instructions

fredlacs:
Started a draft PR to add local setup instructions to the README.


hkalodner:
We've moved back to using the docker setup so this is no longer necessary

PR #487, opened 2020-08-11
add withdraw eth bridge-test

DZGoldman:
Run it like `yarn run test_bridge 0xArbChainAddress` inside `arb-provider-ethers`


DZGoldman:
yeah for sure  (too early to merge anyway atm)

Issue #505, opened 2020-08-24
number of arguments mismatch with depositERC20Message & depositERC721Message


DZGoldman:
Both methods currently throw an error: they're called with a 5th `overrides` parameter in `l1Bridge.ts` which is missing in the implementations (in `GlobalInbox.sol`).

hkalodner:
I don't understand why this doesn't work with the autogenerated typechain interface https://github.com/OffchainLabs/arbitrum/blob/d4dc5a2c596880614d41119a13b90c09922d08de/packages/arb-provider-ethers/src/lib/abi/GlobalInbox.d.ts#L287

hkalodner:
This is now fixed

PR #506, opened 2020-08-24
default to empty object instead of undefined for depositERC20/721Message

DZGoldman:
fixes https://github.com/OffchainLabs/arbitrum/issues/505

Still not at all clear why this is necessary, but this works as a temporary workaround


DZGoldman:
update: this is, indeed, the proper fix; overrides should default to an empty object, never undefined

DZGoldman:
(rebasing)

PR #546, opened 2020-09-03
update contract deployment docs

DZGoldman:


docs/Contract_Deployment.md
```diff
@@ -8,57 +8,46 @@ Arbitrum supports standard EVM contract deployment. This allows standard Solidit
...
     ```js
-    const ethers = require('ethers')
-    const ArbEth = require('arb-provider-ethers')
-    const ProviderBridge = require('arb-ethers-web3-bridge')
+    const wrapProvider = require('arb-ethers-web3-bridge').wrapProvider
+    const HDWalletProvider = require('@truffle/hdwallet-provider')
 
     const mnemonic =
       'jar deny prosper gasp flush glass core corn alarm treat leg smart'
+    const arbProviderUrl = 'http://localhost:1234/'
```
hkalodner, line 31: Probably better to have 8547 here since that's what we've been using

hkalodner (approved):
LGTM

PR #548, opened 2020-09-03
add lockbox and aggregator to glossary (and modify validator)

DZGoldman:
Thinking we try to strictly stick to validator = asserter in our language; thoughts welcome.


docs/Glossary.md
```diff
@@ -3,7 +3,7 @@ id: Glossary
 title: Glossary of Arbitrum Terms
 custom_edit_url: https://github.com/OffchainLabs/arbitrum/edit/master/docs/Glossary.md
 ---
-
+- Aggregator: A party who keeps track of the state of an Arbitrum chain, receives remote procedure calls (RPCs) from clients, and submits batches of transactions on bahalf of users; analogous to a non-mining L1 Ethereum node.
```
hkalodner, line 6: Typo "bahalf"

docs/Glossary.md
```diff
@@ -14,6 +14,7 @@ custom_edit_url: https://github.com/OffchainLabs/arbitrum/edit/master/docs/Gloss
...
 - EthBridge: A group of contracts running on the Ethereum chain, which act as a record-keeper and rule enforcer for Arbitrum chains.
 - Inbox: Holds a sequence of messages sent by clients to the contracts on an Arbitrum Chain. Every chain's inbox is managed by the on-chain EthBridge. Every message to a chain is timestamped with the Ethereum block number when it was put into the chain's inbox.
+- Lockbox: A contract which holds assets (ETH, ERC20 tokens, or ERC721 tokens) a user has withdrawn from the Arbitrum chain, available to the user after their withdrawal is successfully confirmed.
 - Staker: A party who deposits a stake, in Eth, to vouch for a particular node in an Arbitrum Chain. A party who stakes on a false node can expect to lose their stake. An honest staker can recover their stake once the node they are staked on has been confirmed.
-- Validator: A party who keeps track of the state of an Arbitrum Chain, and possibly acts to ensure the chain makes progress correctly.
+- Validator: A party who makes staked, disputable assertions about the state of the Arbitrum chain; i.e., proposing a state update or challenging another valaditor's assertion.
```
hkalodner, line 19: I think that we also want to capture the idea of a passive validator who isn't asserting, but is actively watching assertions and will challenge if one is false

hkalodner (changes requested):

docs/Glossary.md
DZGoldman: watchu think

hkalodner: I like it

PR #559, opened 2020-09-08
Fix block hash

hkalodner:
- Update the behavior of eth_getBlockByHash and eth_getBlockByNumber to use the same hash and parent hash as Ethereum. The previous behavior led to block hashes (which were calculated after modifying some header fields) not matching parent hashes.

Note that at some point we may run into an issue from our block hashes not matching the contents of our blocks served via RPC, but that issue already existed and is no worse with this PR. Also it appears that clients take the hash from the RPC result rather than recalculating it

### DIFF
--- a/packages/arb-tx-aggregator/aggregator/aggregator.go
+++ b/packages/arb-tx-aggregator/aggregator/aggregator.go
@@ -41,7 +41,7 @@ import (
 )
 
 type Server struct {
-	client      ethutils.EthClient
+	Client      ethutils.EthClient
 	chain       common.Address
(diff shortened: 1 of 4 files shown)


hkalodner:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 3
- Added 4
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/OffchainLabs/arbitrum/pullRequest?prid=6144146&bid=20083844)

edfelten (approved):
Looks good.

PR #579, opened 2020-09-30
Fix R and S encoding in RPC interface

hkalodner:
- Make sure R and S values are 32 bytes when returning transaction data


fredlacs (approved):
LGTM

PR #582, opened 2020-10-01
Update blockhash

hkalodner:
- Update ArbOS to https://github.com/OffchainLabs/arb-os/pull/205
- Add test for expected behavior of EVM BLOCKHASH opcode (`return abi.encode(uint256(0))`)


edfelten (approved):
Looks good.

PR #584, opened 2020-10-04
Only process tuple if actually tuple

joshuacolvin0:
`assembleValueFromDeserialized` was trying to parse each item as `TuplePlaceholder` if it was not a `value`.  Changed so that only `TuplePlaceholder` is parsed as `TuplePlaceholder`.


hkalodner (approved):
LGTM, no changes in functionality, but I think this makes it easier to understand

PR #588, opened 2020-10-08
Update ethereumjs-abi version

joshuacolvin0:
eth-sig-util wants to use `latest` version of ethereumjs-abi, but this causes problems with other packages that use a specific version.  The other packages have now updated to a new version, so need to update `resolutions` so that eth-sig-util stays in sync.


hkalodner (approved):
LGTM

PR #591, opened 2020-10-12
Fix stake recovery funcs

hkalodner:
Fix stake recovery functions


edfelten (approved):
Looks good. Glad to see this bug fixed.

PR #615, opened 2020-10-26
Fix checkpointer atomicity

hkalodner:
- Use GetForUpdate instead of Get to make rocksdb transactions atomic


joshuacolvin0 (approved):
LGTM

PR #623, opened 2020-10-29
add address index getter / argument byte serialization into arb-provider-ethers

DZGoldman:
<del>fyi, the actual byte-serialization still lives in the arbiswap repo: https://github.com/OffchainLabs/Arbiswap_V2_mono/blob/byte-serialization/packages/uniswap-sdk/src/byteSerializeParams.ts</del>
<del>
Will move it to APE eventually / post-uniswap (there's some considerations about the different version of ethers-js so I'd rather just wait).  </dev>


DZGoldman:
Byte serialization stuff is on here; also has the typechain updates for the new interfaces.

DZGoldman:
@hkalodner bump :)

packages/arb-provider-ethers/src/lib/l2bridge.ts
```diff
@@ -34,3 +36,30 @@ export async function withdrawEth(
     value,
   })
 }
+
+interface AddressIndexMemo {
+  [address: string]: number
+}
+
+export const getAddressIndex = (() => {
```
hkalodner, line 44: Took a minute for me to figure out how the closure worked with the `arbSys` var, but this is super smart

packages/arb-provider-ethers/src/lib/byte_serialize_params.ts
```diff
@@ -0,0 +1,100 @@
...
+import { getAddressIndex } from './l2bridge'
+import { JsonRpcProvider } from 'ethers/providers'
+
+type PrimativeType = string | number | boolean | BigNumber
+type PrimativeOrPrimativeArray = PrimativeType | PrimativeType[]
+type BytesNumber = 1 | 4 | 8 | 16 | 32
+
+// to use:
+// const mySerializeParamsFunction = initSerializeAndLookUpIndices("myAggregatorUrl")
+export const initSerializeAndLookUpIndices = (arbAggregatorUrl: string) => {
```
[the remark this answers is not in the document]
hkalodner: Also what's the "init" part of this function? Seems a bit confusing

packages/arb-provider-ethers/src/lib/byte_serialize_params.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const serializeParams = async (
+  params: PrimativeOrPrimativeArray[],
+  addressToIndex: (address: string) => Promise<number> = () =>
+    new Promise(exec => exec(-1))
+) => {
+  const formattedParams: string[] = []
+
+  for (let i = 0; i < params.length; i++) {
+    const param = params[i]
```
hkalodner, line 53: Could we do a `for of` loop here?

packages/arb-provider-ethers/src/lib/byte_serialize_params.ts
```diff
@@ -0,0 +1,100 @@
...
+  ) {
+    return toUint(value, 32)
+  } else {
+    throw new Error('unsupported type')
+  }
+}
+
+export const serializeParams = async (
+  params: PrimativeOrPrimativeArray[],
+  addressToIndex: (address: string) => Promise<number> = () =>
```
hkalodner, line 47: I see what you did here for mocking. This is cool 😁

hkalodner (changes requested):
Is there a reason that `token-mock` was added as a submodule?

packages/arb-provider-ethers/src/lib/byte_serialize_params.ts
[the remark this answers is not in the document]
DZGoldman: ...yeah, the idea is to "initialize" (maybe "createSerializer" would have been a better name once) the function w/ the aggregator-url (or aggregator) once, and then have the returned function just store it as a closure; it's a bit of a roundabout way to avoid to declare a class. Admittedly, looking w/ fresh eyes, the pattern does feel a bit convoluted

packages/arb-provider-ethers/src/lib/byte_serialize_params.ts
DZGoldman: ye

hkalodner (approved):
LGTM

PR #624, opened 2020-10-29
Aggregator redirect mode

hkalodner:
This PR adds a `--forward-url=[URL]` parameter to the aggregator which enables it to forward transactions to another aggregator rather that submitting them directly to the chain

If this parameter is used, the aggregator doesn't need to be funded or have a wallet


joshuacolvin0 (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

PR #630, opened 2020-11-12
Update receipt

hkalodner:
- Serve transaction data about buddy deployments and contract transactions
- Add `arbType` and optional `arbSubType` fields to the RPC transaction object
- Set `transactionsRoot` in block result based on transactions included in the block
- Add arbos constructor test


joshuacolvin0 (approved):
LGTM

PR #631, opened 2020-11-13
Remove arb provider go

hkalodner:
Now that we support pre-EIP155 signature, we can use the standard go-ethereum abigen classes directly with the standard ethereum client instead of using our custom `arb-provider-go` provider

- This PR removes `arb-provider-go`
- Update the fibgo integration test to use standard ethereum tooling
- Remove old RPC server that `arg-provider-go` depended one


joshuacolvin0 (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

PR #632, opened 2020-11-13
Ignore duplicate tx receipts

hkalodner:
It's possible for the same transaction to be submitted to the arbitrum chain multiple times. Currently when this occurs the aggregator responds to requests for the receipt associated with the given request id by returning the most recently output log matching that request. In fact, the aggregator should return the version of the request that had a correct sequence number if one exists.

This PR adjusts the behavior of the aggregator so only save a request that had an invalid sequence number, if no request matching that ID already exists.


joshuacolvin0 (approved):
LGTM

PR #634, opened 2020-11-13
Fix batcher

hkalodner:
This PR shifts the behavior of the aggregator to make it quit (triggering a restart if running in Kubernetes) if the connected L1 node rejects its transaction rather than the previous behavior of staying up, but refusing user transactions


joshuacolvin0 (approved):
LGTM

PR #636, opened 2020-11-13
Fix GetTransactionByHash

hkalodner:
Currently `eth_getTransactionByHash` returned an error if the transaction was not found. This PR adjusts the behavior to return null instead following the web3 JSON-RPC spec


joshuacolvin0 (approved):
LGTM

PR #638, opened 2020-11-16
Rpc fixes

hkalodner:
- Add `from` and `to` field to transaction receipt
- Make `getBlock` RPC calls return nil if L2 block has not been processed yet


joshuacolvin0 (approved):
LGTM

PR #639, opened 2020-11-17
Add PUB/SUB support

joshuacolvin0:
Add NewLogs, Chain, ChainHead and NewTx subscription support

### DIFF
--- a/packages/arb-tx-aggregator/batcher/batcher.go
+++ b/packages/arb-tx-aggregator/batcher/batcher.go
@@ -241,8 +241,6 @@ func (m *Batcher) sendBatch(ctx context.Context, inbox arbbridge.GlobalInboxSend
 		return
 	}
 
-	m.newTxFeed.Send(core.NewTxsEvent{Txs: txes})
-
 	m.pendingBatch = m.pendingBatch.newFromExisting()
 	m.pendingSentBatches.PushBack(&pendingSentBatch{
 		txHash: txHash,
@@ -276,6 +274,8 @@ func (m *Batcher) SendTransaction(_ context.Context, tx *types.Transaction) erro
 		log.Err(err).Msg("error processing user transaction")
 		return err
 	}
+
+	m.newTxFeed.Send(core.NewTxsEvent{Txs: []*types.Transaction{tx}})
 
 	txJSON, err := tx.MarshalJSON()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


packages/arb-tx-aggregator/txdb/txdb.go
```diff
@@ -152,6 +162,32 @@ func (db *TxDB) restoreFromCheckpoint(ctx context.Context) error {
...
+			return err
+		}
+		for _, evmLog := range txRes.EVMLogs {
+			oldEthLogs = append(oldEthLogs, &types.Log{
+				Address: evmLog.Address.ToEthAddress(),
+				Topics:  common.NewEthHashesFromHashes(evmLog.Topics),
+			})
+		}
+	}
+	if len(oldEthLogs) > 0 {
```
hkalodner, line 187: Can we use `GetMachineBlockResults` instead of this code?

packages/arb-tx-aggregator/txdb/txdb.go
```diff
@@ -171,6 +207,9 @@ func (db *TxDB) restoreFromCheckpoint(ctx context.Context) error {
 
 func (db *TxDB) AddMessages(ctx context.Context, msgs []arbbridge.MessageDeliveredEvent, finishedBlock *common.BlockId) error {
 	timestamp, err := db.timeGetter.TimestampForBlockHash(ctx, finishedBlock.HeaderHash)
+	db.blockProcFeed.Send(true)
+	defer db.blockProcFeed.Send(false)
```
hkalodner, line 211: Nice use of defer 😁

packages/arb-tx-aggregator/txdb/txdb.go
```diff
@@ -404,6 +491,22 @@ func (db *TxDB) saveAssertion(ctx context.Context, processed processedAssertion)
 			return err
 		}
 
+		ethLogsList, err := db.GetLogs(ctx, block.Hash())
+		if err != nil {
+			return err
+		}
```
hkalodner, line 497: Rather than doing this, I think it'd be better to get the logs from `ethReceipts` defined above

packages/arb-tx-aggregator/batcher/batcher.go
```diff
@@ -236,6 +241,8 @@ func (m *Batcher) sendBatch(ctx context.Context, inbox arbbridge.GlobalInboxSend
 		return
 	}
 
+	m.newTxFeed.Send(core.NewTxsEvent{Txs: txes})
```
hkalodner, line 244: I think the send to this feed should be in the send transaction function rather than here since we want to send to the newTxFeed immediately when we see a transaction and it might be a while before we send a batch

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #640, opened 2020-11-19
Only use a single context.Background()

joshuacolvin0:
If multiple context.Background() objects are used there is a chance that transactions with multiple nonces will be created.  To prevent this from happening, ensure only a single context.Background() is passed throughout the program.


hkalodner (approved):
LGTM

PR #641, opened 2020-11-19
Update block header scheme

hkalodner:
This PR updates the aggregator to serve blocks with their own arbitrum specific block header/hash instead of of former method of using the Ethereum block hash associated with the arbitrum block as our L2 block hash as well


packages/arb-avm-cpp/cmachine/aggregator.go
```diff
@@ -130,22 +137,48 @@ func (as *AggregatorStore) LatestBlock() (*common.BlockId, error) {
 		return nil, errors.New("failed to load block count")
 	}
 	// Last value returned is not an error type
```
joshuacolvin0, line 139: Last value returned is now an error type, so this comment should be removed

joshuacolvin0 (approved):
Other than removing outdated comment, LGTM

joshuacolvin0 (approved):

PR #643, opened 2020-11-19
Optimize docker

hkalodner:
This PR splits up the docker build into small parts, drastically increasing the cached of work that is cached when files change


joshuacolvin0 (approved):
LGTM

PR #646, opened 2020-11-24
TXDB Fixes

hkalodner:
- Save correct block hash for empty block
- Fix handling of incorrect sequence number receipts


joshuacolvin0 (approved):
LGTM

PR #651, opened 2020-11-25
Manage Nonce client-side

joshuacolvin0:
Parity nodes do not count pending transactions when eth_getTransactionCount is called, so to ensure consistent behavior regardless of which type of ethereum node is used, it is easiest to manage Nonce client side.

### DIFF
--- a/packages/arb-validator-core/ethbridge/arbClient.go
+++ b/packages/arb-validator-core/ethbridge/arbClient.go
@@ -202,89 +202,52 @@ func (t *TransactAuth) makeContract(ctx context.Context, contractFunc func(auth 
 	if auth.Nonce == nil {
 		// Not incrementing nonce, so nothing else to do
 		if err != nil {
-			log.Err(err).Str("nonce", "nil").Msg("makeContract")
+			log.Err(err).Str("nonce", "nil").Msg("error when nonce not set")
 			return addr, nil, err
 		}
 
 		txJSON, err := tx.MarshalJSON()
 		if err != nil {
-			log.Err(err).Str("nonce", "nil").Msg("makeContract failed to marshal tx into json")
+			log.Err(err).Str("nonce", "nil").Msg("failed to marshal tx into json")
 			return addr, tx, err
 		}
 
-		log.Info().RawJSON("tx", txJSON).Str("nonce", "nil").Str("sender", t.auth.From.Hex()).Msg("makeContract")
+		log.Info().RawJSON("tx", txJSON).Str("nonce", "nil").Hex("sender", t.auth.From.Bytes()).Msg("make")
 		return addr, nil, err
 	}
 
 	for i := 0; i < smallNonceRepeatCount && err != nil && strings.Contains(err.Error(), smallNonceError); i++ {
 		// Increment nonce and try again
-		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("makeContract incrementing nonce and submitting tx again")
+		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("incrementing nonce and submitting tx again")
 
-		auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
+		t.auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
+		auth.Nonce = t.auth.Nonce
 		addr, tx, _, err = contractFunc(auth)
 	}
 
 	if err != nil {
-		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("makeContract")
+		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("make")
 		return addr, nil, err
 	}
 
 	// Transaction successful, increment nonce for next time
 	txJSON, err := tx.MarshalJSON()
 	if err != nil {
-		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("makeContract failed to marshal tx into json")
+		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("failed to marshal tx into json")
 	} else {
-		log.Info().RawJSON("tx", txJSON).Str("nonce", auth.Nonce.Text(16)).Str("sender", t.auth.From.Hex()).Msg("makeContract")
+		log.Info().RawJSON("tx", txJSON).Str("nonce", auth.Nonce.Text(16)).Hex("sender", t.auth.From.Bytes()).Msg("make")
 	}
 
-	t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
+	t.auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
 	return addr, tx, err
 }
 
 func (t *TransactAuth) makeTx(ctx context.Context, txFunc func(auth *bind.TransactOpts) (*types.Transaction, error)) (*types.Transaction, error) {
-	auth := t.getAuth(ctx)
+	_, tx, err := t.makeContract(ctx, func(auth *bind.TransactOpts) (ethcommon.Address, *types.Transaction, interface{}, error) {
+		tx, err := txFunc(auth)
+		return ethcommon.BigToAddress(big.NewInt(0)), tx, nil, err
+	})
 
-	tx, err := txFunc(auth)
-
-	if auth.Nonce == nil {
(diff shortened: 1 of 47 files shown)


packages/arb-validator-core/ethbridge/arbClient.go
```diff
@@ -224,7 +260,7 @@ func (c *EthArbAuthClient) NewRollup(address common.Address) (arbbridge.ArbRollu
 	return newRollup(address.ToEthAddress(), c.client, c.auth)
 }
 
-func (c *EthArbAuthClient) NewGlobalInbox(address common.Address, rollupAddress common.Address) (arbbridge.GlobalInbox, error) {
+func (c *EthArbAuthClient) NewGlobalInbox(ctx context.Context, address common.Address, rollupAddress common.Address) (arbbridge.GlobalInbox, error) {
```
hkalodner, line 263: I don't think we need a context here

packages/arb-validator-core/ethbridge/arbRollup.go
```diff
@@ -47,24 +47,26 @@ func newRollup(address ethcommon.Address, client ethutils.EthClient, auth *Trans
...
+		call := &bind.TransactOpts{
+			From:    auth.From,
+			Signer:  auth.Signer,
+			Context: ctx,
+		}
+		blankAddress := ethcommon.Address{}
+		st, err := vm.ArbRollup.GetStakeToken(&bind.CallOpts{Context: ctx})
+		if err != nil {
+			return nil, err
+		}
```
hkalodner, line 60: This looks like it can be outside this block

packages/arb-validator-core/ethbridge/arbRollup.go
```diff
@@ -47,24 +47,26 @@ func newRollup(address ethcommon.Address, client ethutils.EthClient, auth *Trans
...
-		call,
-		common.HashSliceToRaw(proof1),
-		common.HashSliceToRaw(proof2),
-	)
+	tx, err := vm.auth.makeTx(ctx, func(auth *bind.TransactOpts) (*types.Transaction, error) {
+		call := &bind.TransactOpts{
+			From:    auth.From,
+			Signer:  auth.Signer,
+			Context: ctx,
+		}
```
hkalodner, line 55: I think we want to just use `auth` instead of building `call` here

hkalodner (changes requested):

packages/arb-validator-core/ethbridge/arbClient.go
joshuacolvin0: I had already removed this :)

packages/arb-validator-core/ethbridge/arbRollup.go
joshuacolvin0: sounds good

packages/arb-validator-core/ethbridge/arbRollup.go
joshuacolvin0: sure

packages/arb-validator-core/ethbridge/arbClient.go
```diff
@@ -190,30 +194,97 @@ type TransactAuth struct {
...
+			return addr, nil, err
+		}
+
+		txJSON, err := tx.MarshalJSON()
+		if err != nil {
+			log.Err(err).Str("nonce", "nil").Msg("makeContract failed to marshal tx into json")
+			return addr, tx, err
+		}
 
+		log.Info().RawJSON("tx", txJSON).Str("nonce", "nil").Str("sender", t.auth.From.Hex()).Msg("makeContract")
```
hkalodner, line 215: Rather than using `Str` for everything, it'd be better to use format specific handlers, in the case of addresses, we can use `.Hex("sender", t.auth.From.Bytes())

packages/arb-validator-core/ethbridge/arbClient.go
```diff
@@ -190,30 +194,97 @@ type TransactAuth struct {
...
+		return addr, nil, err
+	}
+
+	for i := 0; i < smallNonceRepeatCount && err != nil && strings.Contains(err.Error(), smallNonceError); i++ {
+		// Increment nonce and try again
+		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("makeContract incrementing nonce and submitting tx again")
+
+		auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
```
hkalodner, line 223: On the left side you have `auth.Nonce` and the right `t.auth.Nonce`. I'm guessing these should actually be the same

packages/arb-validator-core/ethbridge/arbClient.go
```diff
@@ -190,30 +194,97 @@ type TransactAuth struct {
...
+
+	// Transaction successful, increment nonce for next time
+	txJSON, err := tx.MarshalJSON()
+	if err != nil {
+		log.Err(err).Str("nonce", auth.Nonce.Text(16)).Msg("makeContract failed to marshal tx into json")
+	} else {
+		log.Info().RawJSON("tx", txJSON).Str("nonce", auth.Nonce.Text(16)).Str("sender", t.auth.From.Hex()).Msg("makeContract")
+	}
+
+	t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
```
hkalodner, line 240: We tend to use the pattern `t.auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))`. I'm don't think that's totally necessary, but it makes it a bit clearer

packages/arb-validator-core/ethbridge/arbClient.go
```diff
@@ -190,30 +194,97 @@ type TransactAuth struct {
 	auth *bind.TransactOpts
 }
 
-func (t *TransactAuth) makeTx(ctx context.Context, txFunc func(auth *bind.TransactOpts) (*types.Transaction, error)) (*types.Transaction, error) {
-	const smallNonceRepeatCount = 5
-	const smallNonceError = "Try increasing the gas price or incrementing the nonce."
+func (t *TransactAuth) makeContract(ctx context.Context, contractFunc func(auth *bind.TransactOpts) (ethcommon.Address, *types.Transaction, interface{}, error)) (ethcommon.Address, *types.Transaction, error) {
```
hkalodner, line 197: Can we unify this with `makeTx`? There's a lot of duplication here

packages/arb-validator/cmd/arb-stressed-validator/arb-stressed-validator.go
```diff
@@ -52,7 +54,7 @@ func main() {
 	}
 }
 
-func createStressedManager(rollupAddress common.Address, client arbbridge.ArbClient, contractFile string, dbPath string) (*rollupmanager.Manager, error) {
+func createStressedManager(ctx context.Context, rollupAddress common.Address, client arbbridge.ArbClient, contractFile string, dbPath string) (*rollupmanager.Manager, error) {
 	return rollupmanager.CreateManager(
 		context.Background(),
```
hkalodner, line 59: This looks like it should be `ctx` instead

hkalodner (changes requested):

packages/arb-validator-core/ethbridge/arbClient.go
joshuacolvin0: nope, this is actually on purpose!  `t.auth.Nonce.Add` updates t.auth.Nonce, but I also want to update `auth.Nonce` separately, so this is a convenient way of doing it.

packages/arb-validator-core/ethbridge/arbClient.go
joshuacolvin0: not necessary to do, but I will go along with convention.

packages/arb-validator/cmd/arb-stressed-validator/arb-stressed-validator.go
joshuacolvin0: good catch

packages/arb-validator-core/ethbridge/arbClient.go
joshuacolvin0: done

packages/arb-validator-core/ethbridge/arbClient.go
joshuacolvin0: Handled all applicable instances

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #652, opened 2020-11-28
Enable zerolog to capture stack traces

joshuacolvin0:
When logging error, include `.Stack().` to include stack trace like this:
```
log.Warn().Stack().Err(err).Msg("failed getting code")
```

Notes:
*  `Msg(...)` or `Msgf(...)` must be supplied or the log line will be ignored
* `Stack()` must be before `Err(...)`
* The log level (`Error()`, `Warn()`, etc) must be supplied before `Stack()`


hkalodner (approved):
LGTM

PR #657, opened 2020-12-05
Wrap rest of Deploy* calls with MakeContract

joshuacolvin0:
A couple of contract deployment calls were missed while refactoring with MakeContract


hkalodner (approved):
LGTM

PR #659, opened 2020-12-06
Add more zerolog and reset counter

joshuacolvin0:
In addition to converting more logging to use zerolog, also reset fetchErrorCount in arbClient when errors stop occurring, that way loop will only exit when 5 errors in a row happen instead of 5 accumulated errors over all time.

### DIFF
--- a/packages/arb-validator/chainlistener/announcerListener.go
+++ b/packages/arb-validator/chainlistener/announcerListener.go
@@ -35,28 +35,28 @@ func NewAnnouncerListener(address common.Address) *AnnouncerListener {
 }
 
 func (al *AnnouncerListener) AddedToChain(context.Context, []*structures.Node) {
-	logger.Info().Msg("AddedToChain")
+	al.logger.Info().Msg("AddedToChain")
 }
 
 func (al *AnnouncerListener) RestartingFromLatestValid(context.Context, *structures.Node) {
-	logger.Info().Msg("RestartingFromLatestValid")
+	al.logger.Info().Msg("RestartingFromLatestValid")
 }
 
 func (al *AnnouncerListener) StakeCreated(ctx context.Context, ng *nodegraph.StakedNodeGraph, ev arbbridge.StakeCreatedEvent) {
-	logger.Info().
+	al.logger.Info().
 		Hex("staker", ev.Staker.Bytes()).
 		Hex("node", ev.NodeHash.Bytes()).
 		Msg("StakeCreated")
 }
 
 func (al *AnnouncerListener) StakeRemoved(ctx context.Context, ev arbbridge.StakeRefundedEvent) {
-	logger.Info().
+	al.logger.Info().
 		Hex("staker", ev.Staker.Bytes()).
 		Msg("StakeRemoved")
 }
 
 func (al *AnnouncerListener) StakeMoved(ctx context.Context, ng *nodegraph.StakedNodeGraph, ev arbbridge.StakeMovedEvent) {
-	logger.Info().
+	al.logger.Info().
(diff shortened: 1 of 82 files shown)


packages/arb-validator/nodegraph/stakedNodeGraph.go
```diff
@@ -114,8 +113,9 @@ func (sng *StakedNodeGraph) Equals(s2 *StakedNodeGraph) bool {
 func (sng *StakedNodeGraph) CreateStake(ev arbbridge.StakeCreatedEvent) {
 	nd, ok := sng.nodeFromHash[ev.NodeHash]
 	if !ok {
-		log.Println("Bad location", ev.NodeHash)
-		panic("Tried to create stake on bad node")
+		logger.Panic().
```
hkalodner, line 116: We should probably use `Fatal` here instead of panic for consistency. I don't actually know the different if any between them

packages/arb-validator/nodegraph/stakedNodeGraph.go
```diff
@@ -128,13 +128,23 @@ func (sng *StakedNodeGraph) CreateStake(ev arbbridge.StakeCreatedEvent) {
 func (sng *StakedNodeGraph) MoveStake(stakerAddr common.Address, nodeHash common.Hash) {
 	staker := sng.stakers.Get(stakerAddr)
 	if staker == nil {
-		log.Fatalf("Moved nonexistant staker %v to node %v", stakerAddr, nodeHash)
+		logger.Fatal().
+			Hex("staker", stakerAddr.Bytes()).
+			Hex("node", nodeHash.Bytes()).
+			Msg("Moved nonexistant staker")
+
+		panic("Will not be reached")
```
hkalodner, line 136: I'm assuming this is needed to make the compiler/linter happy?

packages/arb-validator/nodegraph/stakednodegraph_test.go
```diff
@@ -192,7 +191,7 @@ func TestNodePruneInfoBase(t *testing.T) {
 
 	params := stakedNodeGraph.GenerateNodePruneInfo()
 	if len(params) != 0 {
-		log.Println("params ", params)
+		logger.Error().Msgf("params %v", params)
```
hkalodner, line 194: `t.Log` instead

packages/arb-validator/nodegraph/stakednodegraph_test.go
```diff
@@ -232,7 +231,7 @@ func TestNodePruneInfo(t *testing.T) {
...
```
hkalodner, line 234: `t.Log` instead

packages/arb-validator/nodegraph/stakednodegraph_test.go
```diff
@@ -247,7 +246,7 @@ func TestNodePruneInfo2(t *testing.T) {
...
```
hkalodner, line 249: `t.Log` instead

hkalodner (changes requested):

packages/arb-validator/chainlistener/announcerListener.go
```diff
@@ -38,22 +44,19 @@ func (al *AnnouncerListener) RestartingFromLatestValid(context.Context, *structu
 
 func (al *AnnouncerListener) StakeCreated(ctx context.Context, ng *nodegraph.StakedNodeGraph, ev arbbridge.StakeCreatedEvent) {
 	logger.Info().
```
hkalodner, line 46: I think this and the others should be `al.logger`

hkalodner (changes requested):

hkalodner (approved):
LGTM!

PR #661, opened 2020-12-08
Remove duplicate calls during initialization and add logging

joshuacolvin0:
* Remove duplicate code when initializing aggregator
* Add log messages to provide visibility and timing into restoring state from database


hkalodner (approved):
LGTM

PR #662, opened 2020-12-08
Update ArbOS

hkalodner:
Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/256


edfelten (approved):
Approving. Just an ArbOS update.

PR #663, opened 2020-12-08
Prevent address reuse

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/239/files


edfelten (approved):
Looks good.

PR #664, opened 2020-12-09
Constructor fix

hkalodner:
- Add test to verify create2 behavior when deploying into address with existing balance
- Update ArbOS to https://github.com/OffchainLabs/arb-os/pull/243/files


edfelten (approved):
Looks good.

PR #666, opened 2020-12-09
Arbsys refactor

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/241
- Update aggregator syscall support and test


edfelten (approved):
Looks good.

PR #667, opened 2020-12-09
Block log update

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/246
- Add support for updated block log format


packages/arb-evm/evm/blockresult.go
```diff
@@ -0,0 +1,181 @@
...
+	ChainStats    *OutputStatistics
+	PreviousBlock *big.Int
+}
+
+func (b *BlockInfo) LastAVMLog() *big.Int {
+	return new(big.Int).Sub(b.ChainStats.AVMLogCount, big.NewInt(1))
+}
+
+func (b *BlockInfo) FirstAVMLog() *big.Int {
+	return new(big.Int).Sub(b.LastAVMLog(), b.BlockStats.AVMLogCount)
```
edfelten, line 59: It looks like you need to add 1 to this.  If the logs in this block are numbers 10, 11, and 12, then LastAVMLog() will be 12, AVMLogCount will be 3, so FirstAVMLog will return 9.  It should return 10.

packages/arb-evm/evm/blockresult.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (b *BlockInfo) LastAVMSend() *big.Int {
+	return new(big.Int).Sub(b.ChainStats.AVMSendCount, big.NewInt(1))
+}
+
+func (b *BlockInfo) FirstAVMSend() *big.Int {
+	return new(big.Int).Sub(b.LastAVMSend(), b.BlockStats.AVMSendCount)
```
edfelten, line 67: Same off-by-one bug as above -- should add 1 to this.

edfelten (changes requested):
Found a couple of apparent bugs.

hkalodner:
I'm splitting this into two PRs so I'm closing this one

Backwards compatible changes are in https://github.com/OffchainLabs/arbitrum/pull/668
ArbOS breaking changes will be in a new PR

PR #668, opened 2020-12-09
Fix avm send index in block log

hkalodner:
- Split up results.go in evm package
- Expand out arbos block log test
- Fix off by one error in `FirstAVMSend`


edfelten (approved):
Looks good.

PR #669, opened 2020-12-09
Add prev block

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/246
- Add support for updated block log format


edfelten (approved):
Looks good.

PR #671, opened 2020-12-11
Cleanup arbos logs test

hkalodner:
- Fix handling of logs with no topic (previously these incorrectly returned an error)
- Add better debug info for failures in the arbos log test


joshuacolvin0 (approved):
LGTM

PR #672, opened 2020-12-11
Update ArbOS

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/254

Note: I ran the logs generated from the EVM test suite through the go test harness, and everything passes cleanly


edfelten (approved):
Looks good.

PR #673, opened 2020-12-11
add @ethersproject/address, bytes, bignumber dependencies to arb-provider-ethers

DZGoldman:
Need to be direct dependencies for byte-serialization stuff (so far the projects using arb-provider-ethers just happen to have them anyway, as does some sibling in the mono-repo)


hkalodner (approved):
LGTM

PR #676, opened 2020-12-12
Validator cleanup

hkalodner:
- Optimize inbox serialization (don't hash every time)
- Improve validator logging
- Add `--pprof` flag to validator to enable pprof server over port 8081


joshuacolvin0 (approved):
LGTM

PR #682, opened 2020-12-14
Do not send block/log announcements to PUBSUB until processed

joshuacolvin0:
The previous behavior of block and log PUBSUB was to announce as soon as received.  This caused race conditions when clients saw announcement but blocks were not available via RPC yet.  Fix is to delay PUBSUB announcement until block/log data has been added to database.


hkalodner (approved):
LGTM

PR #683, opened 2020-12-14
Add additional logging to fibgo unit test

joshuacolvin0:
`t.Log` is output at different rate than `zerolog`, so add some additional `zerolog` based logging so that we can have a conclusive reason for intermittent pipeline failures.


hkalodner (approved):
LGTM

PR #686, opened 2020-12-15
DB init logging and pprof

joshuacolvin0:
Additional log messages added to better track where time is spent while initializing database.
Also add optional pprof http endpoint


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #688, opened 2020-12-16
Add arbitrum local dev chain

hkalodner:
- Add `arb-dev-aggregator` which launches a standalone local dev chain

Support optional `--save file.json` to save a log file containing the full inbox contents every quitting
Support `--pprof` flag to optionally enable a pprof server over port 8081


joshuacolvin0 (approved):
LGTM

PR #689, opened 2020-12-16
Dev agg improvements

hkalodner:
- Have dev aggregator seed accounts from mnemonic
- Make RPC calls that end in result codes other than EVM return or revert return an error instead of an empty result


joshuacolvin0 (approved):
LGTM

PR #690, opened 2020-12-16
Assorted Fixes

hkalodner:
- Increase max gas limit for calls to aggregator
- Fix potential race conditions in dev aggregator


joshuacolvin0 (approved):
nice

PR #693, opened 2020-12-20
Origin fix

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/268
- Add test for `ORIGIN` opcode


edfelten (approved):
Looks good.

PR #699, opened 2020-12-23
Pretty debug messages

hkalodner:
- Upgrade ArbOS with https://github.com/OffchainLabs/arb-os/pull/271
- Add pretty logging for debug prints


edfelten (approved):
Looks good.

PR #700, opened 2020-12-24
Update ArbOS

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/276

Note: Currently there's no testing of the `EXTCODEHASH` opcode. We should definitely have that to prevent regressions like this in the future


edfelten (approved):
Looks good.

PR #707, opened 2020-12-28
Upgrade arbos

hkalodner:
- Upgrade ArbOS to https://github.com/OffchainLabs/arb-os/pull/288
- Add failed deposit test


edfelten (approved):
Looks good.

PR #711, opened 2020-12-30
Assorted Fixes

hkalodner:
- Add `personal_sign` support
- Adust dev aggregator handling of reorgs and invalid transactions
- Adjust txdb handling of transaction receipts to favor receipt types that increment the nonce


joshuacolvin0 (approved):
LGTM

PR #723, opened 2021-01-11
Hk fixes

hkalodner:
- Add owner controls to rollup contract
- Enable removing of old stakes


packages/arb-bridge-eth/contracts/rollup/ArbRollup.sol
```diff
@@ -283,6 +287,46 @@ contract ArbRollup is IArbRollup, Cloneable, NodeGraph, Staking {
         safeSelfDestruct(msg.sender);
     }
 
+    function ownerToggleStakerAllowListed(bool shouldRequire) external onlyOwner {
```
edfelten, line 290: I think this should be called `ownerToggleStakerAllowList`.

joshuacolvin0 (approved):
LGTM

edfelten (dismissed):
Looks good.  One small naming suggestion, but I'm not going to hold this up over that, given that this code is probably obsolete anyway.

packages/arb-bridge-eth/contracts/rollup/ArbRollup.sol
hkalodner: I'd change this, but I already deployed it

PR #728, opened 2021-01-17
New Rollup Protocol Implementation

hkalodner:


packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+            "INBOX_PAST_END"
+        );
+
+        uint256 timeSinceLastNode = block.number - assertion.beforeProposedBlock;
+        // Verify that assertion meets the minimum Delta time requirement
+        require(timeSinceLastNode >= minimumAssertionPeriod(), "TIME_DELTA");
+
+        // Minimum size requirements: each assertion must satisfy either
+        require(
+            // Consumes at least all inbox messages put into L1 inbox before your prev node’s L1 blocknum
```
edfelten, line 371: Should have a third possibility here, which is that the assertion has the maximum allowed number of sends.  (Also, should enforce a maximum allowed number of sends.)

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    /**
+     * @notice Reduce the amount staked for the sender
+     * @param target Target amount of stake for the staker. If this is below the current minimum, it will be set to minimum instead
+     */
+    function reduceDeposit(uint256 target) external whenNotPaused {
+        checkUnchallengedStaker(msg.sender);
+        uint256 currentRequired = currentRequiredStake();
+        if (target > currentRequired) {
```
edfelten, line 443: I think this should be `target < currentRequired`.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+            target = currentRequired;
+        }
+        reduceStakeTo(msg.sender, target);
+    }
+
+    /**
+     * @notice Reduce the amount staked for the sender
```
edfelten, line 450: Wrong description for this function.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+        if (loserStake > winnerStake) {
+            loserStake -= reduceStakeTo(losingStaker, winnerStake);
+        }
+
+        uint256 amountWon = loserStake / 2;
+        increaseStakeBy(winningStaker, amountWon);
+        loserStake -= amountWon;
+        clearChallenge(winningStaker);
+
+        // TODO: deposit extra loserStake into ArbOS
```
edfelten, line 528: You can give "burned" funds to ArbOS by doing a payable call to ArbOwner.addToReserveFunds(), with the chain owner as the sender.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+     * @param zombieNum Index of the zombie to remove
+     * @param maxNodes Maximum number of nodes to remove the zombie from (to limit the cost of this transaction)
+     */
+    function removeZombie(uint256 zombieNum, uint256 maxNodes) external whenNotPaused {
+        require(zombieNum <= zombieCount(), "NO_SUCH_ZOMBIE");
+        address zombieStakerAddress = zombieAddress(zombieNum);
+        uint256 latestStakedNode = zombieLatestStakedNode(zombieNum);
+        uint256 nodesRemoved = 0;
+        uint256 firstUnresolved = firstUnresolvedNode();
+        while (latestStakedNode > firstUnresolved && nodesRemoved < maxNodes) {
```
edfelten, line 543: I think this should be `latestStakedNode >= firstUnresolved && ...`

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * @notice Calculate the current amount of funds required to place a new stake in the rollup
+     * @return The current minimum stake requirement
+     */
+    function currentRequiredStake() public view returns (uint256) {
```
edfelten, line 578: This doesn't quite do the right thing. We want the stake to escalate if the chain could make progress but isn't.  So we need to not escalate when there are no unresolved nodes beyond latestConfirmed.  And we should start escalating after some unconfirmed node has had its deadline passed.

So I think the right approach is to say that if there are no unresolved nodes after latestConfirmed, then don't escalate.  If there are unresolved nodes, then escalate based on the time since the deadline of firstUnresolvedNode.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+                stakedZombieCount++;
+            }
+        }
+        return stakedZombieCount;
+    }
+
+    /**
+     * @notice Verify that no stake has been placed within the last challenge period
+     */
+    function checkNoRecentStake() public view {
```
edfelten, line 628: Rather than naming functions like this `checkFoo`, it seems better to name them `requireFoo`.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+            require(
+                IERC20(stakeToken).transferFrom(msg.sender, address(this), tokenAmount),
+                "TRANSFER_FAILED"
+            );
+            return tokenAmount;
+        }
+    }
+
+    /**
+     * @notice Send funds to funds to the given address, if staking is eth, transfer eth, otherwise transfer tokens
```
edfelten, line 663: Repeated words in notice line.

edfelten (changes requested):
First batch of comments and change requests.

edfelten (changes requested):
A few more comments.

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -0,0 +1,126 @@
...
+    function destroy() external override onlyRollup {
+        selfdestruct(msg.sender);
+    }
+
+    /**
+     * @notice Mark the given staker as staked on this node
+     * @param staker Address of the staker to mark
+     */
+    function addStaker(address staker) external override onlyRollup {
+        require(!stakers[staker], "ALREADY_STAKED");
```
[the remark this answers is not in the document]
hkalodner: The node contract doesn't have the information. This is only callable by the rollup contract. We could better document this maybe. But I think from a security and correctness perspective, we only really need to deal with functions that are publicly callable

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    /**
+     * @notice Truncate the pending nodes done to the given index
+     * @param _latestNodeCreated New latest node after the truncation
+     */
+    function truncateNodes(uint256 _latestNodeCreated) external onlyOwner {
+        uint256 oldLatestNodeCreated = latestNodeCreated();
+        require(_latestNodeCreated < oldLatestNodeCreated, "TOO_NEW");
+        require(_latestNodeCreated >= firstUnresolvedNode() - 1, "TOO_OLD");
```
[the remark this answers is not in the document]
hkalodner: If we do this then we also need to account for updating the first unresolved node. It didn't seem worth it to do that, but I'm definitely open to it if there's a good motivation

packages/arb-bridge-eth/contracts/bridge/Outbox.sol
```diff
@@ -0,0 +1,145 @@
...
+        // If we've reached here, we've already confirmed that sum(sendLengths) == sendsData.length
+        uint256 messageCount = sendLengths.length;
+        uint256 offset = 0;
+        for (uint256 i = 0; i < messageCount; i++) {
+            handleOutgoingMessage(bytes(sendsData[offset:sendLengths[i]]));
+            offset += sendLengths[i];
+        }
+    }
+
+    function handleOutgoingMessage(bytes memory data) private {
```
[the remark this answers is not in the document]
hkalodner: Can you point me to a spec or code for how this should be handled?

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+
+        for (uint256 i = oldLatestNodeCreated; i >= _latestNodeCreated; i--) {
+            INode node = getNode(i);
+            node.destroy();
+        }
+        updateLatestNodeCreated(_latestNodeCreated);
```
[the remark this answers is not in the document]
hkalodner: Ah whoops. Yep this is a bug. This makes truncation way more complex since it needs to be O(N) in the number of stakers

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+        external
+        whenNotPaused
+    {
+        checkUnresolved();
+        uint256 latest = latestConfirmed();
+        uint256 firstUnresolved = firstUnresolvedNode();
+        INode node = getNode(firstUnresolved);
+        if (node.prev() == latest) {
+            checkNoRecentStake();
+            require(successorWithStake > firstUnresolved, "SUCCESSOR_TO_LOW");
```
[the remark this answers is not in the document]
hkalodner: Fixed

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -0,0 +1,699 @@
...
+    function minimumAssertionPeriod() public view returns (uint256) {
+        return challengePeriodBlocks / 10;
+    }
+
+    /**
+     * @notice Calculate the number of zombies staked on the given node
+     * @param node The node on which to count staked zombies
+     * @return The number of zombies staked on the node
+     */
+    function countStakedZombies(INode node) public view returns (uint256) {
```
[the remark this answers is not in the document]
hkalodner: Added

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Added with  a todo to figure out what the maximum should be

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Fixed

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Fixed

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Great catch, fixed

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: I think that makes sense and it's a pretty simple change. This has been updated

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Good call. I've updated a bunch of those

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Fixed

packages/arb-bridge-eth/contracts/challenge/Challenge.sol
```diff
@@ -16,67 +16,555 @@
...
+        uint256 challengedSegmentLength,
+        bytes32[] inboxAccHashes,
+        bytes32[] inboxDeltaHashes
+    );
     event AsserterTimedOut();
     event ChallengerTimedOut();
+    event OneStepProofCompleted();
+    event ConstraintWin();
 
     // Can online initialize once
```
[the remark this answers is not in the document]
hkalodner: Fixed

packages/arb-bridge-eth/contracts/challenge/Challenge.sol
```diff
@@ -16,67 +16,555 @@
...
+        setExecutionDeadline(_gasClaimed);
+        turn = Turn.Challenger;
+
+        challengeState = 0;
+
+        emit InitiatedChallenge();
+    }
+
+    /**
+     * @notice Mark the given staker as staked on this node
```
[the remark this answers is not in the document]
hkalodner: Fixed up comments in the challenge file

packages/arb-bridge-eth/contracts/challenge/Challenge.sol
```diff
@@ -16,67 +16,555 @@
...
+
+        // Reuse the executionHash variable to store last assertion
+        updateBisectionRoot(_chainHashes, 0, _newSegmentLength);
+        executionHash = _chainHashes[_chainHashes.length - 1];
+
+        kind = Kind.StoppedShort;
+        // Free no longer needed storage
+        inboxConsistencyHash = 0;
+        inboxDeltaHash = 0;
+        executionHash = 0;
```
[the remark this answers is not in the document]
hkalodner: Good catch, that was wrong. Fixed now

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: This is gonna be way more complicated because we support staking in tokens, but the main bridge only support ETH so this will take more work

packages/arb-bridge-eth/contracts/rollup/Node.sol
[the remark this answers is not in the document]
edfelten: OK, I didn't notice that the calling paths to this all do the checks.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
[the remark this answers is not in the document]
edfelten: This one could go either way.   In the interest of simplicity, let's leave it as-is.

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
edfelten: The best way is probably just to deposit the burned tokens to some configurable address at L2.  Then we can create an L2 contract later that disposes of them in some useful way.

packages/arb-bridge-eth/contracts/bridge/Outbox.sol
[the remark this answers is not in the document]
edfelten: Best documentation is in the ArbOS PR.  Basically what is sent is a byte-string of 96 bytes, containing batchNumber (32 bytes), numItemsInBatch (32 bytes), Merkle root (32 bytes).

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -185,23 +191,70 @@ contract Rollup is RollupCore, Pausable, IRollup {
...
     function resume() external onlyOwner {
+        require(!truncating, "STILL_TRUNCATING");
         _unpause();
     }
 
     /**
-     * @notice Truncate the pending nodes done to the given index
-     * @param _latestNodeCreated New latest node after the truncation
+     * @notice Begin the process of trunacting the chain back to the given node
+     * @dev maxItems is used to make sure this doesn't excide the max gas cost
```
edfelten, line 200: Typo: should be "exceed"

edfelten (dismissed):
Looks good.

PR #731, opened 2021-01-19
Implement new challenge period protocol

hkalodner:
Separated into a separate PR to make review easier


edfelten (changes requested):
Found one bug.  Also have an issue for discussion that I raised on Slack.

edfelten (approved):
Looks good.

PR #734, opened 2021-01-21
Rollup protocol changes

hkalodner:


packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -347,11 +347,19 @@ contract RollupCore {
...
+        address stakerAddress,
+        uint256 nodeNum,
+        uint256 confirmPeriodBlocks
+    ) internal returns (uint256) {
         Staker storage staker = _stakerMap[stakerAddress];
         INode node = _nodes[nodeNum];
-        node.addStaker(stakerAddress);
+        uint256 newStakerCount = node.addStaker(stakerAddress);
         staker.latestStakedNode = nodeNum;
+        if (newStakerCount == 0) {
```
edfelten, line 359: I think this should be `newStakerCount == 1`.  (node.addStaker increments the count before returning it)

edfelten (changes requested):
One apparent bug.  Also, I don't think I saw the code that sets newChildConfirmDeadline for a new node that isn't the first child of its parent.

edfelten (approved):
Looks good.

PR #736, opened 2021-01-22
Ethbridge Fixes

hkalodner:
- This PR fixes a number of different small issues in the implementation of the ethbridge


edfelten (approved):
These changes all look good.  I haven't verified whether they address all of the reported issues.

PR #741, opened 2021-01-29
Update GlobalInbox interface in docs

shanefontaine:
Looks like the [GlobalInbox contracts](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-eth/contracts/inbox/GlobalInbox.sol) got updated but the [docs](https://developer.offchainlabs.com/docs/ethereum_interoperability) were not updated to reflect that change.

This fixes the interface for GlobalInbox in the docs.


DZGoldman:
now deprecated (thanks tho!)

PR #742, opened 2021-02-01
Fix output root uniqueness

hkalodner:
- Fix issue allowing outputs to be redeemed multiple times


edfelten (approved):
Looks good to me.  I think this addresses the issue identified previously.

PR #744, opened 2021-02-03
Add ERC777 tokens to bridge

dmihal:
Changes

* The default L2 token is an ERC777 token
* ERC20 tokens can still be created on L2 by calling `depositAsERC20`
* L2 tokens can be migrated between ERC20, ERC777, and custom tokens (custom tokens can not be migrated to ERC20/ERC777 due to being untrusted code)
* create2 is used for creating tokens on L2. This removes the need for storing token mappings in the L2 bridge
* Every deposit includes the number of decimals. This avoids any decimal/granularity edge cases, with minimal additional costs (single SLOAD on deposit)


packages/arb-bridge-eth/contracts/libraries/CloneFactory.sol
```diff
@@ -62,6 +62,19 @@ contract CloneFactory {
         }
     }
 
+    function calculateCreate2CloneAddress(ICloneable target, bytes32 salt) internal view returns (address calculatedAddress) {
+        calculatedAddress = address(uint(keccak256(abi.encodePacked(
+            byte(0xff),
+            address(this),
+            salt,
+            abi.encodePacked(
```
fredlacs, line 70: The contract bytecode needs to be hashed before encoded with the other fields according to the EIP 

https://eips.ethereum.org/EIPS/eip-1014

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -0,0 +1,157 @@
...
+        require(tx.origin == address(this), "ONLY_ETH_PAIR");
+        _;
+    }
+
+    constructor() public {
+        templateERC20 = new StandardArbERC20();
+        templateERC777 = new StandardArbERC777();
+    }
+
+    function mintFromL1(
```
fredlacs, line 52: I think we should be explicit on the calls that use the ERC777 standard, the same way as ERC20. Adding the standard to each function name helps reduce possible error, so users are more explicit about what they are using.

My main concern is that someone may interact with this function without knowing this is using a standard with callback hooks under the hood.

fredlacs (changes requested):
Hey David, awesome PR. 

I quite like your idea of using CREATE2 to deploy the proxies. I found a small mistake on how you calculate the addresses, but easy fix.

I also think it's a good idea to keep the bridge methods more explicit about what is being used under the hood. I'm afraid some devs might deploy logic interacting with ie `mintFromL1` or `updateTokenInfo` and be unaware the underlying token has callback hooks. If we keep the erc20 and erc777 labels explicit, we help minimise error in that front.

dmihal:
Thanks for the feedback @fredlacs! I've updated the code, adding "ERC777" to all relevant function names.

I also replaced CloneFactory with a modified version of Clones.sol from OpenZeppelin, which includes deterministic create2 deployments

fredlacs:
Sweet, I opened a PR at https://github.com/dmihal/arbitrum/pull/1 that uses the openzeppelin code as a library.
Let me know what you think!

DZGoldman:
Can we rebase `new-rollup` into this branch instead of merging the new OZ commit? That way we'll avoid the monster diff (there'll be conflicts to resolve locally in clonefactory when rebasing)

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -113,50 +114,66 @@ contract EthERC20Bridge is L1Buddy {
...
+
+        sendPairedContractTransaction(
+            maxGas,
+            gasPriceBid,
+            abi.encodeWithSelector(_selector, erc20, name, symbol, decimals)
+        );
+    }
+
+    function depositAsERC777(
+        address erc20,
```
DZGoldman, line 136: maybe rename parameter to erc777 ?

dmihal: I don't think so, because the user is depositing an ERC20, not an ERC777

dmihal:
@DZGoldman @fredlacs I've rebased `new-rollup`, so no need for that merge PR :)

PR #747, opened 2021-02-05
Add fuzzer for AVM execution and proofs

PlasmaPower:
Still hacky, but now with scripts to do the hacky stuff for you. Also seems to have a memory leak somewhere. Run:
```
packages/arb-avm-cpp/configure-for-fuzzing.sh
packages/arb-node-core/challenge/fuzz/build.sh
packages/arb-node-core/challenge/fuzz/fuzz.sh
```


packages/arb-avm-cpp/avm_values/src/vmValueParser.cpp
```diff
@@ -140,6 +140,9 @@ Operation operation_from_json(const nlohmann::json& op_json,
         std::cerr << "Invalid opcode " << opcode_json << "\n";
     }
     auto opcode = opcode_json.get<OpCode>();
+    if (!instructionValidity()[(size_t)opcode]) {
```
hkalodner, line 143: Opcodes that aren't defined should still be valid inputs, but will cause an error when executed. I think we probably want to allow the parser to read in invalid opcodes like this and probably print a warning to `cerr` instead of failing

packages/arb-avm-cpp/cavm/cmachine.cpp
```diff
@@ -133,7 +133,12 @@ ByteSlice machineMarshallForProof(CMachine* m) {
     assert(m);
     auto mach = static_cast<Machine*>(m);
     std::vector<unsigned char> buffer;
-    return returnCharVector(mach->marshalForProof());
+    try {
+        return returnCharVector(mach->marshalForProof());
+    } catch (const std::exception& e) {
+        std::cerr << "Error marshalling machine for proof: " << e.what() << std::endl;
+        return returnCharVector(buffer);
+    }
```
hkalodner, line 141: This is good, but we further should make sure to properly detect this case in the `go` code that calls this method. `MarshalForProof` in `cmachine/machine.go` is already setup to be able to return an error. A check should be added there to return an error in the case that the proof is 0-length which should never happen in a non-error case

packages/arb-node-core/challenge/generateProofCases.go
```diff
@@ -0,0 +1,148 @@
...
+	Proof       hexutil.Bytes
+	BufferProof hexutil.Bytes
+}
+
+func GenerateProofCases(contract string, fuzzTarget bool) ([]*ProofData, error) {
+	mach, err := cmachine.New(contract)
+	if err != nil {
+		if fuzzTarget {
+			return make([]*ProofData, 0), nil
+		}
```
hkalodner, line 55: I don't think this should be necessary since in most cases in go, `nil` is equivalent to to an empty slice

packages/arb-node-core/challenge/generateProofCases.go
```diff
@@ -0,0 +1,148 @@
...
+		LogAcc:       ethcommon.Hash{},
+		LogCount:     (*hexutil.Big)(big.NewInt(0)),
+	}
+	nextMessageIndex := big.NewInt(0)
+	proofs := make([]*ProofData, 0)
+	for i := uint64(0); i < maxSteps; i++ {
+		proof, err := mach.MarshalForProof()
+		if fuzzTarget && len(proof) == 0 {
+			return proofs, nil
+		}
```
hkalodner, line 83: I think the change to `MarshalForProof()` to return an error in the case of a zero length proof should eliminate the need for this special case

packages/arb-node-core/challenge/generateProofCases.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type ProofData struct {
+	BeforeCut   ExecutionCutJSON
+	AfterCut    ExecutionCutJSON
+	Proof       hexutil.Bytes
+	BufferProof hexutil.Bytes
+}
+
+func GenerateProofCases(contract string, fuzzTarget bool) ([]*ProofData, error) {
```
hkalodner, line 50: With the other two uses of `fuzzTarget` eliminated, we can just pass `maxSteps` in here as an arg

hkalodner (changes requested):
This is awesome! Left some small comments. One thing not otherwise noted, is that I think it'd be good to put the `fuzz` folder inside the `cmd/machine-fuzz` rather than inside challenge to be a bit more idiomatic

packages/arb-node-core/challenge/generateProofCases.go
PlasmaPower: I wanted to differentiate between types of errors, and panic if GenerateProofCases returned a meaningful error. I'll think about if I can directly panic on meaningful errors in this function though.

PlasmaPower:
I've resolved your comments. This should be ready for review again.

PR #748, opened 2021-02-05
In ec_recover AVM impl, pop values before doing anything with them

PlasmaPower:
This matches the OneStepProof impl. Found with #747.


hkalodner (approved):
LGTM

PR #749, opened 2021-02-05
Update one step proof generation and verification

hkalodner:
- Treat invalid opcodes as equivalent to the error opcode
- Properly handle stack underflow in proofs


packages/arb-bridge-eth/contracts/arch/OneStepProofCommon.sol
```diff
@@ -312,42 +310,53 @@ abstract contract OneStepProofCommon {
...
-        impl(context);
+        if (context.errorOccurred) {
+            if (context.afterMachine.errHandlerHash == CODE_POINT_ERROR) {
+                context.afterMachine.setErrorStop();
+            } else {
+                // Clear error
+                context.errorOccurred = false;
+                context.afterMachine.instructionStackHash = context.afterMachine.errHandlerHash;
+
+                if (!(context.hadImmediate && dataPopCount == 0)) {
```
edfelten, line 354: Don't we need to do something with the stack when the condition in this `if` is false?

edfelten (changes requested):
One suggestion for the spec, and a question about correctness in the one-step proof.

packages/arb-bridge-eth/contracts/arch/OneStepProofCommon.sol
hkalodner: What this should do is clear the stack unless the instruction had an immediate and didn't have any arguments from the spec, in which case by the current behavior, the immediate value should stay on the stack. The semantics of this are the same as though the immediate is pushed to the stack before the instruction is run

docs/AVM_Specification.md
```diff
@@ -166,7 +166,13 @@ A VM that is in the Halted or ErrorStop state cannot execute instructions.
...
 A VM that is in an extensive state can execute an instruction. To execute an instruction, the VM gets the opcode from the Current Codepoint.
 
-If the opcode is not a value that is an opcode of the AVM instruction set, the machine is executed as though the instruction were the error opcode. Otherwise the VM carries out the semantics of the opcode.
+If the opcode is not a value that is an opcode of the AVM instruction set, the machine is executed as though the instruction were the error opcode.
+
+If the opcode is valid, but there are not enough items on the stack or auxstack to execute the opcode, the machine is executed as though the instruction were an error opcode.
+
+Otherwise the VM carries out the semantics of the opcode.
+
+If an error is raised while executing the opcode, remove the items from the stack and auxstack that would have been consumed by the opcode. In the case of underflow, remove the minimum of the number of items consumed and the size of the stack.
```
edfelten, line 175: This paragraph is inconsistent with the statement two paragraphs earlier which says that if there are not enough items on the stack or auxstack (i.e. the instruction will underflow) then execute as though it's an error opcode.

edfelten (changes requested):
One inconsistency in the spec to resolve.

edfelten (approved):
Looks good.

PR #750, opened 2021-02-06
Update AVM spec to clarify Error semantics

edfelten:
In the AVM spec, clarify what the effect of an Error should be on the Stack and the AuxStack.


hkalodner (approved):
LGTM

PR #751, opened 2021-02-08
Fix edge case in code segment creation

PlasmaPower:
I'm not entirely confident in this fix, but it seems to resolve the immediate problem.


hkalodner (approved):
Looks great

PR #752, opened 2021-02-08
Finish implementing sideload instruction

PlasmaPower:


hkalodner (approved):
We'll have to update the c++ AVM with the custom compatible implementation of sideload later, but this is definitely the place to start

PR #753, opened 2021-02-09
Fix buffer proof checking in Solidity

PlasmaPower:
I'm not sure if this is all the issues with buffers, but it fixes two that I found, namely an issue hashing buffers filled with zero bytes over 65 bytes long, and an off by one.


mrsmkl:
Looks like the hashing was quite broken, hopefully it's fixed here https://github.com/OffchainLabs/arbitrum/pull/756

PlasmaPower:
I've merged in #756 which contains more hashing fixes

packages/arb-bridge-eth/contracts/arch/Hashing.sol
```diff
@@ -42,35 +42,49 @@ library Hashing {
         return res;
     }
 
+    /*
+     * Note that dataLength has to be a power of two.
```
hkalodner, line 46: Can we add a require/revert for this? Seems like otherwise this is pretty risky. Alternatively, this should be private to make it clear that the API is unsafe. One third option would be to pass in `n` instead of `2^n`

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #754, opened 2021-02-09
Specify the type code of buffers in the AVM

PlasmaPower:


packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -413,6 +413,8 @@ void typeOp(MachineState& m) {
         m.stack[0] = CODEPT;
     else if (nonstd::holds_alternative<Tuple>(m.stack[0]))
         m.stack[0] = TUPLE;
+    else if (nonstd::holds_alternative<Buffer>(m.stack[0]))
```
hkalodner, line 416: Could you actually replace this whole impl with `nonstd::visit` and a functor. Type safety would've caught this immediately

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #756, opened 2021-02-10
fixing hashing

mrsmkl:


packages/arb-bridge-eth/contracts/arch/Hashing.sol
```diff
@@ -47,30 +47,32 @@ library Hashing {
...
+                return (keccak1(bytes32(0)), true);
             }
-            return keccak1(bytes32(bytes32FromArray(data, startOffset)));
+            bytes32 res = keccak1(bytes32(bytes32FromArray(data, startOffset)));
+            return (res, res == keccak1(bytes32(0)));
         }
-        bytes32 h2 = merkleRoot(data, startOffset + dataLength / 2, dataLength / 2, pack);
-        if (h2 == keccak1(bytes32(0)) && pack) {
-            return merkleRoot(data, startOffset, dataLength / 2, true);
+        (bytes32 h2, bool zero2) = merkleRoot(data, startOffset + dataLength / 2, dataLength / 2, false);
```
PlasmaPower, line 58: I think this last parameter here needs to be `pack`, not `false`. I believe as-is this wouldn't work with a 128 byte buffer.

PlasmaPower (changes requested):
Could you also add a test requiring that the last byte of a 65 byte hash is hashed in? I believe as-is the last byte would be cut off, due to the length halving.

packages/arb-bridge-eth/contracts/arch/Hashing.sol
mrsmkl: `false` should be correct here, the right hand subtree should never be packed, otherwise the tree would be kind of unbalanced.

PlasmaPower: I'm not entirely sure what packing means, but not enabling it for the right hand of the subtree means that you end up with keccak1(keccak1(0)) instead of keccak1(0) for the right hand of the subtree if it's all zeroes and big enough. That's why I suggested the 128 byte test.

PlasmaPower: We start with the root of indicies 0..128. h2 of that is 64..128 with packing disabled. Since packing is disabled, it'll return keccak2(keccak1(0), keccak1(0)) as those are its two halves. That propagates upwards to the root hash, making it not equal to keccak1(0) as expected.

PlasmaPower: Ah, I see, at the top level it ignores the lower level hash because of the new zero return value. I think this is functionally identical to passing in pack here but it works too. :+1:

PlasmaPower (approved):
LGTM

PR #757, opened 2021-02-10
Add setbuffer bounds checking to match C++ AVM impl

PlasmaPower:


docs/AVM_Specification.md
```diff
@@ -265,9 +265,9 @@ The instructions are as follows:
...
-| 0xa4 | setbuffer8 | Pop three values (A), (B) and (C) off the stack. The value A must be a buffer, B must be an integer smaller than 2**64 and C must be an integer. Pushes to stack a new buffer that is same as A except that byte in position B is now C. | 100 |
-| 0xa5 | setbuffer64 | Pop three values (A), (B) and (C) off the stack. The value A must be a buffer, B must be an integer smaller than 2**64-7 and C must be an integer. Pushes to stack a new buffer that is same as A except that bytes in positions B..B+7 are now BE representation of C as 64-bit integer. | 100 |
-| 0xa6 | setbuffer256 | Pop three values (A), (B) and (C) off the stack. The value A must be a buffer, B must be an integer smaller than 2**64-31 and C must be an integer. Pushes to stack a new buffer that is same as A except that bytes in positions B..B+31 are now BE representation of C as 256-bit integer. | 100 |
+| 0xa4 | setbuffer8 | Pop three values (A), (B) and (C) off the stack. The value A must be a buffer, B must be an integer smaller than 2**64 and C must be an integer smaller than 2**8. Pushes to stack a new buffer that is same as A except that byte in position B is now C. | 100 |
```
hkalodner, line 268: I think that this should explicitly state that otherwise this should error. Same for the other opcodes of course

packages/arb-bridge-eth/contracts/arch/OneStepProof2.sol
```diff
@@ -477,7 +477,10 @@ contract OneStepProof2 is IOneStepProof2, OneStepProofCommon {
             handleOpcodeError(context);
             return;
         }
-        require(val2.intVal < (1 << 64), "buffer index must be 64-bit");
+        if (val2.intVal >= 1 << 64) {
```
hkalodner, line 480: I would've expected this to be `1 << 64 - 7` so that the read doesn't overflow. Are we defining overflow here as 0s? Might be nicer to just disallow reading past the max size like we disallow writing past the max size

hkalodner (changes requested):

packages/arb-bridge-eth/contracts/arch/OneStepProof2.sol
PlasmaPower: The 8 in getBuffer8 refers to the number of bits, not the number of bytes, so that's not necessary. If you scroll down a bit to getBuffer64 you'll see we do have a `- 7` there.

hkalodner (approved):
LGTM

PR #760, opened 2021-02-11
Fix setbuffer8 bounds checking

PlasmaPower:
In #757 I missed that the C++ setbuffer8 impl didn't have bounds checking on the value. Bounds checking on the offset is already done with the assumeInt64 (since this is a single byte no extra bytes are needed).


hkalodner (approved):
LGTM

PR #763, opened 2021-02-14
Update openzeppelin dependency and use it for new clone factory logic

fredlacs:
Open-zeppelin added support for ERC1167 with its v3.4.0 release.
This PR updates open-zeppelin to the latest version and uses its library logic for the minimal proxy from ERC1167.


hkalodner (approved):
LGTM

PR #764, opened 2021-02-16
Implement sideload instrumentation in machine

PlasmaPower:
Also refactors Machine::run to use an options 

- [x] Add sideload instrumentation in machineoperation
- [x] Add new APIs to handle sideloads at the Machine level
- [x] Store sideload block numbers in ArbCore
- [x] Add a sideload cursor, or a way to jump to a sideload with an execution cursor?
- [x] Update Go bindings to Machine::run
- [x] Add tests for loading sideload machine


PlasmaPower:
I just realized I haven't updated the Go bindings yet. The C++ side works though.

packages/arb-avm-cpp/avm/src/machinestate/machinestate.cpp
```diff
@@ -29,13 +30,24 @@ namespace {
...
 }  // namespace
 
-AssertionContext::AssertionContext(
-    std::vector<Tuple> inbox_messages,
-    const nonstd::optional<uint256_t>& min_next_block_height,
-    uint256_t messages_to_skip)
-    : inbox_messages(std::move(inbox_messages)),
-      next_block_height(min_next_block_height),
-      inbox_messages_consumed(messages_to_skip) {}
+AssertionContext::AssertionContext(const MachineExecutionConfig& config)
```
hkalodner, line 33: I think we'll want to take `MachineExecutionConfig` by value here so we can copy out the inbox messages unless there's a good reason to do otherwise

packages/arb-avm-cpp/avm/src/machinethread.cpp
```diff
@@ -22,21 +22,15 @@ MachineThread::machine_status_enum MachineThread::status() {
...
     if (machine_status != MACHINE_NONE) {
         return false;
     }
 
     machine_status = MACHINE_RUNNING;
 
     machine_thread = std::make_unique<std::thread>(
-        (std::reference_wrapper<MachineThread>(*this)), max_gas, go_over_gas,
-        std::move(inbox_messages), messages_to_skip, final_message_of_block);
+        (std::reference_wrapper<MachineThread>(*this)), config);
```
hkalodner, line 33: `std::move(config)`

hkalodner (changes requested):
I marked a couple places, but in general if possible, `MachineExecutionConfig` should be taken by value and moved I think since it could contain a bunch of data

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -674,6 +685,23 @@ void ArbCore::operator()() {
                 machine = getMachineUsingStateKeys<MachineThread>(
                     *tx, pending_checkpoint.machine_state_keys, cache);
             } else {
+                // Cache pre-sideload machines
+                if (last_assertion.sideloadBlockNumber) {
+                    auto block = *last_assertion.sideloadBlockNumber;
+                    std::lock_guard<std::mutex> lock(sideload_cache_mutex);
+                    sideload_cache[block] = std::make_unique<Machine>(*machine);
+                    auto it = sideload_cache.begin();
+                    while (it != sideload_cache.end()) {
```
hkalodner, line 693: I think it'd be good to add a comment here explaining this logic (doesn't need to be too detailed)

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
```diff
@@ -25,26 +25,28 @@
...
     std::optional<uint256_t> next_block_height;
     size_t inbox_messages_consumed{0};
     uint256_t numSteps{0};
     uint256_t numGas{0};
-    bool blockingSideload{false};
     std::optional<value> fake_inbox_peek_value;
     std::vector<std::vector<uint8_t>> sends;
     std::vector<value> logs;
     std::vector<value> debug_prints;
+    std::deque<InboxMessage> sideloads;
```
hkalodner, line 44: Just confirming, is there a reason we need a `std::duque` instead of an `std::vector`?

hkalodner (changes requested):

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
PlasmaPower: You can efficiently pop from the front of it, and if you have a list of sideloads you want it's a bit of a pain to specify them in reverse order. An alternative would be keeping an index within the vector, as we do for inbox messages, but this is simpler and it doesn't need to be perfectly efficient.

PlasmaPower:
I've addressed your comments, and switched to a shared_mutex for sideload_cache_mutex now that we have C++17

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
hkalodner: Sounds good to me

hkalodner (approved):
Looks great!

PR #765, opened 2021-02-16
Skip generated abi files when running `go vet`

joshuacolvin0:


PlasmaPower (changes requested):
This needs to update `.pre-commit-config.yaml` and it doesn't seem to expand the subshell there. We might need to make a new script.

.pre-commit-config.yaml
```diff
@@ -55,6 +55,6 @@ repos:
           - 'prettier-plugin-solidity@^1.0.0-alpha.54'
       - id: go-vet
         name: Go Vet
-        entry: ./scripts/run-go-packages 'go vet ./...'
+        entry: ./scripts/run-go-packages "go vet $(go list ./... | grep -v 'arb-node-core/ethbridge[^/]*contracts')"
```
PlasmaPower, line 58: This doesn't work for me. Does it work for you?
```
package $(go is not in GOROOT (/usr/lib/go/src/$(go)
package list is not in GOROOT (/usr/lib/go/src/list)
...
```

PlasmaPower (approved):
LGTM and works

PR #769, opened 2021-02-19
C++17

hkalodner:
Switch arb-avm-cpp from c++14 to c++17


PlasmaPower (approved):
LGTM skimming through it. I will note that we don't *have* to move away from nonstd, though we should.

hkalodner:
On the same page as you there, doesn't seem totally necessary, but seems cleaner

PR #770, opened 2021-02-20
Ensure ArbStorage stays alive while ArbCore is on the Go side

PlasmaPower:
This is important as the C++ ArbStorage owns ArbCore, the Go struct just has a weak pointer.


hkalodner (approved):
LGTM

PR #774, opened 2021-02-25
Updates to ArbCore reorg handling

hkalodner:


packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
```diff
@@ -127,16 +127,14 @@ class ArbCore {
...
-                                         ValueCache& cache) const;
+    rocksdb::Status resolveStagedMessage(Transaction& tx, value& message) const;
     ValueResult<Checkpoint> getCheckpointUsingGas(Transaction& tx,
                                                   const uint256_t& total_gas,
                                                   bool after_gas);
     rocksdb::Status reorgToMessageOrBefore(
         Transaction& tx,
         const uint256_t& message_sequence_number,
-        bool reorg,
+        const bool use_latest,
```
hkalodner, line 137: I generally avoid using `const` with arguments passed by value

hkalodner (commented):
Small nitpic

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
hkalodner: Fixed

PR #775, opened 2021-02-25
Re-implement challenges on top of new rollup

PlasmaPower:
Targeted at #774 since it's built on top of it

So far I've only tested the basic bisect to an OSP case. Continued execution and/or unreachable execution should be tested too.


packages/arb-bridge-eth/contracts/challenge/Challenge.sol
```diff
@@ -222,15 +198,15 @@ contract Challenge is Cloneable, IChallenge {
...
         bytes32[] calldata _merkleNodes,
         uint256 _merkleRoute,
         uint256 _challengedSegmentStart,
         uint256 _challengedSegmentLength,
         bytes32 _oldEndHash,
         uint256 _gasUsedBefore,
         bytes32 _assertionRest
-    ) external executionChallenge onlyOnTurn {
+    ) external onlyOnTurn {
         require(_challengedSegmentLength > 1, "TOO SHORT");
```
hkalodner, line 210: There was some conversation with the auditors about potentially needing to drop this requirement. In the new formulation, I'm not totally sure if that's the case, but I think it might still be true

packages/arb-node-core/challenge/challenger.go
```diff
@@ -109,8 +70,8 @@ type SimpleChallengerImpl interface {
 type ChallengerImpl interface {
```
hkalodner, line 70: Haven't looked all the way through this code, but I think dropping this interface probably makes sense assuming there's only a single implementation. It made much more sense before the other two challenge types were dropped

packages/arb-node-core/challenge/faulty_core.go
```diff
@@ -0,0 +1,67 @@
+package challenge
```
hkalodner, line 1: I haven't looked yet, but if this is only used for tests inside the `challenge` package, it should get an `_test` at the end of the filename. If it's used elsewhere, it's good as is

hkalodner (changes requested):
A few small comments, but looks great

packages/arb-node-core/challenge/faulty_core.go
PlasmaPower: It is used elsewhere, as it has the challenger itself.

PlasmaPower: Oh nevermind I thought you were talking about the package, not the file name. Fixed.

hkalodner (commented):
Changes in repsonse to my comments all look good

hkalodner (approved):
LGTM

PR #776, opened 2021-02-25
Fix ecPairing gas cost calculation and application

PlasmaPower:


packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -840,15 +840,16 @@ uint64_t ec_pairing_variable_gas_cost(const MachineState& m) {
         return gas_cost;
     }
     try {
-        const Tuple* val = &assumeTuple(m.stack[0]);
+        auto val = m.stack[0];
```
hkalodner, line 843: I think we want `const auto&` to avoid the copy or `const value*` if we need it to be reassignable

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #777, opened 2021-02-25
Use sideload based snapshots in arb-rpc-node

PlasmaPower:
Currently breaks tests as the arb-rpc-node tests don't use arbcore, which is required for the sideload based approach


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -436,9 +436,15 @@ rocksdb::Status ArbCore::reorgToMessageOrBefore(
         pending_checkpoint = Checkpoint{};
     }
 
+    auto status =
```
hkalodner, line 439: Seems like there might be an off by one issue here. This checkpoint is still valid so it should be `pending_checkpoint.block_height` I think

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1936,8 +1981,11 @@ ValueResult<std::unique_ptr<Machine>> ArbCore::getMachineForSideload(
     // Check the cache
     {
         std::shared_lock<std::shared_mutex> lock(sideload_cache_mutex);
-        auto it = sideload_cache.find(block_number);
-        if (it != sideload_cache.end()) {
+        // Look for the first value after the value we want
```
hkalodner, line 1984: I think this should be the first value `<= block_number`

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
PlasmaPower: It should be, but C++'s APIs are stupid here and don't provide such a method, so instead we go to the first value greater than the block number, and then go back one.

hkalodner: Ahh, I read the code to fast. Makes sense

hkalodner (approved):
This looks good to me pending getting the arbostest tests working

PR #780, opened 2021-02-27
Store inbox acc more efficiently

hkalodner:
Store the inbox accumulator more efficiently to reduce the code of placing transactions in the inbox


PlasmaPower (approved):
LGTM

PR #782, opened 2021-02-27
Fix logscursor implementation and unit tests

joshuacolvin0:


packages/arb-avm-cpp/cmachine/arbcore.go
```diff
@@ -273,17 +273,18 @@ func (ac *ArbCore) LogsCursorRequest(cursorIndex *big.Int, count *big.Int) error
 	return nil
 }
 
-func (ac *ArbCore) LogsCursorGetLogs(cursorIndex *big.Int) ([]value.Value, error) {
+func (ac *ArbCore) LogsCursorGetLogs(cursorIndex *big.Int) (*big.Int, []value.Value, error) {
 	cursorIndexData := math.U256Bytes(cursorIndex)
-	result := C.arbCoreLogsCursorGetLogs(ac.c, unsafeDataPointer(cursorIndexData))
+	var firstIndex big.Int
+	result := C.arbCoreLogsCursorGetLogs(ac.c, unsafeDataPointer(cursorIndexData), unsafeDataPointer(firstIndex.Bytes()))
```
hkalodner, line 279: I don't think that writing to `firstIndex` is safe since it's a complex type. Instead we'd need a [32]byte to write into

hkalodner: Looks like `LogsCursorGetDeletedLogs` has the same issue

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
```diff
@@ -178,9 +178,12 @@ class ArbCore {
    public:
     // Logs Cursor interaction
     bool logsCursorRequest(size_t cursor_index, uint256_t count);
-    std::optional<std::vector<value>> logsCursorGetLogs(size_t cursor_index);
+    std::optional<std::vector<value>> logsCursorGetLogs(
```
hkalodner, line 181: I'm not a fan out using output parameters this way. I think it'd be better to use an `std::pair` or declare a new struct type for this and `logsCursorGetDeletedLogs`

packages/arb-avm-cpp/data_storage/include/data_storage/aggregator.hpp
```diff
@@ -43,6 +45,8 @@ class AggregatorStore {
     void saveBlockHash(const uint256_t& block_hash, uint64_t block_height);
 
     void reorg(uint64_t block_height);
+    ValueResult<uint256_t> logsProcessedCount() const;
+    rocksdb::Status updateLogsProcessedCount(uint256_t& count);
```
hkalodner, line 49: Seems like this should have a const ref argument instead of a mutable ref

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1833,18 +1841,17 @@ std::optional<std::vector<value>> ArbCore::logsCursorGetLogs(
...
-        logs_cursors[cursor_index].data.size();
-
-    std::vector<value> logs{std::move(logs_cursors[cursor_index].data)};
+    std::vector<value> logs = std::move(logs_cursors[cursor_index].data);
     logs_cursors[cursor_index].data.clear();
 
-    return logs;
+    first_index_out = logs_cursors[cursor_index].current_total_count;
+
+    return std::move(logs);
```
hkalodner, line 1849: `std::move` on returns kills NRVO. Just use `return logs` instead here and in `logsCursorGetDeletedLogs`

packages/arb-rpc-node/txdb/txdb.go
```diff
@@ -95,6 +95,82 @@ func (db *TxDB) GetBlockResults(res *evm.BlockInfo) ([]*evm.TxResult, error) {
...
+	return nil
+}
+
+func (db *TxDB) DeleteLogs(avmLogs []value.Value) error {
+	// Collect all logs that will be removed so they can be sent to rmLogs subscription
+	lastResultIndex := len(avmLogs) - 1
+	var currentBlockHeight uint64
+	blocksFound := false
+	for i, _ := range avmLogs {
+		// Parse L2 blocks in reverse
```
hkalodner, line 128: L2 transaction receipts rather than blocks I think

packages/arb-util/core/logreader.go
```diff
@@ -90,16 +95,44 @@ func (lr *LogReader) getLogs(ctx context.Context) error {
...
 			}
 		}
 
-		if len(deletedLogs) > 0 {
-			err = lr.consumer.DeleteLogs(deletedLogs)
-			if err != nil {
+		if len(logs) > 0 {
+			cmp := firstIndex.Cmp(currentLogCount)
+			if cmp == 1 {
+				// Some logs were skipped, pull these manually
```
hkalodner, line 129: How is it possible to end up in this case. Seems like maybe it should be an error

packages/arb-util/core/logreader.go
```diff
@@ -116,20 +149,24 @@ func (lr *LogReader) getLogs(ctx context.Context) error {
...
+			// Got deleted logs successfully
 			if len(newDeletedLogs) > 0 {
 				err = lr.consumer.DeleteLogs(newDeletedLogs)
 				if err != nil {
 					return err
 				}
+			} else {
+				// Logs not provided to delete
+				// TODO
+				_ = firstNewDeletedIndex
```
hkalodner, line 169: I'm not sure what this todo is for. It doesn't seem like there's anything we can do in this case

hkalodner (changes requested):

packages/arb-avm-cpp/cmachine/arbcore.go
joshuacolvin0: took care of by returning value in struct

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
joshuacolvin0: fixed

packages/arb-avm-cpp/data_storage/include/data_storage/aggregator.hpp
joshuacolvin0: done

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: refactor made this not applicable

packages/arb-rpc-node/txdb/txdb.go
joshuacolvin0: fixed

packages/arb-util/core/logreader.go
joshuacolvin0: I'm happy to treat it as an error

packages/arb-util/core/logreader.go
joshuacolvin0: removed

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
```diff
@@ -178,12 +178,10 @@ class ArbCore {
    public:
     // Logs Cursor interaction
     bool logsCursorRequest(size_t cursor_index, uint256_t count);
-    std::optional<std::vector<value>> logsCursorGetLogs(
-        size_t cursor_index,
-        uint256_t& first_index_out);
-    std::optional<std::vector<value>> logsCursorGetDeletedLogs(
-        size_t cursor_index,
-        uint256_t& first_index_out);
+    std::pair<uint256_t, std::optional<std::vector<value>>> logsCursorGetLogs(
```
hkalodner, line 181: I think a return type of `std::optional<std::pair<uint256_t, std::vector<value>>>` would be cleaner

hkalodner (changes requested):
One more comment. Everything else looks good

hkalodner (approved):
LGTM

PR #783, opened 2021-02-27
Update to arbos version with reordered logs

hkalodner:
- Update ArbOS with https://github.com/OffchainLabs/arb-os/pull/359
- Update ArbOS test for new version


edfelten (approved):
Looks good.

PR #784, opened 2021-02-27
Expose inboxAcc to Go

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -683,13 +693,6 @@ void ArbCore::operator()() {
             auto tx = Transaction::makeTransaction(data_storage);
 
             auto last_assertion = machine->nextAssertion();
-            auto messages_inserted = messageEntryInsertedCountImpl(*tx);
```
hkalodner, line 686: I was wondering about this deletion and then I saw it was dead code 😁

hkalodner (approved):
LGTM

PR #785, opened 2021-02-28
Update monorepo for new arbos charging system

hkalodner:
This PR is a WIP while the arbos PR is still under review

- Updates ArbOS to https://github.com/OffchainLabs/arb-os/pull/355
- Update monorepo to support arbos update


edfelten (approved):
Looks good.

PR #786, opened 2021-02-28
Allow L2 token addresses to be calculated on L1

dmihal:
Some L1 contracts may need to know the address of a corresponding token on L2.

Since we use Create2 to deploy the tokens on L2, we can easily calculate the address in L1.

This PR also means that the L2 template contracts must be manually deployed on L2 before the bridge is set up. Aside from allowing this calculation, it also means that the L1 bridge contract isn't larger than the contract size limit.

### DIFF
--- a/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
+++ b/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
@@ -18,12 +18,13 @@
 
 pragma solidity ^0.6.11;
 
+import "@openzeppelin/contracts/proxy/Clones.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "./L1Buddy.sol";
 import "../arbitrum/ArbTokenBridge.sol";
 
 import "./IExitLiquidityProvider.sol";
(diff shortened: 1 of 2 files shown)


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -34,18 +35,32 @@ contract EthERC20Bridge is L1Buddy {
...
+    }
 
     function connectToChain(uint256 maxGas, uint256 gasPriceBid) external payable {
         // Pay for gas
         if (msg.value > 0) {
             inbox.depositEth{ value: msg.value }(address(this));
         }
+        bytes memory deployCode = abi.encodePacked(
+            type(ArbTokenBridge).creationCode,
+            abi.encode(l2TemplateERC20, l2TemplateERC777)
```
fredlacs, line 57: Should this maybe be encodePacked?

dmihal: I believe this one should be abi.encode, since these are the constructor parameters

fredlacs:
Merge conflicts were solved in separate PR https://github.com/OffchainLabs/arbitrum/pull/890

PR #789, opened 2021-03-01
Change single param Tuple constructors to factories

joshuacolvin0:
There were multiple single parameter overloaded Tuple constructors that have already caused a few problems.


hkalodner (approved):
LGTM

PR #798, opened 2021-03-02
Arb-ts w/ bridge wrapper class & update tooling for typechain 

DZGoldman:
Bridge & tests in arb-ts are WIP, but can be merged as is.


package.json
```diff
@@ -15,8 +15,8 @@
   "scripts": {
     "install:deps": "./scripts/install-deps",
     "install:validator": "./scripts/install-validator",
-    "update-abi": "yarn go:generate && yarn workspace arb-provider-ethers update-abi",
-    "build": "yarn workspace arb-provider-ethers build",
+    "update-abi": "yarn go:generate && yarn workspace arb-ts update-abi && arb-provider-ethers update-abi",
```
fredlacs, line 18: Should we rename to `update:abi` in order to keep consistent with other script names?

package.json
```diff
@@ -46,7 +46,9 @@
       "packages/arb-bridge-eth",
       "packages/arb-provider-ethers",
```
fredlacs, line 47: I think we're ready to remove the old provider package

package.json
```diff
@@ -58,7 +60,7 @@
     "prettier-plugin-solidity": "^1.0.0-alpha.54"
   },
   "resolutions": {
-    "arb-provider-ethers/ethers": "~4.0.47",
+    "arb-provider-ethers/ethers": "~5.0.24",
```
fredlacs, line 63: This won't be needed anymore 😄

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/DecimalConverter.sol
```diff
@@ -4,13 +4,13 @@
 pragma solidity ^0.6.0;
 
 library DecimalConverter {
-  function from777to20(uint256 amount, uint8 decimals) internal pure returns (uint256) {
-    require(decimals <= 18, 'DEC');
-    return amount / (10 ** uint256(18 - decimals));
-  }
+    function from777to20(uint8 decimals, uint256 amount) internal pure returns (uint256) {
```
fredlacs, line 7: Seems like a regression. The contracts using this library assume the amount is the first parameter.

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/DecimalConverter.sol
```diff
@@ -4,13 +4,13 @@
...
+        require(decimals <= 18, "DEC");
+        return amount / (10**uint256(18 - decimals));
+    }
 
-  function from20to777(uint256 amount, uint8 decimals) internal pure returns (uint256) {
-    require(decimals <= 18, 'DEC');
-    return amount * (10 ** uint256(18 - decimals));
-  }
+    function from20to777(uint8 decimals, uint256 amount) internal pure returns (uint256) {
```
fredlacs, line 12: Same issue, flipped parameters.

packages/arb-ts/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "preversion": "yarn lint",
+    "prepare": "install-peers",
+    "build": "./scripts/builder",
+    "watch": "tsc --watch",
+    "test": "mocha",
+    "test:coverage": "nyc mocha",
+    "test:ci": "nyc --reporter=lcovonly mocha --reporter xunit",
+    "lint": "eslint .",
+    "format": "prettier './**/*.{js,json,md,ts,yml}' --write && yarn run lint --fix",
+    "update-abi": "./scripts/update-abi && yarn format",
```
fredlacs, line 29: `update:abi`?

packages/arb-ts/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "int_test": "yarn run mocha integration_test/arb-bridge.test.ts --timeout 15000"
```
fredlacs, line 30: I think `test:integration` would make for a better name

packages/arbos-contracts/package.json
```diff
@@ -14,18 +14,23 @@
   },
   "homepage": "https://offchainlabs.com/",
   "scripts": {
-    "build": "truffle compile"
+    "build": "hardhat compile",
+    "typechain": "hardhat typechain"
```
fredlacs, line 18: I'm assuming we need this to create the typechain for the arb-os precompiles. Can't we access those through the git submodule? If a regular import doesn't work I've got a few hacky fun ideas - which might be better than needing to keep these two in sync.

fredlacs: And if we do keep this workspace, we could rename it to `arb-os-contracts`?

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    this.arbERC20Bridge = ArbTokenBridgeFactory.connect(
+      arbERC20BridgeAddress,
+      l2Signer
+    )
+  }
+
+  public async withdrawETH(value: BigNumber, destinationAddress?: string) {
+    const address = destinationAddress || (await this.getWalletAddress())
+    this.arbSys.withdrawEth(address, {
```
fredlacs, line 51: missing a return here

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,80 @@
...
+    })
+  }
+
+  public async withdrawERC20(
+    erc20l1Address: string,
+    amount: BigNumber,
+    destinationAddress?: string
+  ) {
+    const destination = destinationAddress || (await this.getWalletAddress())
+    return await this.arbERC20Bridge.withdraw(
```
fredlacs, line 62: We probably shouldn't await and instead just return the actual promise so that the library's consumer can deal with the promise resolving.

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  public async depositERC20(
+    erc20L1Address: string,
+    amount: BigNumber,
+    maxGas: BigNumber,
+    gasPriceBid: BigNumber,
+    destinationAddress?: string
+  ) {
+    const destination = destinationAddress || (await this.getWalletAddress())
+    this.ethERC20Bridge.depositAsERC20(
```
fredlacs, line 69: Missing a return

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,90 @@
...
+      erc20L1Address,
+      destination,
+      amount,
+      maxGas,
+      gasPriceBid
+    )
+  }
+
+  public async getERC20L2Address(erc20L1Address: string) {
+    return await this.ethERC20Bridge.customL2Tokens(erc20L1Address)
```
fredlacs, line 79: We should check if there is a customL2Token that was assigned, if not then we can calculate the create2 address for erc20. 

```
const l2Address = ethers.utils.getCreate2Address(
      bridge.address,
      ethers.utils.hexZeroPad(l1ERC20, 32),
      ethers.utils.keccak256(proxyBytecode)
)

File: packages/arb-ts/src/lib/bridge.ts [hunk 13]
```diff
@@ -0,0 +1,56 @@
...
+    amount: BigNumber,
+    destinationAddress?: string
+  ) {
+    return await this.l2Bridge.withdrawERC20(
+      erc20l1Address,
+      amount,
+      destinationAddress
+    )
+  }
+  public async getERC20LlAddress(erc20L2Address: string) {
```
fredlacs, line 53: `getERC20LlAddress` typoed the 1

packages/arb-ts/src/lib/bridge.ts
```diff
@@ -0,0 +1,56 @@
...
+    return await this.l2Bridge.getERC20LlAddress(erc20L2Address)
```
fredlacs, line 54: typoed `getERC20LlAddress`

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,80 @@
...
+      erc20l1Address,
+      destination,
+      amount
+    )
+  }
+
+  public async getERC20LlAddress(erc20L2Address: string) {
```
fredlacs, line 69: typoed `getERC20LlAddress`

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,80 @@
...
+    return await this.arbERC20Bridge.customToken(erc20L2Address)
```
fredlacs, line 70: I don't believe this logic will work to get the token's L1 address

fredlacs (changes requested):

package.json
DZGoldman: soon, but let's hold off on removing until we update the remaining scripts depencing on it (i.e., setup_validators_demo)

package.json
DZGoldman: keeping as per https://github.com/OffchainLabs/arbitrum/pull/798#discussion_r585903281 (altho the version update shouldn't be there!)

packages/arb-ts/src/lib/bridge.ts
DZGoldman: weird

package.json
hkalodner: I think that'd be good

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/DecimalConverter.sol
DZGoldman: ah good catch (rebase issue: linter made some changes, but didn't notice the changes beyond the lint)

packages/arb-ts/src/lib/l2Bridge.ts
DZGoldman: So looking back - this does assume that `notifyCustomToken` has been called for the given token, which isn't guaranteed (else it'll return address(0) )

DZGoldman: what about calling it `attemptGetERC20L1Address` or something?

packages/arb-ts/src/lib/bridge.ts
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2019-2020, Offchain Labs, Inc.
```
hkalodner, line 2: We should change these to 2021

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (!this.l1Signer || !this.inbox) {
+      throw new Error('No eth signer provider')
+    }
+    const address = destinationAddress || (await this.getWalletAddress())
+    return this.inbox.depositEth(address, {
+      value,
+    })
+  }
+
+  public async depositERC20(
```
hkalodner, line 61: I think we want `PayableOverrides` for paying for L2 gas

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,90 @@
...
+    return await this.ethERC20Bridge.customL2Tokens(erc20L1Address)
+  }
+
+  public async getWalletAddress() {
+    const { walletAddressCache } = this
+    if (walletAddressCache) {
+      return walletAddressCache
+    }
+
+    return await this.l1Signer.getAddress()
```
hkalodner, line 88: We should populate the cache if it's not set probably

packages/arb-ts/src/lib/l2Bridge.ts
hkalodner: We should probably add a function to fetch this. Actually this raises a question about the custom token thing which is do we expect it to be ERC20 or ERC777. I think we might need to have the function that specifies to use a custom token also specify the type

packages/arbos-contracts/package.json
hkalodner: Ah yeah, this should really be in arb-os. Would probably make sense to make a PR there to get that setup

hkalodner (changes requested):

packages/arb-ts/src/lib/l2Bridge.ts
DZGoldman: r.e. "add a function to fetch this," isn't `notifyCustomToken` + this the only way? iow, it has to be pushed up to the L2 at some point

packages/arb-ts/src/lib/l1Bridge.ts
DZGoldman: is including `value` not enough?

packages/arb-ts/src/lib/l1Bridge.ts
DZGoldman: true, tho the likely intended use is to interact with it; it should maybe be more of a 'get address if contract is deployed' sort of method. idk

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -66,15 +60,15 @@ export class L2Bridge {
     )
   }
 
-  public async getERC20LlAddress(erc20L2Address: string) {
-    return await this.arbERC20Bridge.customToken(erc20L2Address)
+  public getERC20L1Address(erc20L2Address: string) {
+    return this.arbERC20Bridge.customToken(erc20L2Address)
```
hkalodner, line 64: This seems off. Presumably it should use the calculated address if it's not custom

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -75,16 +61,24 @@ export class L1Bridge {
     )
   }
 
-  public async getERC20L2Address(erc20L1Address: string) {
-    return await this.ethERC20Bridge.customL2Tokens(erc20L1Address)
+  public getERC20L2Address(erc20L1Address: string) {
+    return this.ethERC20Bridge.customL2Tokens(erc20L1Address)
```
hkalodner, line 65: This seems off. Presumably it should use the calculated address if it's not custom

hkalodner (changes requested):

DZGoldman:
Ready for re-review; bridge stuff in arb-ts still needs a few more things, but can be merged as is / close to as is so as to get everything else moving.

(Particularly, custom token handling isn't fully there yet; that will still require some contract-level modification, i.e., a depositCustomToken method)

DZGoldman:
Requires https://github.com/OffchainLabs/arb-os/pull/381
should be in sync with paths of latest arb-os now (h/t @fredlacs)

DZGoldman:
ooo one TODO that should probably just be included here before more is autogenerating the buddy L2 address directly from the L1 address (instead of requiring it as a parameter) — what's the algo to do so?

packages/arb-ts/scripts/update-abi
```diff
@@ -0,0 +1,30 @@
...
+
+echo 'Copying interfaces into arb-ts'
+
+cd ../arb-bridge-eth/build/types
+
+\cp -rf Outbox* Bridge* Inbox* Rollup* ../../../arb-ts/src/lib/abi
+            
+cd ../../../arb-os/contracts/src/types
+
+# TODO: why does arb-os contracts generate factories w/ this different structure
```
fredlacs, line 22: Maybe different hardhat / typechain versions?

fredlacs (approved):
LGTM

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,193 @@
...
+          erc20L1Address,
+          this.l1Signer
+        )
+        const balance = await ethERC20TokenContract.balanceOf(walletAddress)
+
+        const allowance = await ethERC20TokenContract.allowance(
+          walletAddress,
+          indboxAddress
+        )
+        const allowed = await allowance.gte(MIN_APPROVAL.div(2))
```
hkalodner, line 96: It might make more sense to store the current allowance and then add a function to calculate allowed if it's useful

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,193 @@
...
+        }
+      } else {
+        console.warn(`ERC20 at ${erc20L1Address} not found on L1`)
+      }
+    } else {
+      const ethERC20TokenContract = await ERC20Factory.connect(
+        erc20L1Address,
+        this.l1Signer
+      )
+      const balance = await ethERC20TokenContract.balanceOf(walletAddress)
```
hkalodner, line 110: Looks like these run in both the if and the else. Probably should just move them out of the conditional

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,193 @@
...
+      if (!tokenData.ERC20.allowed) {
+        const allowance = await ethERC20TokenContract.allowance(
+          walletAddress,
+          indboxAddress
+        )
+        tokenData.ERC20.allowed = allowance.gte(MIN_APPROVAL.div(2))
+      }
+    }
+
+    // TODO: erc777? meh
```
hkalodner, line 122: This would be for L1 support? Agreed we can push worrying about that

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -0,0 +1,193 @@
...
+    return tokenData
+  }
+
+  public async depositETH(value: BigNumber, destinationAddress?: string) {
+    const address = destinationAddress || (await this.getWalletAddress())
```
hkalodner, line 127: Love the API here

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,206 @@
...
+        )
+      }
+    } else {
+      const arbERC20TokenContract = await StandardArbERC20Factory.connect(
+        l2ERC20Address,
+        this.l2Signer
+      )
+      const balance = await arbERC20TokenContract.balanceOf(walletAddress)
+      tokenData.ERC20.balance = balance
+    }
```
hkalodner, line 156: This seems to be the same as the L1 side. It would be awesome if we could share code between them

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -0,0 +1,206 @@
...
+    return this.l2Tokens
+  }
+
+  public async getAndUpdateTokenData(erc20L1Address: string) {
+    if (!this.l2Tokens[erc20L1Address]) {
+      this.l2Tokens[erc20L1Address] = {
+        ERC20: undefined,
+        ERC777: undefined,
+        CUSTOM: undefined,
```
hkalodner, line 124: I know this is a known issue, but `CUSTOM` still needs to be handled here

hkalodner (changes requested):

PR #804, opened 2021-03-03
Ethbridge deployment

hkalodner:
Get ethbridge deployment working


packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -91,7 +91,7 @@ contract Rollup is Cloneable, RollupCore, Pausable, IRollup {
         rollupEventBridge.rollupInitialized(
             _confirmPeriodBlocks,
             _extraChallengeTimeBlocks,
-            _arbGasSpeedLimitPerBlock,
+            _arbGasSpeedLimitPerBlock / 2,
```
fredlacs, line 94: Seems like a typo? If we need to divide by 2 when delivering to the inbox, shouldn't we probably do that inside the RollupEventBridge logic?
We still store it here `arbGasSpeedLimitPerBlock = _arbGasSpeedLimitPerBlock`

fredlacs (changes requested):

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
hkalodner: Ah good catch. This was a quick hack around the fact that arbos was expecting the speed limit per second instead of per block, but that's been adjusted

hkalodner: Fixed

fredlacs (approved):
LGTM

PR #812, opened 2021-03-04
Fix node conflict logic

hkalodner:


PlasmaPower (approved):
LGTM and seems to work

PR #820, opened 2021-03-04
make saveBlock atomic and implement save/getMessageBatch

joshuacolvin0:


packages/arb-avm-cpp/cavm/caggregator.h
```diff
@@ -52,10 +52,18 @@ typedef struct CRequestInfoStruct CRequestInfo;
 void deleteAggregatorStore(CAggregatorStore* m);
 
 Uint64Result aggregatorBlockCount(const CAggregatorStore* agg);
-int aggregatorSaveBlock(CAggregatorStore* agg,
```
hkalodner, line 55: Need to remove `aggregatorSaveBlockHash` and `aggregatorSaveRequest` as well

packages/arb-avm-cpp/data_storage/src/aggregator.cpp
```diff
@@ -186,15 +176,63 @@ uint64_t AggregatorStore::blockCount() const {
...
+    commitTx(*tx);
+}
+
+uint64_t AggregatorStore::getMessageBatch(const uint256_t& batchNum) {
+    auto tx = data_storage->beginTransaction();
+    auto full_key = messageBatchKey(batchNum);
+
+    std::string value;
+    auto status =
+        tx->GetForUpdate(rocksdb::ReadOptions{}, vecToSlice(full_key), &value);
```
hkalodner, line 199: I think `Get` here is fine because we're just gonna return this value and close the tx

hkalodner (changes requested):

packages/arb-avm-cpp/cavm/caggregator.h
joshuacolvin0: done

packages/arb-avm-cpp/data_storage/src/aggregator.cpp
joshuacolvin0: done

packages/arb-avm-cpp/cmachine/nodestorage_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestMessageBatch(t *testing.T) {
+	dePath := "dbPath"
+
+	if err := os.RemoveAll(dePath); err != nil {
+		logger.Error().Stack().Err(err).Send()
+		t.Fatal(err)
+	}
+
+	arbStorage, err := NewArbStorage("dbPath")
```
hkalodner, line 33: Should probably be `dbPath` rather than the literal

packages/arb-avm-cpp/cmachine/nodestorage_test.go
```diff
@@ -0,0 +1,61 @@
...
+		t.Fatal(err)
+	}
+	if logIndex != testLogIndex {
+		logger.Error().Msg("logIndex doesnt match testLogIndex")
+	}
+
+	if err := os.RemoveAll(dePath); err != nil {
+		logger.Error().Stack().Err(err).Send()
+		t.Fatal(err)
+	}
```
hkalodner, line 60: I think this should be in a defer block near the top so it'll delete the db on failure as well

hkalodner (changes requested):

packages/arb-avm-cpp/cmachine/nodestorage_test.go
joshuacolvin0: 👍

packages/arb-avm-cpp/cmachine/nodestorage_test.go
joshuacolvin0: 👍

hkalodner (approved):
LGTM

PR #822, opened 2021-03-05
Staker fixes 4: electric boogaloo

PlasmaPower:


packages/arb-util/core/execution.go
```diff
@@ -145,8 +145,15 @@ func IsAssertionValid(assertion *Assertion, execTracker *ExecutionTracker, targe
...
-	notEnoughMessages := localExecutionInfo.InboxMessagesRead().Cmp(assertion.InboxMessagesRead()) < 0
-	if notEnoughMessages {
+	if localExecutionInfo.InboxMessagesRead().Cmp(assertion.InboxMessagesRead()) < 0 {
+		// We didn't read enough messages.
+		// This can either mean that our messages lasted longer, or that we are missing messages.
+		if localExecutionInfo.After.TotalGasConsumed.Cmp(assertion.After.TotalGasConsumed) < 0 {
+			// This means we stopped because we're missing messages,
+			// but the on-chain rollup must've had these messages.
+			// Error and try again when we have the messages.
+			return false, errors.New("Missing messages to evaluate assertion")
```
PlasmaPower, line 155: I'd appreciate if someone thought through this logic and made sure it's sound. I think it is but I might've missed an edge case.

packages/arb-node-core/staker/staker_test.go
```diff
@@ -192,11 +212,36 @@ func TestStaker(t *testing.T) {
...
+	}
+}
+
+func TestChallengeRegular(t *testing.T) {
+	adversarialTest(t, challenge.FaultConfig{DistortMachineAtGas: big.NewInt(10000)}, big.NewInt(100000000))
+}
+
+func TestChallengeTimeout(t *testing.T) {
+	adversarialTest(t, challenge.FaultConfig{DistortMachineAtGas: big.NewInt(50000)}, big.NewInt(100000))
+}
```
hkalodner, line 243: Can we actually verify in the test how the challenge ended? Easiest way would probably be via checking to see if the expected was emitted

packages/arb-util/core/execution.go
hkalodner: I think the other possibility here is that we're in at stopped or errored state

hkalodner (changes requested):

packages/arb-util/core/lookup.go
```diff
@@ -202,9 +202,11 @@ func NewExecutionState(c ExecutionCursor) *ExecutionState {
 }
 
 func (e *ExecutionState) Equals(o *ExecutionState) bool {
+	// We don't check InboxAcc here intentionally.
+	// InboxAcc isn't present for assertions,
+	// and it's part of MachineHash anyways
```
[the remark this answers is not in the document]
PlasmaPower: Yeah, I should've phrased that better. I meant that any relevant change to the inbox messages read will be reflected in the machine hash, and that we aren't asserting the inbox acc, it's computed by inbox.

PlasmaPower:
I've resolved your comments

hkalodner (approved):
LGTM, the staker test is looking awesome!

PR #826, opened 2021-03-05
Outbox proof

hkalodner:
- Add support for outbox proofs
- Add integration test for L2 to L1 transactions


packages/arb-evm/evm/merkleRootResult.go
```diff
@@ -136,9 +182,47 @@ func newMerkleTreeFromValue(val value.Value) (MerkleNode, error) {
...
 	}
 }
+
+type MerkleRootProof struct {
+	Nodes []common.Hash
+	Path  []bool
+	Data  []byte
+}
+
+func (m *MerkleRootResult) GenerateProof(index uint64) (*MerkleRootProof, error) {
```
PlasmaPower, line 199: We already have an existing merkle tree and proof generation system somewhere in the ethbridge MerkleTree, is that fundamentally incompatible with this in some way? I'd prefer to only have one merkle tree system in Go.

hkalodner: Agreed that it'd be nice to limit these. The difference is basically that this tree is produced by ArbOS using a different structure and unifying them would require significantly altering ArbOS

PlasmaPower: Ah, I was wondering about that. Makes sense to keep it as-is then.

hkalodner: At least we use the same prover on the solidity side which minimizes the issues there and means we stick to a unified proving strategy

packages/arb-avm-cpp/cmachine/nodestorage_test.go
```diff
@@ -52,12 +52,11 @@ func TestMessageBatch(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	logIndex, err := nodeStore.GetMessageBatch(testBatchNumber)
-	if err != nil {
-		logger.Error().Stack().Err(err).Send()
+	logIndex := nodeStore.GetMessageBatch(testBatchNumber)
+	if logIndex == nil {
 		t.Fatal(err)
```
PlasmaPower, line 57: This should probably be `t.Fatal("GetMessageBatch failed")` or something

packages/arb-util/arbos/arbos.go
```diff
@@ -16,6 +16,8 @@ var ARB_BLS_ADDRESS = ethcommon.HexToAddress("0x00000000000000000000000000000000
 var ARB_FUNCTION_TABLE_ADDRESS = ethcommon.HexToAddress("0x0000000000000000000000000000000000000068")
 var ARB_OWNER_ADDRESS = ethcommon.HexToAddress("0x000000000000000000000000000000000000006B")
 
+var ARB_NODE_INTERFACE_ADDRESS = ethcommon.HexToAddress("0x000000000000000000000000000000000000006E")
```
PlasmaPower, line 19: I'd put this at a much higher address to differentiate it and not conflict with future arbos "contracts"

PlasmaPower (changes requested):

packages/arb-util/arbos/arbos.go
hkalodner: Agree with this, but we'll need to charge ArbOS. I'll flag @edfelten on this question

hkalodner: This was changed to 200 to keep it near the arbos address block, but leave plenty of room for future expansion

PlasmaPower (approved):
LGTM

PR #831, opened 2021-03-06
Fix OSP challenge tests which broke with an ArbOS upgrade

PlasmaPower:
They should be more reliable with the new gas points, which hopefully don't assume anything about the instructions ArbOS contains


hkalodner (approved):
LGTM

PR #835, opened 2021-03-07
Reimplement findStakerConflict in a more efficient manner

PlasmaPower:


PlasmaPower:
Against malicious adversaries, it seems the original algorithm has preferable costs

PR #839, opened 2021-03-07
Remove buddy

hkalodner:
Update arbos to complete removal of builtin buddy contract facility in favor of buddy contract pattern built over the generic transaction bridge


joshuacolvin0 (approved):
LGTM

PR #843, opened 2021-03-08
Get rid of inbox reader error channel

PlasmaPower:
The inbox reader was blocking on this channel when it wasn't used.


hkalodner (approved):
LGTM

PR #844, opened 2021-03-08
Adding maximal modified index to buffers, variable gas costs

mrsmkl:
Need to clean up some...


packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -1091,6 +1091,29 @@ void getbuffer256(MachineState& m) {
...
+    if (m.stack.stacksize() < 3) {
+        return 0;
+    }
+    auto offset = std::get_if<uint256_t>(&m.stack[0]);
+    auto buf = std::get_if<Buffer>(&m.stack[2]);
+    if (!buf || !offset) {
+        return 0;
+    }
+    if (*offset + inc > std::numeric_limits<uint64_t>::max()) {
+        return 0;
```
edfelten, line 1104: The instruction should have a defined non-zero ArbGas cost in error cases like this.

packages/arb-avm-cpp/avm_values/include/avm_values/buffer.hpp
```diff
@@ -210,25 +210,31 @@ class RawBuffer {
...
+    Buffer() {
+        buf = std::make_shared<RawBuffer>(); 
+        maxAccess = 0;
+    }
 
     Buffer(const std::vector<uint8_t>& data) : Buffer() {
         for (uint64_t i = 0; i < data.size(); i++) {
             buf = std::make_shared<RawBuffer>(buf->set(i, data[i]));
         }
+        maxAccess = data.size();
```
edfelten, line 229: There is an off-by-one issue here.  In most of the code, `maxAccess` is the maximum index that was modified. But here, `data.size()` returns the maximum index in `data`, plus 1.

Obviously, either approach is OK, but we should be consistent.

packages/arb-avm-cpp/avm_values/include/avm_values/opcodes.hpp
```diff
@@ -487,12 +487,12 @@ const std::unordered_map<OpCode, uint64_t> InstructionArbGasCost = {
     {OpCode::DEBUG_PRINT, 1},
 
     {OpCode::NEW_BUFFER, 1},
-    {OpCode::GET_BUFFER8, 10},
-    {OpCode::GET_BUFFER64, 10},
-    {OpCode::GET_BUFFER256, 10},
-    {OpCode::SET_BUFFER8, 100},
-    {OpCode::SET_BUFFER64, 100},
-    {OpCode::SET_BUFFER256, 100},
+    {OpCode::GET_BUFFER8, 3},
```
edfelten, line 490: Best to leave a comment here, explaining the logic behind how gas costs of getBuffer ops were determined.  Either they should depend on the set of the buffer (reflecting the need to walk a tree), or set to a constant equal to the cost with a maximum-size tree.

packages/arb-avm-cpp/avm_values/include/avm_values/opcodes.hpp
```diff
@@ -487,12 +487,12 @@ const std::unordered_map<OpCode, uint64_t> InstructionArbGasCost = {
...
+    {OpCode::GET_BUFFER64, 3},
+    {OpCode::GET_BUFFER256, 3},
+    {OpCode::SET_BUFFER8, 240},
```
edfelten, line 493: Best to leave a comment saying these are placeholders, and the real gas cost will be computed dynamically based on the state of the buffer.

packages/arb-avm-cpp/avm_values/src/buffer.cpp
```diff
@@ -19,12 +19,21 @@
 
 #include <ethash/keccak.hpp>
 
-uint256_t zero_hash(uint64_t sz) {
-    if (sz == 5) {
-        return hash(0);
+struct Zeros {
```
edfelten, line 22: Name should be more descriptive.

packages/arb-avm-cpp/avm_values/src/buffer.cpp
```diff
@@ -159,7 +168,7 @@ std::vector<RawBuffer> RawBuffer::serialize(
     if (level > 0) {
         value_vector.push_back(1);
         for (uint64_t i = 0; i < NODE_SIZE; i++) {
-            uint256_t hash_ = hash2(123, (*node)[i].hash());
+            uint256_t hash_ = hash2(1234, (*node)[i].hash());
```
edfelten, line 171: Presumably the maxAccess field should also affect the hash.  Use that instead of `1234`?

packages/arb-avm-cpp/avm_values/include/avm_values/buffer.hpp
```diff
@@ -210,25 +210,31 @@ class RawBuffer {
 class Buffer {
    public:
     std::shared_ptr<RawBuffer> buf;
+    uint64_t maxAccess;
```
edfelten, line 213: Should this be the maximum index accessed, or the maximum index written?   I think written makes more sense, because we don't need to build the tree for reads beyond the end of the tree.

packages/arb-bridge-eth/contracts/arch/Hashing.sol
```diff
@@ -129,7 +129,7 @@ library Hashing {
         } else if (val.typeCode == Value.hashOnlyTypeCode()) {
             return bytes32(val.intVal);
         } else if (val.typeCode == Value.bufferTypeCode()) {
-            return keccak256(abi.encodePacked(uint256(123), val.bufferHash));
+            return keccak256(abi.encodePacked(val.size, val.bufferHash));
```
edfelten, line 132: This is inconsistent with the hashing done elsewhere in this PR.  The approach here seems like the correct one.

edfelten (changes requested):
Several comments and changes requested.

packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
mrsmkl: In those cases the cost will be the fixed cost from `opcodes.hpp` just like for pairing...

packages/arb-avm-cpp/avm_values/include/avm_values/buffer.hpp
mrsmkl: Yeah, best change the name.

packages/arb-avm-cpp/avm_values/src/buffer.cpp
mrsmkl: This is for serializing inner nodes in the tree, so there's no `maxAccess` for them. I'll try to make that clearer.

packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -1091,6 +1091,30 @@ void getbuffer256(MachineState& m) {
...
+    // The fixed cost of the the buffer opcodes is applied elsewhere
+    if (m.stack.stacksize() < 3) {
+        return 0;
+    }
+    auto offset = std::get_if<uint256_t>(&m.stack[0]);
+    auto buf = std::get_if<Buffer>(&m.stack[2]);
+    if (!buf || !offset) {
+        return 0;
+    }
+    if (*offset + inc > std::numeric_limits<uint64_t>::max()) {
```
PlasmaPower, line 1104: I think this check is a bit different than the one in Solidity. It doesn't check if offset + inc overflows a uint256, and it doesn't check m.stack[1]. If this check isn't identical to the checks done before deducting the variable gas in Solidity, it could cause a divergence.

packages/arb-bridge-eth/contracts/arch/OneStepProof2.sol
```diff
@@ -522,9 +551,20 @@ contract OneStepProof2 is OneStepProofCommon {
             handleOpcodeError(context);
             return;
         }
+    
+        uint256 mx = max(val1.size, val2.intVal+7);
+
+        if (deductGas(context, uint64(2*setbuffer_extra_gas_cost(mx)))) {
+            // When we run out of gas, we only charge for an error + gas_set
+            // That means we need to deduct the previously charged base cost here
+            context.gas -= SETBUFFER_GAS_COST;
```
PlasmaPower, line 560: Since this is for setBuffer64 the gas cost deducted here needs to be doubled.

packages/arb-bridge-eth/contracts/arch/OneStepProof2.sol
```diff
@@ -539,9 +579,20 @@ contract OneStepProof2 is OneStepProofCommon {
             handleOpcodeError(context);
             return;
         }
+
+        uint256 mx = max(val1.size, val2.intVal+31);
+    
+        if (deductGas(context, uint64(2*setbuffer_extra_gas_cost(mx)))) {
+            // When we run out of gas, we only charge for an error + gas_set
+            // That means we need to deduct the previously charged base cost here
+            context.gas -= SETBUFFER_GAS_COST;
```
PlasmaPower, line 588: Gas deducted needs to be doubled here too.

PlasmaPower (changes requested):

PR #846, opened 2021-03-09
custom tokens / token bridge cleanup WIP

DZGoldman:
Marking WIP since this should probably wait to incorporate Ed's new replayable L2 transaction stuff before merging.

Includes remainder of contract & client side custom token logic, some bug fixes, contract code DRYing, & c.  

(should automate update:abi at this point — pre-commit hook?)


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -145,48 +143,67 @@ contract EthERC20Bridge is L1Buddy {
...
         address erc20,
         address destination,
         uint256 amount,
         uint256 maxGas,
         uint256 gasPriceBid
     ) external payable onlyIfConnected {
+        depositToken(erc20, destination, amount, maxGas, gasPriceBid, StandardTokenType.ERC20);
+    }
+
+    function depositAsCustomToken(
```
hkalodner, line 190: Looks like this is currently private. What the plan for calling it?

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/StandardTokenType.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
hkalodner, line 1: Apache 2.0 unless there's a reason to do otherwise

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
hkalodner: Also we probably want to restrict this to only working if a custom token has been deployed

hkalodner (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
DZGoldman: y and y

PR #864, opened 2021-03-11
Improve staker strategies and edge case handling

PlasmaPower:


packages/arb-bridge-eth/contracts/validator/Validator.sol
```diff
@@ -70,14 +70,26 @@ contract Validator {
     function returnOldDeposits(IRollup rollup, address payable[] calldata stakers) external {
         uint256 stakerCount = stakers.length;
         for (uint256 i = 0; i < stakerCount; i++) {
-            try rollup.returnOldDeposit(stakers[i]) {} catch {}
+            try rollup.returnOldDeposit(stakers[i]) {} catch (bytes memory error) {
+                if (error.length == 0) {
+                    // Assume out of gas
+                    // We need to revert here so gas estimation works
+                    require(false, "GAS");
+                }
```
hkalodner, line 78: Ah nice 😁

packages/arb-bridge-eth/contracts/validator/ValidatorUtils.sol
```diff
@@ -265,4 +265,14 @@ contract ValidatorUtils {
         }
         return (challenges, hasMoreStakers);
     }
+
+    function areUnresolvedNodesLinear(Rollup rollup) external view returns (bool) {
```
hkalodner, line 269: Might be good to add some documentation here that it's O(n) in the number of nodes, but safe since in the happy case where there are no forks we want assume that there won't be too many nodes and in the malicious case the function will terminate early

packages/arb-node-core/staker/staker.go
```diff
@@ -91,24 +91,61 @@ func (s *Staker) Act(ctx context.Context) (*types.Transaction, error) {
...
+			effectiveStrategy = StakeLatestStrategy
 		}
 	}
 
-	if info != nil {
+	// Resolve nodes if either we're on the make nodes strategy,
+	// or we're on the stake latest strategy but don't have a stake
+	// (attempt to reduce the current required stake).
+	shouldResolveNodes := s.strategy >= MakeNodesStrategy
+	if !shouldResolveNodes && s.strategy >= StakeLatestStrategy && info == nil {
```
hkalodner, line 111: I think maybe we want `effectiveStrategy` here?

hkalodner (approved):
LGTM

hkalodner (changes requested):
Whoops, previous approve was an accident

packages/arb-bridge-eth/contracts/validator/ValidatorUtils.sol
PlasmaPower: I think a better way to describe it is O(depth) but yeah I'll document that

hkalodner (approved):
LGTM

PR #865, opened 2021-03-11
Create ReadSnapshotTransaction

joshuacolvin0:
also cleanup the existing Read*Transaction objects

### DIFF
--- a/packages/arb-avm-cpp/app/main.cpp
+++ b/packages/arb-avm-cpp/app/main.cpp
@@ -119,7 +119,7 @@ int main(int argc, char* argv[]) {
               << assertion.gasCount << " gas ending in state "
               << static_cast<int>(mach->currentStatus()) << "\n";
 
-    auto tx = storage.makeReadWriteTransaction();
+    auto tx = storage.getReadWriteTransaction();
     saveMachine(*tx, *mach);
     tx->commit();
 
--- a/packages/arb-avm-cpp/app/main.cpp
+++ b/packages/arb-avm-cpp/app/main.cpp
@@ -119,7 +119,7 @@ int main(int argc, char* argv[]) {
               << assertion.gasCount << " gas ending in state "
               << static_cast<int>(mach->currentStatus()) << "\n";
 
-    auto tx = storage.getReadWriteTransaction();
+    auto tx = storage.makeReadWriteTransaction();
     saveMachine(*tx, *mach);
     tx->commit();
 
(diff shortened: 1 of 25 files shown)


packages/arb-avm-cpp/app/main.cpp
```diff
@@ -119,7 +119,7 @@ int main(int argc, char* argv[]) {
               << assertion.gasCount << " gas ending in state "
               << static_cast<int>(mach->currentStatus()) << "\n";
 
-    auto tx = storage.makeReadWriteTransaction();
+    auto tx = storage.getReadWriteTransaction();
```
hkalodner, line 122: What's the logic behind the name change? Seems like we're making a new one, not getting an existing one

packages/arb-avm-cpp/data_storage/include/data_storage/readtransaction.hpp
```diff
@@ -14,29 +14,25 @@
...
 
+class ReadSnapshotTransaction;
 class ReadWriteTransaction;
 
 class ReadTransaction {
    private:
     std::unique_ptr<Transaction> transaction{};
     rocksdb::ReadOptions read_options{};
 
+    friend ReadSnapshotTransaction;
```
hkalodner, line 30: Rather than making that class a friend, the variables it needs to access should just be `protected` I think

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -177,7 +177,8 @@ void saveNextSegmentID(ReadWriteTransaction& tx, uint64_t next_segment_id) {
     }
 }
 
-uint64_t getNextSegmentID(ReadTransaction& tx) {
+uint64_t getNextSegmentID(std::shared_ptr<DataStorage> store) {
+    ReadTransaction tx(store);
```
hkalodner, line 181: `std::move`

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/include/data_storage/readtransaction.hpp
joshuacolvin0: 👍

packages/arb-avm-cpp/data_storage/src/value/code.cpp
joshuacolvin0: 👍

packages/arb-avm-cpp/app/main.cpp
joshuacolvin0: 👍

hkalodner (approved):
LGTM

PR #870, opened 2021-03-12
Add permit() & transferAndCall() to token bridge ERC20

dmihal:
Adds ERC677 transferAndCall & ERC2612 permit to the bridged ERC20 token.


packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/StandardArbERC20.sol
```diff
@@ -43,6 +52,12 @@ contract StandardArbERC20 is OZERC20, Cloneable, IArbToken {
         _decimals = decimals_;
     }
 
+    function DOMAIN_SEPARATOR() external view returns (bytes32) {
+        uint256 chainId;
+        assembly {chainId := chainid()}
+        return chainId == deploymentChainId ? _DOMAIN_SEPARATOR : _calculateDomainSeparator(chainId);
```
fredlacs, line 58: Is there any reason for this comparison?
I don't believe the chainId ever changes.

dmihal: We used this comparison in WETH10, apparently the chainId _can_ change in the case of a DAO-style fork, and this can cause issues with some permits.

This was the most gas-efficient code we came up with to work around that issue. However, maybe chain forks isn't a concern on Arbitrum?

fredlacs: Ah interesting, thats a good point. Chain ids shouldn't be an issue for arbitrum in that case, as the chain id is a function of the Layer 1's rollup address

PR #874, opened 2021-03-12
various Arb ts updates

DZGoldman:


packages/arb-ts/src/lib/bridge.ts
```diff
@@ -78,4 +97,11 @@ export class Bridge extends L2Bridge {
       destinationAddress
     )
   }
+  public getAndUpdateL1TokenData(erc20l1Address: string) {
+    return this.l1Bridge.getAndUpdateL1TokenData(erc20l1Address)
+  }
+
+  public async getAndUpdateL1EthBalance() {
```
fredlacs, line 104: I believe this doesn't need to be async

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -213,4 +250,9 @@ export class L1Bridge {
     this.inboxCached = InboxFactory.connect(inboxAddress, this.l1Signer)
     return this.inboxCached
   }
+
+  public async getAndUpdateL1EthBalance(): Promise<BigNumber> {
+    const bal = await this.l1Signer.getBalance()
+    return (this.l1EthBalance = bal)
```
fredlacs, line 256: this doesn't seem very _idiomatic_? The L-value gets returned after being assigned?

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -214,6 +219,20 @@ export class L2Bridge {
...
+    try {
+      const arbERC20 = StandardArbERC20Factory.connect(
+        erc20L2Address,
+        this.l2Signer
+      )
+      return arbERC20.l1Address()
+    } catch (e) {
+      console.warn('Could not get L1 Address')
+
+      return
```
fredlacs, line 232: Do we want to throw an error or return undefined?

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -229,4 +248,9 @@ export class L2Bridge {
     this.walletAddressCache = await this.l2Signer.getAddress()
     return this.walletAddressCache
   }
+
+  public async getAndUpdateL2EthBalance(): Promise<BigNumber> {
+    const bal = await this.l2Signer.getBalance()
+    return (this.l2EthBalance = bal)
```
fredlacs, line 254: Same as previous

fredlacs (changes requested):

packages/arb-ts/src/lib/bridge.ts
DZGoldman: we want it async because we want to both save it to the state of bridge instance and return it; you're right that's un-idiomatic tho, will clean

packages/arb-ts/src/lib/l2Bridge.ts
DZGoldman: having a check that doesn't throw proved useful; the consumer of the API can do w/ this what they want (throw, chill, etc.)

DZGoldman:
re-based and merge-pilled (idk) @fredlacs

fredlacs (approved):

PR #878, opened 2021-03-12
deposit eth integration test (via arb-ts bridge)

DZGoldman:


packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -12,13 +15,55 @@ const preFundedWallet = new Wallet(preFundedSignerPK, ethProvider)
...
-  l2TestWallet
-)
+let arb20BridgeAddress: string
+
+let bridge: Bridge | undefined
+
+describe('setup bridge', () => {
+  it('instantiate bridge ', async () => {
+    const ethBridge = EthERC20BridgeFactory.connect(
+      erc20BridgeAddress,
```
fredlacs, line 25: I think this should be the inbox address, but you removed it from the config?

DZGoldman: you're right

DZGoldman:
closing for https://github.com/OffchainLabs/arbitrum/pull/895

PR #884, opened 2021-03-14
Snapshotting

hkalodner:


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1173,69 +1179,65 @@ rocksdb::Status ArbCore::advanceExecutionCursorImpl(
...
+                    std::cout << "Error getting messages for execution cursor"
+                              << std::endl;
+                    return get_messages_result.status;
+                }
+                if (!get_messages_result.data.first) {
+                    // Reorg occurred, need to recreate machine
+                    handle_reorg = true;
+                    if (!possible_reorg) {
+                        std::cerr << "Warning: Unexpected execution cursor "
+                                     "reorg detected"
```
PlasmaPower, line 1227: I don't think this is accurate now that there's no longer a consistent snapshot for the entire execution.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -2119,25 +2121,32 @@ ValueResult<std::unique_ptr<Machine>> ArbCore::getMachineForSideload(
...
-        std::get<MachineStateKeys>(closest_checkpoint));
+        gas_target = position_res.data;
+        execution_cursor = std::make_unique<ExecutionCursor>(
+            std::get<MachineStateKeys>(closest_checkpoint));
+    }
 
-    auto status = advanceExecutionCursorImpl(
-        tx, *execution_cursor, position_res.data, false, 10, cache, false);
+    auto status = advanceExecutionCursorImpl(*execution_cursor, gas_target,
+                                             false, 10, cache, false);
```
PlasmaPower, line 2147: Isn't the assumption here that we don't need to advance the execution cursor, and advancing it often wouldn't work because of the staged message?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1173,69 +1179,64 @@ rocksdb::Status ArbCore::advanceExecutionCursorImpl(
...
+                auto get_messages_result = executionCursorGetMessages(
+                    tx, execution_cursor, message_group_size);
+                if (!get_messages_result.status.ok()) {
+                    std::cout << "Error getting messages for execution cursor"
+                              << std::endl;
+                    return get_messages_result.status;
+                }
+                if (!get_messages_result.data.first) {
+                    // Reorg occurred, need to recreate machine
+                    handle_reorg = true;
```
joshuacolvin0, line 1224: I don't know if it would print too many logs, but at least for initial debugging it might be useful to have a debug message here to tell us the message/gas that it reorged at?

hkalodner: Agreed. Added a `cout`

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
hkalodner: I don't think we want to assume that we'll always have a checkpoint for every sideload. It's true currently, but we may want to adjust our strategy over time and this should work efficiently in the case where a checkpoint exists

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
hkalodner: Agreed, updated the PR

joshuacolvin0 (approved):
LGTM

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
PlasmaPower: Hmm... I guess the staged message restriction only applies to the latest sideload, so this works.

PR #891, opened 2021-03-15
Calculate L2 token address in L1

fredlacs:


DZGoldman (approved):

DZGoldman:
approved, this is now rebased onto https://github.com/OffchainLabs/arbitrum/pull/877, so we can just merge it in there

PR #896, opened 2021-03-17
Fix nonce error detection, add db cleanup on startup, properly close db on exit

joshuacolvin0:


packages/arb-checkpointer/checkpointing/indexedCheckpointer.go
```diff
@@ -212,7 +212,15 @@ func restoreLatestState(
 			continue
 		}
 		logger.Info().Msg("Restored state")
-		return nil
+		success := db.FlushCheckpointStorage()
+		if !success {
+			logger.Error().Msg("Failed flushing database")
+		}
+		logger.Info().Msg("Flushed database")
+		db.CloseCheckpointStorage()
```
hkalodner, line 220: If we're continuing to launch after this, don't we want to keep the db open?

hkalodner (changes requested):

packages/arb-checkpointer/checkpointing/indexedCheckpointer.go
joshuacolvin0: forgot to push second commit

hkalodner (approved):
LGTM

PR #905, opened 2021-03-17
Validator should continually retry

joshuacolvin0:
Adjust settings to retry more often on nonce errors, and ensure validator continuously retries when there is an error posting batches


hkalodner (approved):
LGTM

PR #907, opened 2021-03-18
Fix typechain factories

DZGoldman:
Verbose explanation of an issue that's not all that exciting incoming:

Typechain v5 puts typechain contract factories in a subdirectory and names them MyContract__factory instead of MyContractFactory to avoid name clashes (see https://github.com/ethereum-ts/TypeChain/issues/225 and . i.e., us, with NodeFactory.sol). So we should always use the __factory contract in `abi/factories` for actual typechain factories. 

The reason were seeing both, i.e., both `InboxFactory.ts` and `/factories/Inbox__factory.ts` was _not_ because we currently have mismatched typechain versions, but because we _previously_ did, and typechain doesn't clear out old, out-of-use files from `build/types` after a `hardhat compile` (which they probably should) (so, I, e.g. had a NodeFactory.ts that kept getting copied over into arb-ts and messing up imports).

The agro thing would be to rm -rf all `types` directories as part of the update:abi script, but I opt to just tell everyone "delete your `build/types` directories in arb-bridge-eth, arb-os, and arb-bridge-peripherals" once after this is merged.

I also added update:abi as a postinstall script in arb-ts (open to other suggestions), and changed update-abi to move every interface into arb-ts instead of just the ones we currently need because yolo.


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

packages/arb-ts/src/lib/bridge.ts
```diff
@@ -40,6 +38,8 @@ interface BuddyDeployEventResult {
   withdrawalId: BigNumber
   success: boolean
 }
+import { Bridge__factory } from './abi/factories/Bridge__Factory'
```
fredlacs, line 41: Do we want these on the top of the file?

hkalodner: Nice catch 😁

fredlacs (approved):
LGTM

PR #917, opened 2021-03-18
Bridge code review fixes

fredlacs:
This PR addresses the code review done on the token bridge.

It does not address https://github.com/OffchainLabs/arbitrum/pull/877#discussion_r596500488 as we have quite a backlog abigen cleanup to do. I suggest we address that in a separate PR. These changes also include a new hardcoded event from the buddy deployer, but can be cleaned up in the future with changes to the abi gen


hkalodner (approved):
LGTM

PR #923, opened 2021-03-18
Bridge post deposit call

fredlacs:


packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -86,35 +87,94 @@ contract ArbTokenBridge is CloneFactory {
...
         address account,
         uint256 amount,
-        uint8 decimals
+        uint8 decimals,
+        bytes calldata postMintCall
     ) external onlyEthPair {
         IArbToken token = ensureERC20TokenExists(l1ERC20, decimals);
         token.bridgeMint(account, amount);
+
+        (bool success, address backupAddr) = executePostMintCall(postMintCall);
```
dmihal, line 161: This opens up a potential attack.

Imagine if the user deposits a token, sets the `account` as the address of another bridged token contract, and sets the `postMintCall` to the data to mint infinite tokens to the attacker.

dmihal (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -86,35 +93,70 @@ contract ArbTokenBridge is CloneFactory {
...
+        address sender,
+        address dest,
         uint256 amount,
-        uint8 decimals
+        uint8 decimals,
+        bytes calldata callHookData
     ) external onlyEthPair {
         IArbToken token = ensureERC777TokenExists(l1ERC20, decimals);
-        token.bridgeMint(account, amount);
+        token.bridgeMint(dest, amount, callHookData);
```
hkalodner, line 121: I think we need a try-catch and mint to sender for this probably

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -86,35 +93,70 @@ contract ArbTokenBridge is CloneFactory {
...
-        uint256 amount
+        uint256 amount,
+        bytes calldata callHookData
     ) external onlyEthPair {
         address tokenAddress = customToken[l1ERC20];
         require(tokenAddress != address(0), "Custom Token doesn't exist");
         IArbToken token = IArbToken(tokenAddress);
-        token.bridgeMint(account, amount);
+        token.bridgeMint(account, amount, callHookData);
+        // TODO: should we add postMintCall? custom token logic could implement it in bridgeMint
```
hkalodner, line 159: It can't implement it without knowing the sender though

hkalodner (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: they can encode it in the callHookData, just not as efficient

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: If the sender is encoded in the callhookdata then they can achieve the same functionality.

I'm a bit hesitant of adding a post-call hook on something that triggers a post transfer hook. It should be alright, but I'm hesitant.

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
hkalodner: That doesn't work if the sender is an EOA right? To rely on that it would have to come from an L1 trusted contract or something like that

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -41,6 +46,8 @@ contract ArbTokenBridge is CloneFactory {
     ICloneable public immutable templateERC777;
     address public immutable l1Pair;
 
+    event MintAndCallTriggered(bool success);
```
hkalodner, line 48: We might want other info in this event?

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -86,35 +93,99 @@ contract ArbTokenBridge is CloneFactory {
...
+            } catch {
+                // if reverted, then credit sender's account
+                // TODO: should try to submit callHookData?
+                token.bridgeMint(sender, amount, '');
+                emit MintAndCallTriggered(false);
+            }
+        } else {
+            // if no hook data, credit the dest
+            token.bridgeMint(dest, amount, '');
+        }
```
hkalodner, line 187: Could these common blocks be refactored?

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: Added in

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: Fair, added in the try catch

hkalodner (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: We can

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
fredlacs: On it

fredlacs: Do we want to also emit the data used in the call? I added in sender, dest and amount.

hkalodner (approved):
LGTM

PR #932, opened 2021-03-20
Add support for retryable receipts

hkalodner:


packages/arb-evm/evm/incoming.go
```diff
@@ -41,7 +41,7 @@ func FilterEthTxResults(results []*TxResult) []*ProcessedTx {
 	for _, res := range results {
 		kind := res.IncomingRequest.Kind
 		// Ignore other message types
-		if kind != message.L2Type {
+		if kind != message.L2Type && kind != message.RetryableType {
 			continue
 		}
 		processed, err := GetTransaction(res)
```
PlasmaPower, line 47: This'll fail if it isn't RetryableType (maybe the check there just needs extended too)

PlasmaPower (changes requested):

PlasmaPower (approved):
Ignore my previous comment, LGTM

PR #934, opened 2021-03-21
Token bridge updates

hkalodner:


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -190,8 +192,14 @@ contract EthERC20Bridge is L1Buddy {
         uint256 maxGas,
         uint256 gasPriceBid
     ) external payable onlyIfConnected returns (uint256) {
-        string memory name = SafeERC20Namer.tokenName(erc20);
-        string memory symbol = SafeERC20Namer.tokenSymbol(erc20);
+        string memory name = "";
+        string memory symbol = "";
+        try ERC20(erc20).name() returns (string memory _name) {
```
fredlacs, line 197: Should we use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/IERC20Metadata.sol

This way we keep interfaces all around for the calls. Also ERC20 is a name-clash with the StandardTokenType enum.

fredlacs: Ah seems like it's still unreleased https://github.com/OpenZeppelin/openzeppelin-contracts/blob/74e03de7604ca99c2f8e476476a6dc362779c356/CHANGELOG.md#unreleased

fredlacs (approved):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -190,8 +192,14 @@ contract EthERC20Bridge is L1Buddy {
...
+            name = _name;
```
fredlacs, line 198: This return type isn't always a string, it can be bytes32 such as for the MKR token https://etherscan.io/address/0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2#readContract

I think the cleanest approach is to staticcall the `name()` function and check the return data length. If its padded to 32 bytes we decode it to a string.

fredlacs (commented):
In https://github.com/OffchainLabs/arbitrum/pull/937 the parsing of the fields is pushed to the L2. It also makes safer staticcalls to `name`, `symbol` and `decimals` which aren't always available in the standard format.

PR #935, opened 2021-03-22
Aggregator healthcheck

baburgess:
Add a basic health check into the RPC node


hkalodner (changes requested):

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -0,0 +1,269 @@
...
+type Log struct {
+	Err    error
+	Sev    string
+	Var    string
+	Comp   string
+	Debug  bool
+	Config bool
+
+	ValStr    string
+	ValBigInt big.Int
```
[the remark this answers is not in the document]
baburgess: The edge case I was worried about was if the channel was buffering and the variable was rewritten after the Log struct was pushed to the healthcheck.  This would cause unexpected behavior I think where the newly written value would be logged versus the intended value which is why I was passing by value.

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -124,16 +134,19 @@ func (config *configStruct) loadConfig() {
 	config.primaryHealthcheckRPC = ""
 	config.successCode = defaultSuccessCode
 	config.blockDifferenceTolerance = defaultBlockDifferenceTolerance
+
+	return &config
 }
 
 func checkEndpoint(config *configStruct, endpoint string) healthcheck.Check {
```
hkalodner, line 141: Strings are copied by value so as is this won't change if the original config variable changes. If we want that, we need endpoint to be a pointer or to read from configStruct directly

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -151,38 +164,24 @@ func checkInboxReader(config *configStruct, state *healthState) healthcheck.Chec
 	check := healthcheck.Async(func() error {
 		//Lock config mutex for read operation
 		state.mu.Lock()
+		defer state.mu.Unlock()
 		//Calculate out the block difference
 		const subtractionBigInt = 0
-		blockDifference := big.NewInt(subtractionBigInt).Sub(&state.inboxReader.caughtUpTarget, &state.inboxReader.currentHeight)
+		blockDifference := big.NewInt(subtractionBigInt).Sub(state.inboxReader.caughtUpTarget, state.inboxReader.currentHeight)
```
hkalodner, line 170: Cleaner way to do this is `new(big.Int)`

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -151,38 +164,24 @@ func checkInboxReader(config *configStruct, state *healthState) healthcheck.Chec
...
 		//Set the tolerance we are willing to accept
 		tolerance := big.NewInt(config.blockDifferenceTolerance)
 		//Compare the tolerance using CmpAbs, fail if > then tolerance
 		const greaterThanReturn = 1
 		if blockDifference.CmpAbs(tolerance) == greaterThanReturn {
```
hkalodner, line 175: Our standard way of doing this is just directly `> 0` instead of using a var and `== greaterThanReturn`

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -200,38 +199,18 @@ func nodeReadinessChecks(health healthcheck.Handler, httpMux *http.ServeMux, aSy
...
-func healthcheckServer(httpMux *http.ServeMux, config configStruct) {
-	http.ListenAndServe(config.healthcheckRPC, httpMux)
-}
-
 func waitConfig(config *configStruct) {
 	config.mu.Lock()
+	defer config.mu.Unlock()
 	for config.openethereumHealthcheckRPC == "" {
-		config.mu.Unlock()
 		time.Sleep(config.loopDelayTimer)
```
hkalodner, line 206: Probably want to avoid holding the lock over the sleep. In general I'd say using the `Lock` and defer `Unlock` is better unless there's something slow in the function in which case making it more fine-grain is better

packages/arb-rpc-node/cmd/arb-node/arb-node.go
```diff
@@ -70,6 +70,13 @@ func main() {
...
 	const largeChannelBuffer = 200
 	healthChan := make(chan nodehealth.Log, largeChannelBuffer)
+
+	go func() {
+		err := nodehealth.NodeHealthCheck(healthChan)
+		if err != nil {
+			log.Error().Err(err).Msg("healthcheck server failed")
+		}
+	}()
 	go nodehealth.NodeHealthCheck(healthChan)
```
hkalodner, line 80: This should probably be removed

hkalodner (changes requested):

packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -0,0 +1,269 @@
...
+	config.successCode = defaultSuccessCode
+	config.blockDifferenceTolerance = defaultBlockDifferenceTolerance
+}
+
+func checkEndpoint(config *configStruct, endpoint string) healthcheck.Check {
+	check := healthcheck.Async(func() error {
+		//Lock config mutex for read operation
+		config.mu.Lock()
+		//Retrieve status code from healthcheck endpoint
+		res, err := http.Get(endpoint + "/ready")
```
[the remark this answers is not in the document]
baburgess: I changed endpoint to being passed through a pointer.  The lock should be required since the channel listener could write to it during a read which would cause an error I thought?

hkalodner (changes requested):
One tiny comment, but looks good to me otherwise. Pinging @joshuacolvin0 to take a look as well

joshuacolvin0 (approved):
LGTM

hkalodner (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

PR #937, opened 2021-03-22
Add safer erc20 information handling

fredlacs:
Tokens such as MKR have `name` and `symbol` encoded as `bytes32`. Not all tokens implement `decimals`, with this approach  we handle the parsing logic on the L2


hkalodner (approved):
LGTM

PR #944, opened 2021-03-23
Cleanup old checkpoints

joshuacolvin0:


hkalodner (changes requested):

packages/arb-avm-cpp/cmachine/arbcore.go
```diff
@@ -67,6 +67,16 @@ func (ac *ArbCore) MachineIdle() bool {
 	return status == 1
 }
 
+func (ac *ArbCore) CheckpointMinMessageIndex(messageIndex *big.Int) error {
```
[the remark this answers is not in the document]
joshuacolvin0: 👍

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -888,11 +891,26 @@ void ArbCore::operator()() {
...
+
+        if (!machineIdle() && checkpoints_min_message_index > 0) {
+            auto status = deleteOldCheckpoint(checkpoints_min_message_index);
+            if (!status.ok()) {
+                if (!status.IsNotFound()) {
+                    core_error_string = "error deleting old checkpoints";
+                    machine_error = true;
+                    std::cerr << "ArbCore error: " << core_error_string << ": "
+                              << status.ToString() << "\n";
+                    break;
```
PlasmaPower, line 908: Do we really need to stop ArbCore here? I think it's perfectly safe to log an error and keep going.

joshuacolvin0: good point.

PR #947, opened 2021-03-23
include hxhash data in getTokenWithdrawEventData; add missing return

DZGoldman:


packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -184,6 +187,7 @@ export class L2Bridge {
           contract: arbERC20TokenContract,
           balance,
         }
+        return tokenData
```
hkalodner, line 190: Don't we want to keep going so we can also potentially fill in ERC-777 data?

hkalodner (approved):
LGTM

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -223,10 +223,13 @@ export class L2Bridge {
       )
       const balance = await arbERC777TokenContract.balanceOf(walletAddress)
       tokenData.ERC777.balance = balance
+    }
+    if (tokenData.ERC20 || tokenData.ERC777 || tokenData.CUSTOM) {
       return tokenData
```
DZGoldman, line 228: this gets all token types and preserves the return semantics we want

packages/arb-ts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "arb-ts",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
fredlacs, line 3: Is it `v0.0.4` now?

PR #948, opened 2021-03-23
remove update:abi from arb-ts post-install

DZGoldman:
triggers a build fail, + we clearly need to figure out a better solution anyway, so removing it from post-install for now


hkalodner (approved):
LGTM

PR #953, opened 2021-03-24
Speed up validator syncing

PlasmaPower:


packages/arb-util/core/execution.go
```diff
@@ -86,16 +95,23 @@ func (e *ExecutionTracker) fillInCursors(max int) error {
...
+		return nil, errors.New("invalid gas used")
 	}
 	if err := e.fillInCursors(index); err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	cursor := e.cursors[index]
+	return e.cursors[index], nil
+}
```
hkalodner, line 108: Without a clone this is pretty unsafe since it's a pointer to a cursor that's used internally

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #955, opened 2021-03-25
Delay the validator after making a transaction

PlasmaPower:
With only a second of delay, the node often wouldn't see the results of its own transaction before its next action.


hkalodner (approved):
LGTM

PR #956, opened 2021-03-25
Inside arbitrum suggestions (WIP)

DZGoldman:
mostly just little clarifications / phrasing tweaks so far, take it or leave it


DZGoldman:
- In Rollup Protocol.md, I think the list "Rules for Confirming or Rejecting Rollup Blocks" my give the impression that stakers have to be more active than they really do (i.e., everyone has to keep chasing the latest block). Might need to look w/  fresh eyes.
- Also in Rollup_Protocol.md, the example of blocks 93 and 95 left me wondering about race conditions (i.e., Alice and Bob both honestly producing a different block 93 on top of 92.)

edfelten (approved):
These are all great.  Thanks!

PR #970, opened 2021-03-26
Inside arbitrum suggestions, 2nd half

DZGoldman:
@edfelten not sure how /if it's possible to post comments on parts that aren't in the diff, so just going to include some more notes:

In AVM.md
- `The use of a Layer 2 trusted operating system does require some support in the architecture, for example to allow the OS to limit and track resource usage by contracts.`
Can you clarify “support in the architecture.”? Does this mean “some additional support by the L1 contracts”?
- `Any Layer 2 protocol that relies on assertions and dispute resolution (which includes at least all rollup protocols) must define a rule for Merkle-hashing the full state of the virtual machine so that claims about parts of the state can be efficiently made to the base layer.`
Nitpick: "all rollup protocols" is a bit too strong; i.e., Fuel gets away with no state serialization at all. 
- `Code is added to an AVM in two ways. First, some code is created when the AVM starts running. This code is read in from an AVM executable file (a .mexe file) and preloaded by the AVM emulator.`
Is it accurate to say that the code created when the AVM starts is the ArbOS system-wide stuff, and the rest are user-deployed contracts deployed during runtime? And if so, maybe clarify? 

Arbgas_and_Fees.md
 - `It follows that if the parent's ArbGas claim is false, at least one of the sub-claims must make a false ArbGas claim. So a challenger who knows that the parent's ArbGas claim is false will always be able to find a claim that has a false ArbGas claim.`
 "Parent" threw me a little here; I read it as parent block, not "parent bisected claim in the dispute." not sure how best to clarify


docs/Inside_Arbitrum/ArbGas_and_Fees.md
```diff
@@ -59,9 +59,9 @@ Fees are charged for four resources that a transaction can use:
 * *L2 tx*: a base fee for each L2 transaction, to cover the cost of servicing a transaction
 * *L1 calldata*: a fee per units of L1 calldata directly attributable to the transaction (each non-zero byte of calldata is 16 units, and each zero byte of calldata is 4 units)
 * *computation*: a fee per unit of ArbGas used by the transaction
-* *storage*: a fee per location of EVM storage, based on the net increase in EVM storage due to the transaction
+* *storage*: a fee per location of AVM storage, based on the net increase in EVM storage due to the transaction
```
DZGoldman, line 62: I think we should be consistent with "L2 EVM storage" or "AVM storage"

docs/Inside_Arbitrum/ArbGas_and_Fees.md
```diff
@@ -84,7 +84,7 @@ If these conditions are not all met, the transaction is treated as not submitted
 
 ### Price for storage
 
-Transactions are charged based on the net increase they cause in the total amount of L2 EVM contract storage that exists. Decreases in contract storage do not receive a credit. Each storage location costs 2000 times the estimated L1 gas price. This means that storage costs about 10% as much as it does on the L1 chain.
+Transactions are charged based on the net increase they cause in the total amount of AVM contract storage that exists. Decreases in contract storage do not receive a credit. Each storage location costs 2000 times the estimated L1 gas price. This means that storage costs about 10% as much as it does on the L1 chain.
```
DZGoldman, line 87: https://github.com/OffchainLabs/arbitrum/pull/970/files#r602470691

docs/Inside_Arbitrum/ArbGas_and_Fees.md
edfelten: This is meant to refer to storage of EVM contracts that are running on top of Arbitrum/ArbOS.  So EVM storage is closer to correct.

docs/Inside_Arbitrum/ArbGas_and_Fees.md
edfelten, line 87: Sam as above. This should be EVM storage.

docs/Inside_Arbitrum/AVM.md
```diff
@@ -3,9 +3,9 @@ id: AVM
...
+The Arbitrum Virtual Machine (AVM) is the interface between the Layer 1 and Layer 2 parts of Arbitrum. Layer 1 *provides* the AVM interface and ensures correct execution of the virtual machine. Layer 2 *runs on* the AVM virtual machine and provides the functionality to deploy and run contracts, track balances, and all of the things a smart-contract-enabled blockchain needs to do.
 
-**Every Arbitrum chain has a single AVM** which does all of the computation and maintains all of the storage for everything that happens on the chain. Unlike some other systems which have a separate “VM” for each contract, Arbitrum uses a single virtual machine for the whole chain. The management of multiple contracts on an Arbitrum chain is done by software that runs on top of the AVM. 
+**Every Arbitrum chain has a single AVM** which does all of the computation and maintains all of the storage for everything that happens on the chain. Unlike some other systems which have a separate “VM” for each contract, Arbitrum uses a single virtual machine for the whole chain, much like Ethereum. The management of multiple contracts on an Arbitrum chain is done by software that runs on top of the AVM.
```
edfelten, line 8: I prefer to drop the "much like Ethereum" because Ethereum blurs the line between the VM and the surrounding system, whereas we draw that line more cleanly.

edfelten (changes requested):
A few suggested changes.  Happy to discuss any of these.

PR #973, opened 2021-03-26
Fix issue with empty logscursor after reorg

joshuacolvin0:
When a reorg happens and logscursor had waiting logs before reorg, but no new logs or deleted logs after reorg, should be switched back to REQUESTED so that error state is not entered.


hkalodner (approved):
LGTM

PR #986, opened 2021-03-30
Disable rocksdb bloom filter

joshuacolvin0:
Disable bloom filter until rocksdb is upgraded to 6.12+ when optimize_filters_for_memory is available to reduce memory usage.


hkalodner (approved):
LGTM

PR #987, opened 2021-03-30
Merge next into new-rollup

hkalodner:


packages/arb-rpc-node/batcher/batcher.go
```diff
@@ -241,10 +241,13 @@ func (m *Batcher) sendBatch(ctx context.Context, inbox l2TxSender) {
 	for _, tx := range txes {
 		batchTxes = append(batchTxes, message.NewCompressedECDSAFromEth(tx))
 	}
+	var err error
```
joshuacolvin0, line 244: not needed?

joshuacolvin0 (approved):
LGTM

packages/arb-rpc-node/batcher/batcher.go
hkalodner: Removed

PR #988, opened 2021-03-30
Update zerolog and cleanup logging

hkalodner:


packages/arb-evm/evm/log.go
```diff
@@ -28,7 +28,7 @@ import (
 	"strings"
 )
 
-var logger = log.With().Caller().Str("component", "evm").Logger()
+var logger = log.With().Stack().Caller().Stack().Str("component", "evm").Logger()
```
joshuacolvin0, line 31: Don't need two `Stack()`s?

joshuacolvin0 (approved):
LGTM

PlasmaPower:
LGTM minus Josh's comment

packages/arb-evm/evm/log.go
hkalodner: Fixed

PR #995, opened 2021-03-31
Custom tokens: e2e tests, interfaces, arb-ts tooling, etc

DZGoldman:
ICustomToken.sol  and IArbCustomToken.sol are supposed to serve as the "minimal-assumed-interfaces" for both.


packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -210,20 +215,37 @@ export class BridgeHelper {
 
   static getUpdateTokenInfoEventResult = async (
     l1Transaction: providers.TransactionReceipt,
```
DZGoldman, line 217: these event functions are a bit silly and should obviously by dryed up; with ts-generics magic we can preserve the dynamic returned interfaces I think. But I'd rather do that in a subsequent PR

fredlacs (commented):
Should we move the custom token test contracts from `packages/arb-bridge-peripherals/contracts/buddybridge/test` to `packages/arb-bridge-peripherals/contracts/tokenbridge/test`?

packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -385,14 +386,172 @@ describe('ERC20', () => {
...
-  //   console.log(l1EventData.length, withdrawTokenData.length)
-  //   expect(l1EventData.length).to.equal(withdrawTokenData.length)
-  // })
+})
+
+describe('CustomToken', () => {
+  let l1CustomToken: TestCustomTokenL1
+  let l2CustomToken: TestArbCustomToken
+  if (!existantCustomTokenL1 && !existantCustomTokenL2) {
+    it('sets up a new custom token, L1 and L2', async () => {
```
fredlacs, line 395: This looks like a test setup, should we move it to a `before` hook? Then we can evaluate the `if/else` inside the before hook instead of using it to determine which test cases we run.

packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -385,14 +386,172 @@ describe('ERC20', () => {
...
+      existantCustomTokenL2,
+      l2TestWallet
+    )
+    prettyLog(
+      `Connected to pre-deployed custom token addresses. L1:${l1CustomToken.address} L2: ${l2CustomToken.address}`
+    )
+  }
+
+  it('setup: mint some custom token and send to test address', async () => {
+    const mintRes = await l1CustomToken.mint()
```
fredlacs, line 471: Might seem a bit useless since its only for testing, but for the future it might be useful to inject the amount to mint as a parameter. Then we don't need to hardcode the `40000000` a couple of lines below

packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -385,14 +386,172 @@ describe('ERC20', () => {
...
+        l1CustomToken.address
+      )
+      expect(l2AddressHopefully).to.equal(l2CustomToken.address)
+    })
+  } else {
+    l1CustomToken = TestCustomTokenL1__factory.connect(
+      existantCustomTokenL1,
+      preFundedWallet
+    )
+    l2CustomToken = TestArbCustomToken__factory.connect(
```
fredlacs, line 461: We should probably check in the L1 Eth bridge that the `l1CustomToken` registered this address as its logic in the L2

packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -385,14 +386,172 @@ describe('ERC20', () => {
...
+    )
+
+    const retriableReceipt = await arbProvider.waitForTransaction(
+      l2RetriableHash
+    )
+
+    expect(retriableReceipt.status).to.equal(1)
+  })
+
+  it('wallet has expected balance after custom token deposit', async () => {
```
fredlacs, line 527: Should this not be part of the previous `deposit` test case?

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -192,16 +199,14 @@ export class BridgeHelper {
 
   static getDepositTokenEventData = async (
     l1Transaction: providers.TransactionReceipt,
-    tokenType: 'ERC20' | 'ERC777' = 'ERC20',
-    l2BridgeAddress?: string
+    l2BridgeAddress: string
```
fredlacs, line 202: Is there a reason to not make the address optional here? We only need the bridge interface since we pass in the receipt. I did it this way because we can just hardcode a random value to parse logs from the receipt using the interface

fredlacs: The same applies to the other methods that parse logs. There probably is a neater way to get the interface using typechain, but I wasn't able to figure it out

fredlacs: We can deal with this in a subsequent PR, as noted in https://github.com/OffchainLabs/arbitrum/pull/995/files#r605180872

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -36,7 +36,7 @@ const ARB_RETRYABLE_TX_ADDRESS = '0x000000000000000000000000000000000000006E'
 export interface L2TokenData {
   ERC20?: { contract: StandardArbERC20; balance: BigNumber }
   ERC777?: { contract: StandardArbERC777; balance: BigNumber }
-  CUSTOM?: { contract: IArbToken; balance: BigNumber } // Force custom to have l1Address (in, ie., IArbToken)
+  CUSTOM?: { contract: StandardArbERC20; balance: BigNumber } // Here we don't use the particlar custom token's interface; for the sake of this sdk that's fine
```
fredlacs, line 39: I don't think its fine - this can lead to us making silly mistakes in the future. We should probably use the typechain class generated from the custom token interface - we can make it an abstract contract instead of interface.

fredlacs (changes requested):

fredlacs:
While looking over the PR I noticed the `ensureTokenExists` codepath in the L2 ArbTokenBridge could use some refactoring. Should we also address it in this PR or leave it for a future one?

packages/arb-ts/integration_test/arb-bridge.test.ts
DZGoldman: no strong reason either way; i lean toward more, smaller cases if possible for more granular error logging

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: subsequent PRs 👍

DZGoldman:
(e2e tests pass)

fredlacs (approved):
LGTM

PR #1000, opened 2021-04-02
deprecate arb-provider-ethers

DZGoldman:
still need to update frontend_integration.md, yes; will be part of a bigger doc-upgrade


fredlacs (approved):
LGTM

PR #1001, opened 2021-04-02
Enable local Arbitrum on MacOS

jliphard:
Installation of Arbitrum currently fails on a fresh MacOS install. This PR updates slightly outdated documentation (`Local_blockchain.md`) and fixes some tiny little things that new Arbitrum users might encounter when locally deploying. With these changes, v0.8.0 should work right out of the box on MacOS.


hkalodner (approved):
This is awesome, thanks to much for the contribution!

PR #1004, opened 2021-04-06
Track nonce locally

hkalodner:
Track nonce locally instead of relying on remote node to report the correct nonce


joshuacolvin0 (approved):
LGTM

PR #1012, opened 2021-04-08
WIP: integration test for custom token fallback case

DZGoldman:
The deposit is failing; I may be consuming it wrong / misunderstanding the flow. I'm wondering if there's an issue with the fact that `hasTriedDeploy` doesn't get updated on the L1 side after a TMT gets deployed on the L2 side?


fredlacs (approved):
LGTM.

The test previously failed since the typechain interface didn't have the latest contract bytecode. I pushed the fix.

PR #1017, opened 2021-04-09
Improve bridge withdrawal and deposit

fredlacs:


fredlacs:
This PR still needs to update the typechain interfaces to include the latest bytecode.

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/ArbTokenBridge.sol
```diff
@@ -176,33 +183,24 @@ contract ArbTokenBridge is ProxySetter, IArbTokenBridge {
...
-            customToken[l1ERC20] != address(0)
-                ? customToken[l1ERC20]
-                : _calculateL2TokenAddress(l1ERC20);
+        address expectedSender = calculateL2TokenAddress(l1ERC20);
 
-        require(msg.sender == expectedSender, "Withdraw can only be triggered by expected sender");
+        // users can withdraw if its a standard erc20 token deployed by the bridge
+        // TODO: what happens if this was a rebasing stablecoin and user was supposed to hold less at time of withdrawal?
+        require(
+            msg.sender == expectedSender || msg.sender == calculateL2ERC20TokenAddress(l1ERC20),
```
DZGoldman, line 201: can you explain why checking both is required here?

fredlacs: Lets say a L1 token registers a custom implementation in the L2 but there is no contract in that L2 address
If a user tries depositing, the L2 will create a Temporary Migrateable Token.

If the L2 custom contract doesn't get deployed, the expectedSender according to the `customToken` mapping will be the custom implementation. Here we also allow withdrawals from the Temporary Migrateable Token.

I considered returning the TMT address in `calculateL2TokenAddress(l1ERC20)` but I think it might be simpler to have the L1/L2 address oracles be aligned, and handle the TMT edge case explicitly whenever it shows up. I'm not convinced this is the best approach tho

DZGoldman (approved):
lgtm

PR #1018, opened 2021-04-09
Add extra safety checks to token bridge

fredlacs:


fredlacs:
There are interface breaking changes in this PR.
Typechain interface still needs to be updated.

PR #1020, opened 2021-04-10
WIP: Initial work in broadcaster

joshuacolvin0:
based on gobwas websocket chat example


PlasmaPower:
I think we should rip out (most of) `clients.go`. The client shouldn't really need to make any requests, other than maybe supporting requesting messages since X sequence number, which we could put in the URL. I am surprised to see how complex the server is though.

joshuacolvin0:
> I think we should rip out (most of) `clients.go`. The client shouldn't really need to make any requests, other than maybe supporting requesting messages since X sequence number, which we could put in the URL. I am surprised to see how complex the server is though.

Totally, this code is just from the example chat server.  Leaving that part in was convenient for initial testing. 
1. Add channel so that sequencer can easily send new messages to feed
2. Remove client's ability to send commands, add PING/PONG support if needed
3. Convert from test (JSON) to binary

PR #1021, opened 2021-04-11
RPC and validator healthcheck update 

baburgess:


packages/arb-node-core/nodehealth/nodehealth.go
```diff
@@ -96,10 +99,24 @@ func updateConfig(config *configStruct, logMessage Log) {
 	defer config.mu.Unlock()
 	//Load the configuration message into the config struct
 	if logMessage.Var == "openethereumHealthcheckRPC" {
-		config.openethereumHealthcheckRPC = logMessage.ValStr
+		u, err := url.Parse(logMessage.ValStr)
+		if err != nil {
+			return
+		}
+		config.openethereumHealthcheckRPC = u.Hostname()
```
hkalodner, line 106: Not sure why we're stripping out the port here and delivering the hostname and port separately?

hkalodner (changes requested):

packages/arb-node-core/nodehealth/nodehealth.go
baburgess: The healthcheck is on a separate port from the RPC endpoint that is passed in via a command line argument.  This approach gives us configurability for the default healthcheck port inside the nodehealth file versus having to update it across the code base.

hkalodner: Ah cool

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #1029, opened 2021-04-13
update FrontendIntegration.md

DZGoldman:
Should we merge this directly on to `master` so the public page updates? People are poking around for the hackathon


DZGoldman:
I support building the docs from `develop` as per standup discussion @hkalodner

fredlacs (approved):

PR #1031, opened 2021-04-13
Token bridge natspec documentation 

DZGoldman:


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/EthERC20Bridge.sol
```diff
@@ -60,12 +65,14 @@ contract EthERC20Bridge is IEthERC20Bridge, TokenAddressHandler {
         _;
     }
 
+    /**
+     * @notice Initialize L1 bridge
+     * @param _inbox Address of Arbitrum chain's L1 Inbox.sol contract
+     * @param _l2TemplateERC20 Address of template ERC20 (i.e, StandardArbERC20.sol). Used for salt in computing L2 address.
+     * @param _l2ArbTokenBridgeAddress Address of L2 side of token bridge (ArbTokenBridge.sol)
+     */
     function initialize(
```
DZGoldman, line 74: this is the only code I change as the vars are unused; if it's too annoying at the moment to coordinate w/ others using the bridge deploy scripts etc. we can revert

fredlacs: I think its fine, but then we should also update the deploy scripts in this PR

fredlacs (approved):
LGTM

PR #1043, opened 2021-04-16
Add notice about L2 to L1 ticket not being consumed in reverts

fredlacs:


edfelten (approved):
Thanks for adding this clarification.

PR #1051, opened 2021-04-21
Code cache

hkalodner:
- Merge https://github.com/OffchainLabs/arb-os/pull/425
- Adds general testing to all of the arbos test suite that the amount of gas used while processing a transaction is within a bounded limit of the amount of l2 computation charged to the user
- Add specific testing for code caching
- Output debugprints as test logs instead of regular logs so that they're nicely interspersed with other test logs

### DIFF
--- a/packages/arb-rpc-node/arbostest/helper_test.go
+++ b/packages/arb-rpc-node/arbostest/helper_test.go
@@ -38,6 +38,8 @@ import (
 	"github.com/offchainlabs/arbitrum/packages/arb-util/value"
 )
 
+const printArbOSLog = false
+
 func initMsg(t *testing.T, options []message.ChainConfigOption) message.Init {
 	params := protocol.ChainParams{
 		StakeRequirement:          big.NewInt(0),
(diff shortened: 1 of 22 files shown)


packages/arb-rpc-node/arbostest/helper_test.go
```diff
@@ -196,29 +196,53 @@ func failIfError(t *testing.T, err error) {
...
 	t.Helper()
-	logs, sends, snap, assertion := runAssertionWithoutPrint(t, inboxMessages, logCount, sendCount)
-	testCase, err := inbox.TestVectorJSON(inboxMessages, assertion.Logs, assertion.Sends)
-	failIfError(t, err)
-	t.Log(string(testCase))
-	return logs, sends, snap, assertion
+	logs, sends, snap := runAssertionWithoutPrint(t, inboxMessages, logCount, sendCount)
+	//testCase, err := inbox.TestVectorJSON(inboxMessages, logs, sends)
+	//failIfError(t, err)
+	//t.Log(string(testCase))
```
joshuacolvin0, line 204: Should these lines just be deleted?

joshuacolvin0 (dismissed):
LGTM

packages/arb-rpc-node/arbostest/helper_test.go
hkalodner: I actually want to leave them in in case we want to re-enable them. Basically that produces output that can be used to recreate the test in an external environment (the aolog format). That could be used for example to run the test directly against the bare-metal AVM, but it adds a bunch of annoying log output even when we don't need it which is why I commented it out

joshuacolvin0 (approved):

joshuacolvin0 (approved):
LGTM

PR #1052, opened 2021-04-21
Fix bug in passing config to healthcheck

baburgess:
Pass configuration to healthcheck after the flag parse


joshuacolvin0 (approved):
LGTM

PR #1053, opened 2021-04-22
arb-ts autogened typedocs & some cleanup

DZGoldman:


packages/arb-ts/src/lib/bridge.ts
```diff
@@ -32,6 +35,14 @@ export class Bridge extends L2Bridge {
     ethSigner: Signer,
     arbSigner: Signer
   ) {
+    Promise.all([ethSigner.getAddress(), arbSigner.getAddress()]).then(
+      ([ethSignerAddress, arbSignerAddress]) => {
+        if (ethSignerAddress !== arbSignerAddress) {
+          throw new Error('L1 & L2 wallets must be of the same address')
```
fredlacs, line 41: Does the library not work if we break this assumption?
I think everything still works with different signers.

fredlacs (approved):
LGTM

PR #1054, opened 2021-04-23
Solidity docgen

DZGoldman:
This dumps md files for all contracts into docs/sol_contract_docs/md_docs. I think keeping them un-versioned here and having the developer_website repo build and consume them makes sense


fredlacs (approved):
LGTM

PR #1056, opened 2021-04-23
Compact all family columns on startup

joshuacolvin0:
Compacting rocksdb database improves performance and reduces number of files.


hkalodner (approved):
LGTM

PR #1058, opened 2021-04-25
Improve error handling

hkalodner:
Significantly reduce the usage of fatal and panic


PlasmaPower:
I think this is a good change, but for the purposes of ensuring a clean shutdown we could also just catch panics and shut down cleanly from there.

hkalodner:
Good point, the combination would probably be wise

hkalodner:
Actually I've reviewed this more and I don't believe that we need to explicitly catch panics. `panic` has a stack unwinding behavior in go like exceptions in c++ where the defer blocks are executed as the stack unwinds. The main executables do their cleanup via defers, so these will be handled cleanly.

The more problematic code is where an `os.exit` is triggered which is unrecoverable and leads to immediate exit. These has been removed from the codebase (`log.Fatal` uses this)

The other change here is capturing SIGINT and SIGTERM signals in the main thread to be able to wind down nicely if they occur. Kubernetes uses SIGTERM to take down pods, so this should make those more graceful

joshuacolvin0 (approved):
LGTM

packages/arb-rpc-node/batcher/batcher.go
```diff
@@ -237,46 +215,95 @@ func newBatcher(
...
-		if err != nil {
-			time.Sleep(5 * time.Second)
-			logger.Error().Err(err).Msg("error calling SendL2MessageFromOrigin, retrying")
-			continue
-		}
+	logger.Info().Int("txcount", len(txes)).Msg("Submitting batch")
+	batchData := message.NewSafeL2Message(batchTx).AsData()
+	txHash, err := globalInbox.SendL2MessageFromOrigin(ctx, batchData)
+	if err != nil {
+		return false, errors.Wrap(err, "error calling SendL2MessageFromOrigin")
```
PlasmaPower, line 252: Just double checking that the removal of retry logic here is intentional (maybe it was added somewhere else and I missed it?)

PlasmaPower (approved):
LGTM

packages/arb-rpc-node/batcher/batcher.go
hkalodner: So there was retry logic both here and in the code above this code. I decided that it would be nicer to flatten out the code and have a single codepath for retries in order to reduce the level of nesting of the code

joshuacolvin0 (approved):
LGTM

PlasmaPower (approved):
LGTM

PR #1063, opened 2021-04-27
Update deadline computation

edfelten:
Change the formula for computing a rollup block's deadline, to ensure that a block's deadline is not earlier than the deadlines of any of its older siblings.  We do this because we can't assume that a validator knows how to respond to a block, until it has evaluated the correctness of that block's older siblings.


hkalodner (approved):
LGTM

PR #1073, opened 2021-05-04
Implement broadcasting sequencer batch feed

joshuacolvin0:


packages/arb-node-core/cmd/arb-validator/arb-validator.go
```diff
@@ -106,6 +107,7 @@ func startup() error {
 	disableOpenEthereumCheck := flagSet.Bool("disable-openethereum-check", false, "disable checking the health of the OpenEthereum node")
 	healthcheckMetrics := flagSet.Bool("metrics", false, "enable prometheus endpoint")
 	healthcheckRPC := flagSet.String("healthcheck-rpc", "", "address to bind the healthcheck RPC to")
+	sequencerWebsocketURL := flagSet.String("sequencer-websocket-url", "", "websocket address of sequencer feed")
```
PlasmaPower, line 110: We never want to run the validator with the sequencer feed, as it cannot make assertions about what's not on chain yet. This functionality should not be present in the validator binary.

PlasmaPower (changes requested):

packages/arb-node-core/cmd/arb-validator/arb-validator.go
joshuacolvin0: removed

hkalodner (approved):
LGTM

PR #1083, opened 2021-05-07
Changes to support custom inboxes and outboxes

shemnon:
Changes to support custom inbox and outbox extension.
* make some private access internal
* make some functions virtual as well as override

Fix a versioning conflict:
* @ethersproject/bignumber -> 5.1.1, including forced resolution


shemnon:
Do you have a CLA or other license hurdle I need to accept?

hkalodner (approved):
Looks good to me. You're all setup CLA-wise

PR #1088, opened 2021-05-10
Reorg sequencer messages if max delay has passed

PlasmaPower:


packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -101,11 +104,19 @@ func NewSequencerBatcher(
...
+	maxDelaySeconds, err := sequencerInbox.MaxDelaySeconds(callOpts)
+	if err != nil {
+		return nil, err
+	}
+
 	feedBroadcaster := broadcaster.NewBroadcaster(broadcasterSettings)
 	err = feedBroadcaster.Start(ctx)
 	if err != nil {
 		logger.Warn().Err(err).Msg("error starting feed broadcaster")
-		return nil, err
```
hkalodner, line 108: Was removing this intentional?

packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -395,6 +408,58 @@ func (b *SequencerBatcher) createBatch(ctx context.Context, newMsgCount *big.Int
...
+		if newestMsgCount.Cmp(newMsgCount) > 0 {
+			newerBatchItems, err := b.db.GetSequencerBatchItems(newMsgCount, newestMsgCount)
+			if err != nil {
+				return false, err
+			}
+			batchItems = append(batchItems, newerBatchItems...)
+		}
+
+		var previousSeqBatchAcc common.Hash
+		if prevMsgCount.Cmp(big.NewInt(1)) > 0 {
```
hkalodner, line 431: Why is this greater than 1 instead of greater than 0?

hkalodner (changes requested):

packages/arb-rpc-node/batcher/sequencerBatcher.go
PlasmaPower: No, that got caught in a merge, thanks

packages/arb-rpc-node/batcher/sequencerBatcher.go
PlasmaPower: Interesting, I copied this from deliverDelayedMessages, which was also apparently broken haha. I'll fix them both.

hkalodner (approved):
LGTM

PR #1091, opened 2021-05-11
Refactor deployment

fredlacs:
Refactor deploy scripts and add script to upgrade contracts.

This deletes the hardhat deployment script used in tests, and instead uses the `hardhat-deploy` plugin to also manage deployments for test cases


packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -162,20 +191,23 @@ abstract contract AbsRollup is Cloneable, RollupCore, Pausable, IRollup {
      */
     function setInbox(address _inbox, bool _enabled) external onlyOwner {
         delayedBridge.setInbox(address(_inbox), _enabled);
+        emit OwnerFunctionCalled();
```
hkalodner, line 167: Would be good to know which one. Maybe a string or int arg to the event?

packages/arb-bridge-eth/scripts/upgrade.ts
```diff
@@ -0,0 +1,183 @@
...
+  if (accounts[0].address.toLowerCase() !== rollupCreatorOwner.toLowerCase()) {
+    throw new Error('Current account must be rollup creator owner')
+  }
+
+  const linkedBridgeCreator = await rollupCreator.bridgeCreator()
+  const linkedRollupTemplate = await rollupCreator.rollupTemplate()
+  const linkedChallengeFactory = await rollupCreator.challengeFactory()
+  const linkedNodeFactory = await rollupCreator.nodeFactory()
+
+  // check if current linked templates match the supplied previous ones
```
hkalodner, line 92: Are these supposed to be the addresses of the proxies or the implementations

hkalodner (changes requested):

PR #1092, opened 2021-05-11
Expose create mint call data

DZGoldman:
this is to make it easy for the client to query calldata size for the retryable's maxsubmission cost


fredlacs (approved):
LGTM

hkalodner (approved):
LGTM

PR #1094, opened 2021-05-11
Don't initialize broadcast client confirmation channel by default

PlasmaPower:
This was causing stalls for non-relay clients as the channel wasn't being read from.


joshuacolvin0 (approved):
LGTM

PR #1096, opened 2021-05-11
outbox/retryable  stuff in arb-ts

DZGoldman:
leaving this here but let's wait to merge until we have everything deployed & passing on kovan5


DZGoldman:
This is ready for review and such @fredlacs . The Ether e2e test will (/should) pass once the node fix is in for L1MessageType_L2FundedByL1 receipts.

packages/arb-ts/integration_test/arb-bridge.test.ts
```diff
@@ -352,18 +330,20 @@ describe('CustomToken', () => {
       )
       l2CustomToken = await customL2TokenFactory.deploy(
         bridge.arbTokenBridge.address,
-        l1CustomToken.address
+        l1CustomToken.address,
+        { gasLimit: 100000000 }
```
fredlacs, line 334: Should we be hardcoding gas values like this? If we're getting a wrong `eth_estimateGas` we should probably look into it

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -141,6 +167,22 @@ export class BridgeHelper {
     )
   }
 
+  static calculateRetryableAutoReedemTxnHash = async (
```
fredlacs, line 170: I don't particularly like this name, how about `calculateUserTxFromRetryable`

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -265,14 +307,14 @@ export class BridgeHelper {
...
-    const logs = l2Transaction.logs.filter(
+    const logs = l1Transaction.logs.filter(
       log =>
         log.topics[0] === eventTopics.InboxMessageDelivered ||
         log.topics[0] === eventTopics.InboxMessageDeliveredFromOrigin
     )
 
     if (logs.length === 0) return undefined
-    return logs.map(log => BigNumber.from(log.topics[1]))
+    return logs.map(log => BigNumber.from(log.topics[1]))[0]
```
fredlacs, line 317: I don't think we should return the first element of the array. We can keep this neutral since you could have triggered the event multiple times.

fredlacs (changes requested):

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: @edfelten

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: fair

packages/arb-ts/integration_test/arb-bridge.test.ts
DZGoldman: yes, this is a lazy patch atm

fredlacs (approved):
LGTM

PR #1097, opened 2021-05-12
Improve broadcaster logging for debugging

joshuacolvin0:


packages/arb-util/broadcastclient/broadcastclient.go
```diff
@@ -101,15 +101,19 @@ func (bc *BroadcastClient) backgroundReader(messageReceiver chan broadcaster.Bro
...
-
 		res := broadcaster.BroadcastMessage{}
 		err = json.Unmarshal(msg, &res)
 		if err != nil {
 			logger.Error().Err(err).Msg("error unmarshalling message")
 			continue
 		}
 
+		if len(res.Messages) > 0 {
+			logger.Debug().Hex("acc", res.Messages[0].FeedItem.BatchItem.Accumulator.Bytes()).Msg("received batch item")
```
PlasmaPower, line 112: I'm not sure if we ever send more than one message at a time, but we should probably log them or at least how many there are.

PlasmaPower (changes requested):

packages/arb-util/broadcastclient/broadcastclient.go
joshuacolvin0: sounds good

PlasmaPower (approved):
LGTM

hkalodner (approved):
LGTM

PR #1101, opened 2021-05-12
Refactor broadcast to use one thread per client instead of threadpool

joshuacolvin0:
Using threadpool to broadcast messages causes race condition that may deliver messages out of order

### DIFF
--- a/packages/arb-util/broadcaster/clientconnection.go
+++ b/packages/arb-util/broadcaster/clientconnection.go
@@ -31,6 +31,7 @@ import (
 	"github.com/mailru/easygo/netpoll"
 )
 
+const MaxSendCount = 10
 const MaxSendQueue = 20
 
 // ClientConnection represents client connection.
@@ -64,6 +65,7 @@ func (cc *ClientConnection) Start(parentCtx context.Context) {
 
 	cc.out = make(chan []byte, MaxSendQueue)
 	go func() {
+		defer cancelFunc()
 		defer close(cc.out)
 		for {
 			select {
--- a/packages/arb-util/broadcaster/clientconnection.go
+++ b/packages/arb-util/broadcaster/clientconnection.go
@@ -17,8 +17,12 @@
 package broadcaster
 
 import (
+	"context"
 	"encoding/json"
 	"io"
+	"math/rand"
+	"net"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


packages/arb-util/broadcaster/clientmanager.go
```diff
@@ -271,25 +263,22 @@ func (cm *ClientManager) startWriter(ctx context.Context) {
 			case <-ctx.Done():
 				return
 			case data := <-cm.out:
-				cm.mu.RLock()
-				// Copy list so data can be written to each client without lock held
-				clientList := make([]*ClientConnection, len(cm.clientPtrMap))
-				var i uint64
+				clientDeleteList := make([]*ClientConnection, 0, len(cm.clientPtrMap))
+				cm.mu.Lock()
```
hkalodner, line 267: Not sure if this will be an issue, but if we're getting a lot of messages this thread could try to hold the mutex almost all the time. Probably should leave as-is for now, but something to look out for

packages/arb-util/broadcaster/clientmanager.go
```diff
@@ -271,25 +263,22 @@ func (cm *ClientManager) startWriter(ctx context.Context) {
...
+				// Lock mutex while writing to channels to ensure items delivered in order
 				for client := range cm.clientPtrMap {
-					clientList[i] = client
-					i++
+					if len(client.out) > MaxSendQueue {
```
hkalodner, line 270: I don't think this can ever be true since the buffer is size `MaxSendQueue`

hkalodner (changes requested):

packages/arb-util/broadcaster/clientmanager.go
joshuacolvin0: good catch

packages/arb-util/broadcaster/clientconnection.go
```diff
@@ -27,15 +31,56 @@ import (
...
+	}
+}
+
+func (cc *ClientConnection) Start(parentCtx context.Context) {
+	ctx, cancelFunc := context.WithCancel(parentCtx)
+	cc.cancelFunc = cancelFunc
+
+	cc.out = make(chan []byte, MaxSendQueue)
+	go func() {
+		defer close(cc.out)
```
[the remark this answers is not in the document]
joshuacolvin0: :+1:

packages/arb-util/broadcaster/clientmanager.go
joshuacolvin0: sounds good, can send up to X number of messages if ready to send each time mutex is locked

hkalodner (approved):
LGTM

PR #1103, opened 2021-05-12
Fix race condition checking on expired clients

joshuacolvin0:


packages/arb-util/broadcaster/clientmanager.go
```diff
@@ -238,7 +238,9 @@ func (cm *ClientManager) verifyClients() {
 	var deadClientCount uint64
 	var aliveClientCount uint64
 	for client := range cm.clientPtrMap {
+		client.timeoutMutex.Lock()
```
hkalodner, line 241: I think it'd be a little cleaner to keep the mutex stuff in `ClientConnection` and just add a public `GetLestHeard()` that does the mutex control

hkalodner (changes requested):

packages/arb-util/broadcaster/clientmanager.go
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1108, opened 2021-05-13
eth_gasPrice and Arb cli

hkalodner:
- Add eth_gasPrice support
- Add arb CLI for interacting with the arbitrum chain


joshuacolvin0 (approved):
LGTM

PR #1109, opened 2021-05-13
Reorg out bad txs in sequencer mode

PlasmaPower:


packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -202,6 +221,93 @@ func (b *SequencerBatcher) SendTransaction(_ context.Context, startTx *types.Tra
...
+	var eachSucceeded []bool
+
+	newLogCount, err := b.db.GetLogCount()
+	if err != nil {
+		return err
+	}
+	txLogs, err := b.db.GetLogs(logCount, new(big.Int).Sub(newLogCount, logCount))
+	if err != nil {
+		return err
+	}
```
hkalodner, line 280: The machine may not be done processing the messages here. I think we might need to wait for it to be idle?

hkalodner (changes requested):

packages/arb-rpc-node/batcher/sequencerBatcher.go
PlasmaPower: I thought DeliverMessagesAndWait did that.. oops. We do need to wait for the machine to be idle.

PlasmaPower:
I've confirmed that at least the basics of this work via the sequencer test. It passes when the transactions don't have any value attached, but when the transactions do have value it fails as the account has no balance.

packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -155,6 +157,39 @@ func (b *SequencerBatcher) SubscribeNewTxsEvent(ch chan<- ethcore.NewTxsEvent) e
 	return b.newTxFeed.Subscribe(ch)
 }
 
+const maxExcludeComputation int64 = 10_000
+
+func shouldIncludeTxResult(txRes *evm.TxResult) bool {
+	if txRes.ResultCode == 0 {
```
hkalodner, line 163: We have constants for these. `evm.ReturnCode` and `evm.RevertCode` I think

packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -202,6 +256,106 @@ func (b *SequencerBatcher) SendTransaction(_ context.Context, startTx *types.Tra
...
+			newTxResults, err := txLogsToResults(txLogs)
+			if err != nil {
+				return err
+			}
+			if !newTxResults[txHash] {
+				err = core.DeliverMessagesAndWait(b.db, prevAcc, nil, nil, msgCount)
+				if err != nil {
+					return err
+				}
+				resultChans[i] <- errors.New("transaction errored")
```
hkalodner, line 348: We need to clean up the errors a bit. I think we can use https://github.com/OffchainLabs/arbitrum/blob/cc8d690934ba3bb7507bc72c34c35d363fc23b70/packages/arb-rpc-node/web3/eth.go#L208 with `ganacheMode` false, but it might need some refactoring since as-is there'd be a dependency loop. Could probably move the function to `evm`.

I think it'd also make sense

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1112, opened 2021-05-13
Fix reorg cache deleting checkpoints

PlasmaPower:
Even if we don't need to load a checkpoint, we still need to iterate over checkpoints to delete the old ones.


hkalodner (approved):
LGTM

PR #1104, opened 2021-05-13
Add some more metrics (eth/avm height)

shemnon:
Add some metrics to trace Ethereum block height and AVM block height.

PR #1122, opened 2021-05-14
Process and rpc metrics

shemnon:
Add standard Go and Process metrics.
Weave in RPC calls along with success/fail results.

(follow on to pr #1104)


joshuacolvin0:
Merged into #1232

PR #1127, opened 2021-05-16
Clean up golang abigen

hkalodner:
- Rather than an external dependency on `abigen` and `solc` with independent compilation settings for solidity compilation, this shifts the go abi generation over to using the go-ethereum library directly. This let's us get rid of a bunch of ugly shell scripting and should make abi generation way simpler and less error prone as dependencies change
- One big thing to note is that this assumes that you've run `yarn build` in `arb-bridge-eth` first since it depends on the build output of that process


joshuacolvin0 (approved):
Very cool

joshuacolvin0 (approved):
LGTM

PR #1132, opened 2021-05-18
return error when sideload not in database

joshuacolvin0:


PlasmaPower:
The Valid check should be after the status check, so if we hit some other error, that'll be returned instead of NotFound.

PlasmaPower (approved):
LGTM

PR #1134, opened 2021-05-18
Add cancel support in WaitToCatchup

joshuacolvin0:
Allow program to gracefully shutdown even when catching up with transactions during startup


hkalodner (approved):
LGTM

PR #1147, opened 2021-05-20
Make rpc/ws bind address configurable

joshuacolvin0:


packages/arb-rpc-node/cmd/arb-node/arb-node.go
```diff
@@ -119,10 +123,9 @@ func startup() error {
 		return errors.Wrap(err, "error parsing arguments")
 	}
 
-	if fs.NArg() != 3 || (!*sequencerMode && *sequencerURL == "") {
-		fmt.Printf("usage      sequencer: arb-node --sequencer [optional arguments] %s %s", cmdhelp.WalletArgsString, utils.RollupArgsString)
-		fmt.Printf("   or   primary node: arb-node --sequencer.addr=<sequencer address> --inbox=<inbox address> [optional arguments] %s %s", cmdhelp.WalletArgsString, utils.RollupArgsString)
```
hkalodner, line 124: I think we actually do want to keep this renamed to "aggregator node"

hkalodner (changes requested):

packages/arb-rpc-node/cmd/arb-node/arb-node.go
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1151, opened 2021-05-20
WIP: calculate path from batch size and leaf index client side

DZGoldman:
Not working as is: my understanding is the path to leaf at index 0 for tree of size 2 should be 1 (since leaves are added from the right); as is `generateEncodedMerklePath(0,2) => 0`.

A live (on kovan5) example txn pairing is https://explorer5.arbitrum.io/#/tx/0x4d601f2b51e326f28424e8c6fc8cddd4e01dda2caa432e0481de894f299593e5 (init eth withdraw) 
and https://kovan.etherscan.io/tx/0x51efc9b5ee3ee7af5d7d4d318a146fd94ce5f96fe28fce14d648698a01c2fb51 (execute in outbox).

It's batch — batch 60 (0x3c) — has size 2, and the L2 txn above is at index 0 (as per its `L2ToL1Transaction` log. The `OutBoxTransactionExecuted` log shows it emitted with a path (confusingly called "index" in the event) of 1.

I presume the issue here is still just with the generateEncodedMerklePath calculation (and not an arbos issue).

### DIFF
--- a/packages/arb-ts/src/lib/bridge_helpers.ts
+++ b/packages/arb-ts/src/lib/bridge_helpers.ts
@@ -91,6 +91,13 @@ export enum OutgoingMessageState {
   EXECUTED,
 }
 
+export interface OutboxEntryCreated {
+  indexed: BigNumber
+  outboxIndex: BigNumber
+  outputRoot: string
+  numInBatch: BigNumber
+}
+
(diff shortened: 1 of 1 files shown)


DZGoldman:
closing for https://github.com/OffchainLabs/arbitrum/pull/1157

PR #1156, opened 2021-05-21
Docs update

fredlacs:


docs/Differences_Ethereum_Overview.md
```diff
@@ -0,0 +1,62 @@
...
+### Transaction Receipts
+
+Transaction receipts contain the following extra fields
+
+#### L1 Block Number
+
+The Layer 1 block number for the transaction, as specified in [Time in Arbitrum](Time_In_Arbitrum.md).
+
+#### Fee Stats
+
```
DZGoldman, line 39: Maybe include table here too explaining each field? I think this would be quite nice, as quick shorthand way to understand the gas/fee architecture

docs/User_Quickstart.md
```diff
@@ -0,0 +1,31 @@
+---
+id: User_Quickstart
+title: Arbitrum User Quickstart
+sidebar_label: User Quickstart
+custom_edit_url:https://github.com/OffchainLabs/arbitrum/edit/master/docs/Developer_Quickstart.md
+---
+
+Arbitrum is a suite of Ethereum scaling solutions that enables high-throughput, low cost smart contracts while remaining trustlessly secure. Arbitrum has three modes: AnyTrust Channels, AnyTrust Sidechains, and Arbitrum Rollup. The following documentation describes how to use Arbitrum Rollup, which is currently live on testnet.
```
DZGoldman, line 8: Do we even want to mention the other two modes at this stage (/generally)? Pinging @edfelten @hkalodner @sgoldfed

DZGoldman (approved):
Looks good — pushed very miner tweaks and left a few suggestions, but def merge-able as-is.

docs/Differences_Ethereum_Overview.md
fredlacs: I think we should just have a whole page on the fee model, any shorthand here might just be lacking and more confusing imo

docs/User_Quickstart.md
fredlacs: I opted to keep this constant with the "developer quickstart" section. We can update this in a subsequent PR

PR #1157, opened 2021-05-21
use path returned by nodeinterface for messageHasExecuted

DZGoldman:
ugh, h/t @fredlacs, we can just use the path directly instead of https://github.com/OffchainLabs/arbitrum/pull/1151


fredlacs (approved):
LGTM

PR #1162, opened 2021-05-24
fix gas in deposit token

DZGoldman:
this is just the deposit fix from https://github.com/OffchainLabs/arbitrum/pull/1158 (none of the build-fix stuff)


fredlacs (approved):
lgtm

PR #1164, opened 2021-05-25
More optimizations

hkalodner:


packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -135,25 +152,116 @@ struct CodeSnapshot {
 };
```
PlasmaPower, line 152: This isn't unique to this PR, but `op_count` on the line above in `CodeSnapshot` should actually be called `next_segment_num` as that's its value.

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
```diff
@@ -247,9 +243,16 @@ SaveResults saveMachine(ReadWriteTransaction& transaction,
...
     }
 
-    auto status = saveMachineState(transaction, machine);
-    if (!status.ok()) {
-        return {0, status};
+    auto machine_save_res = saveMachineState(transaction, machine);
+    if (!machine_save_res.first.ok()) {
+        return {0, machine_save_res.first};
+    }
+    auto core_code = dynamic_cast<CoreCode*>(machine.machine_state.code.get());
```
PlasmaPower, line 250: It seems to me like this wouldn't contain a CoreCode in practice. Is this function called anywhere? Tests maybe? Maybe rename it to saveTestMachine if that's the only time it's usable.

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -34,8 +34,13 @@ namespace {
 constexpr auto max_code_segment_key = "max_code_segment";
 constexpr auto segment_key_prefix = std::array<char, 1>{87};
 constexpr auto segment_key_size = segment_key_prefix.size() + sizeof(uint64_t);
+constexpr auto metadata_value_size =
+    sizeof(uint32_t) + sizeof(uint64_t) + sizeof(uint64_t) + sizeof(uint64_t);
+constexpr auto segment_chunk_key_size =
+    segment_key_prefix.size() + sizeof(uint64_t) + sizeof(uint64_t) + 1;
```
PlasmaPower, line 40: This doesn't need to be in this PR, but we should consider putting these in their own tables.

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -135,25 +152,116 @@ struct CodeSnapshot {
...
+
     mutable std::mutex mutex;
     std::unordered_map<uint64_t, std::shared_ptr<CodeSegment>> segments;
     uint64_t next_segment_num;
 
+    const std::shared_ptr<CodeSegment>& getSegment(uint64_t segment_num) const {
+        return segments.at(segment_num);
+    }
+
+    uint64_t nextSegmentNum() { return next_segment_num++; }
```
PlasmaPower, line 235: We might want to make this throw. I'm pretty sure it should never be called outside maybe tests? In practice we should be adding segments to a RunningCode and then committing them to the CoreCode. Directly adding them to CoreCode in non-test usage might mean a conflict with RunningCode which would be very bad I think.

PlasmaPower (changes requested):

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -133,42 +211,88 @@ std::vector<unsigned char> prepareToSaveCodeSegment(
...
+    for (uint64_t i = 0; i < metadata.chunk_count; i++) {
+        auto key = segment_op_chunk_key(segment_id, i);
+        rocksdb::PinnableSlice val;
+        auto s = tx.refCountedGet(vecToSlice(key), &val);
+        if (!s.ok()) {
+            throw std::runtime_error("failed to load segment op chunk");
+        }
+        std::vector<unsigned char> data{val.data(), val.data() + val.size()};
+        val.Reset();
+        extractRawOperations(raw_ops, data);
```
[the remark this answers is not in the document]
hkalodner: Done

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -198,36 +303,88 @@ class Code {
...
+
+    CodePointStub addSegment() {
+        const std::lock_guard<std::mutex> lock(mutex);
+        return addSegmentImpl();
+    }
+
+    CodePointStub addOperation(const CodePointRef& ref, Operation op) {
+        const std::lock_guard<std::mutex> lock(mutex);
+        if (ref.segment < first_segment) {
+            return parent->addOperation(ref, std::move(op));
```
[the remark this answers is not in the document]
hkalodner: Done

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
hkalodner: Done

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
hkalodner: Renamed this method as suggested

packages/arb-avm-cpp/data_storage/src/value/code.cpp
hkalodner: Agreed it's somewhat awkward as is. My thinking is that we likely get some advantage by having the keys that will be read or written to nearby having a shared prefix so that rocksdb will hopefully be able to do the lookups more efficiently

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
hkalodner: This is used in a lot of different tests so I think we hold off on this for now and fix in a separate PR later

joshuacolvin0 (dismissed):
LGTM

PlasmaPower (approved):
LGTM

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -235,4 +349,110 @@ class Code {
...
+
+    uint64_t fillInCode(
+        std::unordered_map<uint64_t, std::shared_ptr<CodeSegment>>&
+            parent_segments,
+        const std::map<uint64_t, uint64_t>& segment_counts) const {
+        const std::lock_guard<std::mutex> lock(mutex);
+        auto it = segment_counts.lower_bound(first_segment);
+        auto end = segment_counts.end();
+        for (; it != end; ++it) {
+            parent_segments[it->first] = getSegment(it->first);
```
PlasmaPower, line 386: Can we assert that parent_segments doesn't have this key yet here? That would protect against collisions.

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -235,4 +349,110 @@ class Code {
...
+        }
+        return nextSegmentNum();
+    }
+
+    std::shared_ptr<Code> getParent() const { return parent; }
+
+    uint64_t getNextSegmentNum() const {
+        const std::lock_guard<std::mutex> lock(mutex);
+        return first_segment + segment_list.size();
```
PlasmaPower, line 395: This should call nextSegmentNum (and maybe the latter should be named more clearly).

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
```diff
@@ -235,4 +349,110 @@ class Code {
...
+    friend CodeBase<RunningCode>;
+
+    mutable std::mutex mutex;
+    uint64_t first_segment;
+    std::vector<std::shared_ptr<CodeSegment>> segment_list;
+
+    std::shared_ptr<Code> parent;
+
+    const std::shared_ptr<CodeSegment>& getSegment(uint64_t segment_num) const {
+        return segment_list.at(segment_num - first_segment);
```
PlasmaPower, line 362: If this forwards to parent when `segment_num < first_segment`, we could remove special casing in loadCodeSegment and loadCodePoint.

PlasmaPower (commented):
The rebase LGTM, I just noticed a couple areas for clarity in RunningCode. I might do another review of the (de)serialization changes later.

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -45,84 +50,152 @@ std::array<unsigned char, segment_key_size> segment_key(uint64_t segment_id) {
...
+                        key.begin());
+
+    auto big_id = boost::endian::native_to_big(segment_id);
+    auto big_id_ptr = reinterpret_cast<const char*>(&big_id);
+    it = std::copy(big_id_ptr, big_id_ptr + sizeof(big_id), it);
+
+    auto big_chunk = boost::endian::native_to_big(chunk);
+    auto big_chunk_ptr = reinterpret_cast<const char*>(&big_chunk);
+    std::copy(big_chunk_ptr, big_chunk_ptr + sizeof(big_chunk), it);
+    key.back() = 0;
```
PlasmaPower, line 67: Shouldn't this be `key.push_back(0);` or similar, instead of overwriting a byte of big_chunk? That seems dangerous given that it's big endian.

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -45,84 +50,152 @@ std::array<unsigned char, segment_key_size> segment_key(uint64_t segment_id) {
...
-    return cps;
 }
 
-void serializeCodePoint(const CodePoint& cp,
-                        std::vector<unsigned char>& serialized_code) {
-    // Ignore referemces to other code segments
-    serialized_code.push_back(cp.op.immediate ? 1 : 0);
-    serialized_code.push_back(static_cast<unsigned char>(cp.op.opcode));
-    marshal_uint256_t(cp.nextHash, serialized_code);
+void extractRawHashes(std::vector<uint256_t>& hashes, const char* ptr) {
```
PlasmaPower, line 157: IMO it's a bit dangerous that this doesn't take in a reference to the pointer whereas `extractRawOperations` does

PlasmaPower (changes requested):

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -133,42 +206,88 @@ std::vector<unsigned char> prepareToSaveCodeSegment(
...
+    std::vector<uint256_t> hashes;
+    hashes.reserve(metadata.chunk_count / 10 + 1);
+    for (uint64_t i = 0; i < metadata.chunk_count; i++) {
+        auto key = segment_hash_chunk_key(segment_id, i);
+        rocksdb::PinnableSlice val;
+        auto s = tx.refCountedGet(vecToSlice(key), &val);
+        if (!s.ok()) {
+            throw std::runtime_error("failed to load segment hash chunk");
+        }
+        std::vector<unsigned char> data{val.data(), val.data() + val.size()};
```
PlasmaPower, line 252: `data` is never used

packages/arb-avm-cpp/data_storage/src/value/code.cpp
```diff
@@ -133,42 +206,88 @@ std::vector<unsigned char> prepareToSaveCodeSegment(
...
     }
-    return serialized_code;
+    std::vector<unsigned char> hash_data;
+    marshal_uint64_t(snapshot.cached_hash_count - metadata.hash_count,
+                     hash_data);
+    for (uint64_t i = metadata.hash_count; i < snapshot.cached_hash_count;
+         i++) {
+        marshal_uint256_t(snapshot.segment->loadCachedHash(i), hash_data);
+    }
+    metadata.op_count = snapshot.op_count;
```
PlasmaPower, line 216: I think `hash_count` needs updated here too.

PlasmaPower (changes requested):

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
hkalodner: Currently the way this is setup, there's a mutex on every level, and getSegment doesn't acquire the mutex so it can't be chained up as is. The way this is setup currently, it's only used for the common implementation stuff in CodeBase to try and minimize duplication

There's not a great scheme for making the mutex and general safety properties of the code self-documenting in the current setup. I'm think about whether there'd be a way to reorganize it to make it more clear

packages/arb-avm-cpp/data_storage/src/value/code.cpp
hkalodner: key is an array rather than a vector so this should just set the last byte which is uninitialized before this

packages/arb-avm-cpp/data_storage/src/value/code.cpp
hkalodner: Agreed. Modified `extractRawHashes` to take the pointer by reference. At some point it might be good to move away from using in-out params for this, but that would be a much bigger lift

packages/arb-avm-cpp/data_storage/src/value/code.cpp
hkalodner: Fixed

packages/arb-avm-cpp/data_storage/src/value/code.cpp
PlasmaPower: Ah, I missed that, thanks. Could we add an assert before this that `it - key == segment_key_size - 1`?

packages/arb-avm-cpp/data_storage/src/value/code.cpp
hkalodner: Fixed

packages/arb-avm-cpp/avm_values/include/avm_values/code.hpp
hkalodner: Did some refactoring to make this more clear

PlasmaPower (dismissed):

PR #1165, opened 2021-05-25
Fix sequencer batcher retrying tx batch components individually

PlasmaPower:
In this case, the new block batch item wouldn't be correct.


hkalodner (approved):
LGTM

PR #1170, opened 2021-05-25
Optionally use RPC to get current eth gas price

PlasmaPower:
There's probably a better architecture for passing around the gas price url in arb-node but for now I've just extended the existing structure.


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #1174, opened 2021-05-26
make sure failure connecting to server doesn't block

joshuacolvin0:


packages/arb-util/broadcastclient/broadcastclient.go
```diff
@@ -60,81 +60,103 @@ func NewBroadcastClient(websocketUrl string, lastInboxSeqNum *big.Int) *Broadcas
 	}
 }
 
-func (bc *BroadcastClient) Connect() (chan broadcaster.BroadcastFeedMessage, error) {
+func (bc *BroadcastClient) Connect(ctx context.Context) (chan broadcaster.BroadcastFeedMessage, error) {
```
hkalodner, line 63: Looks like `Connect` can no longer return an error so it'd be good to simplify the return

joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1176, opened 2021-05-26
Update arb-os

hkalodner:


edfelten (approved):
Approve pending CI

PlasmaPower:
Shouldn't this be targeting `next`?

hkalodner:
No need since it's not compatibility breaking and the arbos version linked to sequencer2 already is different from the kovan5 one. Once we hit a arbos version that breaks backwards compatibility, we'll split off into next, which I think will be the next one which will be the next one

PR #1177, opened 2021-05-26
Rollup implementation facets

fredlacs:
This PR refactors the rollup contracts to have two separate logics.
It delegates calls from the owner to the Admin logic facet and calls from users to the User logic facet.

No changes in the logic were done, except:
 - Added the `whenNotPaused` modifier to the `completeChallenge` function
 - Added a validator whitelist
 - Sequencer bridge and challenge query the rollup for parameters
 
Certain functions on the user logic facet now have the `onlyValidator` modifier to prevent non-validators from calling them.
 
It is worth mentioning that both the `onlyValidator` and `whenNotPaused` checks can be refactored to be made during dispatch instead of in the target function themselves.

The `IRollup` interface was deleted, and now consumers are suggested to use the `RollupUserFacet` and `RollupAdminFacet` ABI's respectively.


hkalodner (dismissed):
LGTM

hkalodner (approved):
LGTM

PR #1180, opened 2021-05-27
Cleanup broadcastclient reconnect

joshuacolvin0:
Allow reconnect to be cancelled and don't delay initial connect.


hkalodner (approved):
LGTM

PR #1182, opened 2021-05-27
Add retry to batcher

hkalodner:


joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

packages/arb-util/broadcastclient/broadcastclient.go
```diff
@@ -164,7 +153,11 @@ func (bc *BroadcastClient) RetryConnect(ctx context.Context, messageReceiver cha
 		}
 
 		bc.RetryCount++
-		_ = bc.connect(ctx, messageReceiver)
+		_, err := bc.connect(ctx, messageReceiver)
+		if err == nil {
+			bc.retrying = false
```
PlasmaPower, line 158: This retrying field seems to never be used as far as I can tell

PlasmaPower (changes requested):
Other than that, LGTM

packages/arb-util/broadcastclient/broadcastclient.go
hkalodner: Yeah it's never used except I think in a test. This whole component needs some work, so this PR just needs to get it in a stable state

PR #1185, opened 2021-05-27
Fix infinite recursion

joshuacolvin0:
Remove possibility of infinite recursion when connection error in broadcastclient.

Also increase sleep in unit test, will refactor to remove sleep completely later.


PlasmaPower:
Doesn't this conflict with Harry's PR to the degree that now the old background reader will stop if the connection is successful?

joshuacolvin0:
> Doesn't this conflict with Harry's PR to the degree that now the old background reader will stop if the connection is successful?

Good catch

PlasmaPower (approved):
LGTM

PR #1190, opened 2021-05-27
Ethbridge proxy patch

fredlacs:
Add upgradeability to challenge, node and OSP contracts


packages/arb-bridge-eth/deploy/ChallengeFactory.ts
```diff
@@ -10,9 +10,29 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
   const osp2 = await deployments.get('OneStepProof2')
   const osp3 = await deployments.get('OneStepProofHash')
 
+  const proxyAdmin = await deploy('ProxyAdmin', {
```
hkalodner, line 13: Is this the only proxy admin in the L1? If not we should split it up into it's own deployment and use it for everything probably

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1215, opened 2021-05-29
Remove "patent pending"

hiddenlayer:
Followup to https://github.com/OffchainLabs/arbitrum/pull/808

My understanding is that [Arbitrum is no longer "patent pending"](https://twitter.com/arbitrum/status/1367216325219786755?lang=en). This PR removes this language from the package readmes.


edfelten (approved):
Thanks for catching this!

PR #1222, opened 2021-05-31
Improve sequencer batch submission

PlasmaPower:
Adds replace by fee support and moves waiting for the receipt to the background


hkalodner (approved):
LGTM

PR #1232, opened 2021-06-02
Add more metrics for the validator to the health check metrics feed

shemnon:
metrics covering the following -
* Go stats
* Process stats
* Eth and Arb heights
* RPC calls
* inbox message processing

including a CLI option to set a prefix for metric names.

Includes PRs #1104 and #1122


joshuacolvin0 (approved):
LGTM

PR #1238, opened 2021-06-03
Improve message when incorrect parameters supplied

joshuacolvin0:
When user provides incorrect parameters, format usage instructions to be more readable.

Also allow graceful shutdown during startup


hkalodner (approved):
LGTM

PR #1246, opened 2021-06-06
Add new C++ driven proof machine fuzzer

PlasmaPower:
This also features a new machine serialization format aimed to be suitable for fuzzing, and fixes various issues found by the fuzzer.


packages/arb-avm-cpp/CMakeLists.txt
```diff
@@ -57,7 +57,25 @@ set(AVM_VERSION_STRING ${AVM_VERSION_MAJOR}.${AVM_VERSION_MINOR}.${AVM_VERSION_P
...
+    SET(CMAKE_C_COMPILER "/usr/bin/clang" CACHE PATH "" FORCE)
+    SET(CMAKE_CXX_COMPILER "/usr/bin/clang++" CACHE PATH "" FORCE)
+    add_compile_options(-fsanitize=address,fuzzer-no-link)
+    add_link_options(-fsanitize=address,fuzzer)
+elseif (${AVM_FUZZER} MATCHES "honggfuzz")
+    SET(CMAKE_C_COMPILER "/usr/local/bin/hfuzz-clang" CACHE PATH "" FORCE)
+    SET(CMAKE_CXX_COMPILER "/usr/local/bin/hfuzz-clang++" CACHE PATH "" FORCE)
+    add_compile_options(-fsanitize=address)
+    add_link_options(-fsanitize=address)
+endif ()
```
hkalodner, line 78: I don't love the hardcoded paths. Maybe better to move to setting via the command line instead via `cmake -DCMAKE_CXX_COMPILER=afl-clang-lto++ ..` instead? I was looking at https://github.com/jefftrull/json_spirit/blob/develop/fuzzing/CMakeLists.txt for example

packages/arb-avm-cpp/avm_values/src/value.cpp
```diff
@@ -268,7 +269,19 @@ struct ValuePrinter {
...
+        if (b.data_length() <= 64) {
+            os << "0x";
+            std::ios prev_flags(nullptr);
+            prev_flags.copyfmt(os);
+            for (auto b : b.toFlatVector()) {
+                os << std::hex << std::setw(2) << std::setfill('0') << (int)b;
+            }
+            os.copyfmt(prev_flags);
+        } else {
+            os << b.hash();
```
hkalodner, line 282: I think either we should just print the data in all cases or we should change the hash output to include the word hash or something to distinguish between a 32 byte buffer and a buffer longer than 64 bytes

packages/arb-avm-cpp/fuzz_target/utils.hpp
```diff
@@ -0,0 +1,60 @@
...
+#pragma once
+
+#include <avm/machine.hpp>
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+template <class... Args>
+void fuzz_require(bool check, Args... message) {
```
hkalodner, line 26: I doubt performance is an issue here, but you likely want `Args&&...` for perfect forwarding. Also, nice variadic expansion below 😁

hkalodner (changes requested):

packages/arb-avm-cpp/avm_values/src/value.cpp
PlasmaPower: Hash isn't 0x prefixed but yeah probably

packages/arb-avm-cpp/fuzz_target/utils.hpp
PlasmaPower: Haha thanks I stole it from stack overflow. I'll fix it.

packages/arb-avm-cpp/CMakeLists.txt
PlasmaPower: My goal here was for this to be as easy to setup as possible. Ideally I'd specify a non-absolute path in the CMakeLists.txt but but that doesn't seem possible.

hkalodner (dismissed):
LGTM

hkalodner (approved):
LGTM

PR #1249, opened 2021-06-07
Fix eth_getStorageAt

hkalodner:


PlasmaPower (approved):
LGTM

PlasmaPower:
There's also a hexutil.Int right? I'm not sure of the comparative advantages/disadvantages, but would it support `0x0` where Bytes may not?

joshuacolvin0 (approved):
LGTM

hkalodner:
The bug was that we were using `hexutil.Big` which is the int version, but that rejects data with leading 0s as invalid which standard tooling expects to work. This shifts us over to accepting arbitrary hex input which is what we want

PR #1264, opened 2021-06-11
Fix issues with logs cursor

joshuacolvin0:
- When reorg deleted logs, logs cursor wouldn't register that logs were deleted until some other activity happened.
- If logs were deleted between when logs retrieved and logs confirmed, deleted logs would be ignored until some other activity happened.


hkalodner (approved):
LGTM

PR #1275, opened 2021-06-14
Add gateway addresses / update arb-ts readme 

DZGoldman:


packages/arb-ts/README.md
```diff
@@ -12,8 +12,8 @@ const connectedL1Wallet = new Wallet(myPrivateKey, ethProvider)
 const connectedL2Wallet = new Wallet(myPrivateKey, arbProvider)
 
 const bridge = new Bridge(
```
fredlacs, line 14: I think we should include the new "init" bridge pattern here
https://github.com/OffchainLabs/arbitrum/blob/next/packages/arb-ts/src/lib/bridge.ts#L55-L59

fredlacs (approved):
[editted out]

packages/arb-ts/README.md
fredlacs, line 14: I think we should include the new "init" bridge pattern here
https://github.com/OffchainLabs/arbitrum/blob/next/packages/arb-ts/src/lib/bridge.ts#L55-L59

fredlacs (changes requested):

fredlacs (approved):
LGTM

PR #1289, opened 2021-06-20
DIfferent default persistent storage per chain

joshuacolvin0:
Make `~/.arbitrum/<chain name>` the default directory to store database and keys

### DIFF
--- a/packages/arb-util/configuration/configuration.go
+++ b/packages/arb-util/configuration/configuration.go
@@ -24,27 +24,19 @@ import (
 
 var logger = log.With().Caller().Stack().Str("component", "configuration").Logger()
 
-type Bridge struct {
-	Utils struct {
-		Address string `koanf:"address"`
-	} `koanf:"utils"`
-}
-
 type FeedInput struct {
 	Timeout time.Duration `koanf:"timeout"`
 	URL     string        `koanf:"url"`
 }
 
 type FeedOutput struct {
-	Addr string `koanf:"addr"`
-	HTTP struct {
-		Timeout time.Duration `koanf:"timeout"`
-	} `koanf:"http"`
-	Port    string        `koanf:"port"`
-	Ping    time.Duration `koanf:"ping"`
-	Timeout time.Duration `koanf:"timeout"`
-	Queue   int           `koanf:"queue"`
-	Workers int           `koanf:"workers"`
+	Addr          string        `koanf:"addr"`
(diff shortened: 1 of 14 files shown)


packages/arb-util/configuration/configuration.go
```diff
@@ -200,9 +205,13 @@ func Parse(ctx context.Context) (*Config, *Wallet, *ethutils.RPCEthClient, *big.
 		return nil, nil, nil, nil, errors.Wrap(err, "error getting --l1.url")
 	}
 
+	if len(l1URL) == 0 {
+		return nil, nil, nil, nil, errors.New("required parameter --l1.url is missing")
+	}
+
```
hkalodner, line 211: Do we want to support a loading everything including this from a config file?

packages/arb-util/configuration/configuration.go
```diff
@@ -261,25 +276,59 @@ func Parse(ctx context.Context) (*Config, *Wallet, *ethutils.RPCEthClient, *big.
...
-		return out, nil, nil, nil, nil
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return nil, nil, nil, nil, errors.Wrap(err, "Unable to read users home directory")
 	}
 
-	if len(out.Persistent.Database.Path) == 0 {
-		out.Persistent.Database.Path = path.Join(out.Persistent.Storage.Path, "arbStorage")
+	// Make persistent storage directory relative to home directory if not already absolute
+	out.Persistent.Storage.Path = path.Join(homeDir, out.Persistent.Storage.Path)
```
hkalodner, line 285: Doesn't this force the path to always be in the home directory? I'm on board with using `~/.arbitrum` as a default, but I don't think we want to only support relative paths to the home dir since that seems unnecessarily restrictive

hkalodner (changes requested):
I think the fallback behavior of machine path vs url overrides my comment about unifying those in the previous PR

packages/arb-util/configuration/configuration.go
joshuacolvin0: yup, loading everything from config file is supported

packages/arb-util/configuration/configuration.go
joshuacolvin0: I mistakenly assumed path.Join would do the right thing.  Updated to allow absolute paths

PR #1294, opened 2021-06-22
Add upgrade validation using OZ ts libraries

fredlacs:


DZGoldman (approved):

DZGoldman (approved):

DZGoldman:
so the cbor dependency is indeed just incompatible with node w/ node < 10.16.3, can reproduce locally. we could just update the node version in CI / in the developer docs installation instructions, maybe it's time anyway, but idk.

DZGoldman:
or we could make openzeppelin/upgrades-core an optional-dependency

DZGoldman:
made it an optional dependency; can move it to dev dependencies if/when we require node v > 10 generally, but it's only used for this one script here so nbd

DZGoldman (approved):

PR #1305, opened 2021-06-24
Add unit test for l2 weth deposit and withdrawal

fredlacs:


packages/arb-bridge-peripherals/test/wethBridge.l2.ts
```diff
@@ -0,0 +1,118 @@
...
+    const sender = '0x6300000000000000000000000000000000000002'
+    const amount = '10'
+    const dest = sender
+
+    const data = ethers.utils.defaultAbiCoder.encode(
+      ['bytes', 'bytes'],
+      ['0x', '0x']
+    )
+
+    console.log('here')
```
DZGoldman, line 69: here.

DZGoldman (approved):

packages/arb-bridge-peripherals/test/wethBridge.l2.ts
fredlacs: same

PR #1306, opened 2021-06-25
allow empty strings to be used as wallet passwords

twt--:
This change allows support for 0 length wallet passwords.  The recent change to support config files made it so you are prompted for a password if you set an empty password on your wallet.


hkalodner (approved):
LGTM

PR #1307, opened 2021-06-25
take configuration params from ARBITRUM_* environment variables

twt--:
Allows config parameters to be set via environment variables, copying https://github.com/knadh/koanf#reading-environment-variables and using an env var prefix of `ARBITRUM_` and double underscores to represent dashes in parameter names


joshuacolvin0 (approved):
LGTM

PR #1308, opened 2021-06-25
Use `--node.type` and other config cleanups

joshuacolvin0:
* Read all configuration values from environment variables
* Use node.type to explicitly choose type of node to run


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #1310, opened 2021-06-25
Merge next into master

hkalodner:
Since the mainnet release has stabilized we're going to return to having master as the main development branch


joshuacolvin0 (approved):
LGTM

PR #1318, opened 2021-06-26
update integration tests

DZGoldman:
clean up / refactor, makes them all "deterministic" (or at least more self contained); some other little arb-ts updates


packages/arb-ts/src/lib/bridge.ts
```diff
@@ -655,20 +654,20 @@ export class Bridge {
...
     gatewayAddresses: string[]
   ) {
-    const gasPriceBid = await this.l2Bridge.l2Provider.getGasPrice()
+    const gasPriceBid = await this.l2Provider.getGasPrice()
 
     const maxSubmissionPrice = (
       await this.l2Bridge.getTxnSubmissionPrice(
         // 20 per address, 100 as buffer/ estimate for any additional calldata
-        100 + 20 * (tokenAddresses.length + gatewayAddresses.length)
+        300 + 20 * (tokenAddresses.length + gatewayAddresses.length)
```
DZGoldman, line 670: this actually fails w/ the lower value; this is only a convenience method for internal use, so let's just leave it

packages/arb-ts/integration_test/eth.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const etherBalance = await bridge.getL2EthBalance()
+
+    const totalEth = etherBalance
+      .add(ethToWithdraw)
+      .add(withdrawEthRec.gasUsed.mul(inWei[5]))
+
+    // TODO
+    console.log(
+      `This number should be zero...? ${initialBalance
```
DZGoldman, line 123: This is effectively working (i.e., number is v close to 0) but there is still some small amount of gas spent that isn't being accounted for. I think it's just because `getPricesInWei`  still isn't precise enough even w/ the blocktag; i.e., we would need a way to query what the price in wei was for the particular transaction.

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -225,7 +224,7 @@ export class L1Bridge {
     const gatewayAddress = await this.getGatewayAddress(erc20L1Address)
     return tokenData.ERC20.contract.functions.approve(
```
DZGoldman, line 225: this is a little rider i'm sneaking into this PR (approval amount as param)

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -219,9 +216,7 @@ export class L2Bridge {
     return this.walletAddressCache
   }
 
-  public async getAndUpdateL2EthBalance(): Promise<BigNumber> {
```
DZGoldman, line 222: another rider - stop storing balance as a class variable; useless (client can/will probably store it anyway) and confusing

DZGoldman:
merging onto gateway-patch so i could access the updated interfaces; after approval, i'll push up the updated abis and bump/publish the arb-ts version

packages/arb-ts/integration_test/testHelpers.ts
```diff
@@ -60,3 +83,77 @@ export const warn = (text: string) => {
   console.log(chalk.red(`WARNING: ${text}`))
   console.log()
 }
+
+export const instantiateRandomBridge = () => {
```
fredlacs, line 87: should we rename this to make it clear that the bridge itself is not random but rather the PK used in it

packages/arb-ts/integration_test/sanity.test.ts
```diff
@@ -0,0 +1,216 @@
...
+
+import { AeWETH__factory } from '../src/lib/abi/factories/AeWETH__factory'
+
+import { instantiateBridge } from '../scripts/instantiate_bridge'
+import { existentTestERC20, existentTestCustomToken } from './testHelpers'
+import { constants } from 'ethers'
+
+describe('sanity checks (read-only)', async () => {
+  it('standard gateways public storage vars properly set', async () => {
+    const { bridge, l1Network, l2Network } = await instantiateBridge()
```
fredlacs, line 22: any reason to not instantiate this once in a before hook instead of each test case? here we don't use a random wallet in each test

packages/arb-ts/integration_test/standarderc20.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  it.skip('deposits erc20 (with L2 Eth funding)', async () => {
+    const { bridge } = await instantiateRandomBridge()
+    await fundL1(bridge)
+    await fundL2(bridge)
+    await depositTokenTest(bridge)
+  })
+
+  it('withdraws erc20', async function () {
+    const tokenWithdrawAmount = BigNumber.from(1)
+    const { bridge } = await instantiateRandomBridge()
```
fredlacs, line 52: should we just instantiate this in the beforeEach hook?

packages/arb-ts/integration_test/testHelpers.ts
```diff
@@ -60,3 +83,77 @@ export const warn = (text: string) => {
...
+export const skipIfMainnet = (() => {
+  let chainId = ''
+  return async (testContext: Mocha.Context) => {
+    if (!chainId) {
+      const { l1Network } = await instantiateRandomBridge()
+      chainId = l1Network.chainID
+    }
+    if (chainId === '1') {
+      console.log("You're writing to the chain on mainnet lol stop")
+      testContext.skip()
```
fredlacs, line 156: why would we skip a particular test/context instead of just halting altogether?

packages/arb-ts/scripts/instantiate_bridge.ts
```diff
@@ -7,24 +7,29 @@ dotenv.config()
 
 const pk = process.env['DEVNET_PRIVKEY'] as string
 const mnemonic = process.env['DEV_MNEMONIC'] as string
+const verbose = process.env['VERBOSE'] as string
+
 const defaultNetworkId = 4
```
fredlacs, line 12: should we allow an env variable to override the default network id?

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -263,6 +262,21 @@ export class BridgeHelper {
     )
   }
 
+  static getOutboundTransferData = async (
+    gatewayAddress: string,
+    provider: providers.Provider,
+    filter: ethers.providers.Filter = {}
+  ) => {
+    const contract = L1ERC20Gateway__factory.connect(gatewayAddress, provider)
+    const logs = await BridgeHelper.getEventLogs(
```
fredlacs, line 271: This isn't type safe, we can instead use typechain interface for this

some sample code I used in a similar script
```
const event = bridge.l2Bridge.l2GatewayRouter.filters.OutboundTransferInitiated(null, user, null, null, null, null);

    const withdrawals = await bridge.l2Bridge.l2GatewayRouter.attach(l2Network.tokenBridge.l2ERC20Gateway).queryFilter(event, "0x0", "latest");
```

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -263,6 +262,21 @@ export class BridgeHelper {
...
+      'OutboundTransferInitiated',
+      contract,
+      [],
+      filter
+    )
+    return logs
+  }
+
   public static getEventLogs = (
```
fredlacs, line 280: I think we should use type safe alternatives even if we make our code a bit more WET.
typechain interfaces cover most of the boilerplate code

fredlacs (changes requested):
Left a couple small comments

packages/arb-ts/integration_test/testHelpers.ts
DZGoldman: i want to leave room for some tests to be acceptable to run still (i.e., sanity.test) so we skip on a case by case basis

packages/arb-ts/scripts/instantiate_bridge.ts
DZGoldman: so that's what the --networkID flag does...

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: yeah, let's do this update as a separate PR

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: agreed

DZGoldman: (separate PR)

packages/arb-ts/scripts/instantiate_bridge.ts
DZGoldman: kind nice having this be a command line flag (at least in my experience

packages/arb-ts/integration_test/standarderc20.test.ts
DZGoldman: prob, but punting

fredlacs (approved):
LGTM

PR #1320, opened 2021-06-27
Remove arb provider ethers / update CI

DZGoldman:
fixes https://github.com/OffchainLabs/arbitrum/issues/1319

clean up mono-repo; nobody should be using this _or even looking at it_ .
We can show it as a deprecated package here https://github.com/OffchainLabs/arb-provider-ethers-deprecated-

Added simple CI tests for arb-ts for now (build/lint)


fredlacs (approved):
One thing I'd add is pushing a depreaction notice in the repo's readme at https://github.com/OffchainLabs/arb-provider-ethers-deprecated-/blob/master/package.json

We can then publish that so it shows up in the npm page that its deprecated

Regardless, this PR LGTM

PR #1324, opened 2021-06-28
change branch download from sequencer2 to master

RahulMaganti47:


docs/Installation.md
```diff
@@ -71,7 +71,7 @@ Here are the important dependencies in case you are not running on a supported O
 Download the Arbitrum Monorepo from source:
 
 ```bash
-git clone -b sequencer2 https://github.com/offchainlabs/arbitrum.git
```
fredlacs, line 74: I think we should explicitly name the master branch here since we were previously telling users to use `sequencer2`

ie `git clone -b master https://github.com/offchainlabs/arbitrum.git`

fredlacs (approved):
LGTM

PR #1327, opened 2021-06-28
Total stake and fork metrics

shemnon:
Add two metrics
* arbitrum_ethereum_unresolved_fork is 1 when the staker detects non-linear unresolved nodes
* arbitrum_rollupwatcher_staked_total{account=<acct>} is the balance of the staker when the staker info is created


shemnon:
I have one minor concern, in that a staker that exits may persist in the metrics for the stakerinfo with the last non-zero stake.  Not sure where I would add code to make sure the metric is either deleted or set to zero, any suggestions?

joshuacolvin0 (approved):
LGTM

PR #1329, opened 2021-06-29
update token list generation script

DZGoldman:
- try/catch is for unset addresses (i.e., set to the zero addresses
- l1/l2 name mismatch is allowed (i.e., usdc)


fredlacs (approved):
LGTM

PR #1330, opened 2021-06-29
have tryOutboxExecute return txn result (not receipt)

DZGoldman:
fits general pattern of client handling waiting for receipt, etc.


packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -669,12 +666,11 @@ export class BridgeHelper {
 
     console.log('got proof')
 
-    const outboxExecuteTransactionReceipt = await BridgeHelper.tryOutboxExecute(
+    return await BridgeHelper.tryOutboxExecute(
```
fredlacs, line 669: should we just return the promise instead of `return await`?
I don't think it makes a difference in this case

DZGoldman: agreed

fredlacs (approved):
LGTM

PR #1335, opened 2021-06-29
Configurable environment variable prefix

joshuacolvin0:
The previous behavior of loading all environment variables with a prefix of `ARBITRUM_` as configuration options runs into a problem if


packages/arb-util/configuration/configuration.go
```diff
@@ -194,6 +195,8 @@ func ParseValidator(ctx context.Context) (*Config, *Wallet, *ethutils.RPCEthClie
 func ParseNonRelay(ctx context.Context, f *flag.FlagSet) (*Config, *Wallet, *ethutils.RPCEthClient, *big.Int, error) {
 	f.String("bridge-utils-address", "", "bridgeutils contract address")
 
+	f.String("env-prefix", "", "environment variables with given prefix will be loaded as configuration values")
```
hkalodner, line 198: Don't we want to also provide this for relays?

hkalodner (changes requested):

packages/arb-util/configuration/configuration.go
hkalodner: We probably want this in `beginCommonParse`

joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1336, opened 2021-06-30
Add log message when custom rollup address supplied

joshuacolvin0:
If the rollup address is specified the default mainnet/testnet settings are not applied, which could cause unexpected results.  Add additional log message in this case, that way it is easier to tell what is going on.


hkalodner (approved):
LGTM

PR #1338, opened 2021-06-30
Generic-Custom gateway documentation 

DZGoldman:


docs/Bridging_Assets.md
```diff
@@ -110,3 +110,29 @@ Note that in the system described above, one pair of Gateway contracts handles t
...
+
+### The Arbitrum Generic Custom Gateway
+
+Just because a token has requirements beyond what are offered via the StandardERC20 gateway, that doesn't necessarily mean that a unique Gateway needs to be taylor-made for the token in question. Our Generic-Custom Gateway is designed to be flexible enough to be suitable for most (but not necessarily all) custom fungible token needs. As a general rule:
+
+**If your token's total supply on L2 only ever increases when tokens are deposited from L1 (i.e., no L2-interest accrual, no special L2 minting affordance, etc), but your token requires customization beyond what the StandardERC20 Gateway provides, the Generic-Custom Gateway is likely the right solution for you!**
+
+Some examples of token features suitable for the Generic-Custom Gateway:
+
+- L2 token contract upgradable via a proxy
```
fredlacs, line 122: We could add that you can use custom tokens if you want to determine the L2 token address yourself

docs/Bridging_Assets.md
```diff
@@ -110,3 +110,29 @@ Note that in the system described above, one pair of Gateway contracts handles t
...
 
 No matter the complexity of a particular token's bridging needs, a gateway can in principle be created it to accommodate it within our canonical bridging system.
+
+### The Arbitrum Generic Custom Gateway
+
+Just because a token has requirements beyond what are offered via the StandardERC20 gateway, that doesn't necessarily mean that a unique Gateway needs to be taylor-made for the token in question. Our Generic-Custom Gateway is designed to be flexible enough to be suitable for most (but not necessarily all) custom fungible token needs. As a general rule:
+
+**If your token's total supply on L2 only ever increases when tokens are deposited from L1 (i.e., no L2-interest accrual, no special L2 minting affordance, etc), but your token requires customization beyond what the StandardERC20 Gateway provides, the Generic-Custom Gateway is likely the right solution for you!**
```
fredlacs, line 118: I think this warning is a bit confusing, we should re-word it a bit. We can brainstorm later :)

docs/Bridging_Assets.md
```diff
@@ -110,3 +110,29 @@ Note that in the system described above, one pair of Gateway contracts handles t
...
+Follow the following steps to get your token set up to use the Generic Custom Gateway
+
+**1. Deploy your token on Arbitrum**
+
+- Your token should conform to the minimum [IArbToken](./sol_contract_docs/md_docs/arb-bridge-peripherals/tokenbridge/arbitrum/IArbToken.md)
+  interface; i.e., it should have `bridgeMint` and `bridgeBurn` methods only callable by the L2CustomGateway contract, and the address of its corresponding Ethereum token accessible via `l1Address`. For an example implementation, see [TestArbCustomToken](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/test/TestArbCustomToken.sol).
+
+**2. Register Your Token on L1 to Your Token on L2 via the L1CustomGateway Contract**
+This can be done permissionlessly if your L1 token's contract can make an external call to `L1CustomGateway.registerTokenToL2` (see, i.e., [TestCustomTokenL1](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/test/TestCustomTokenL1.sol)), or via an admin call on our end.
+**2. Register Your Token on L1 To the L1Gateway Router**
```
fredlacs, line 137: Typo? Think you mean `**3.`

docs/Bridging_Assets.md
```diff
@@ -110,3 +110,29 @@ Note that in the system described above, one pair of Gateway contracts handles t
...
+
+Follow the following steps to get your token set up to use the Generic Custom Gateway
+
+**1. Deploy your token on Arbitrum**
+
+- Your token should conform to the minimum [IArbToken](./sol_contract_docs/md_docs/arb-bridge-peripherals/tokenbridge/arbitrum/IArbToken.md)
+  interface; i.e., it should have `bridgeMint` and `bridgeBurn` methods only callable by the L2CustomGateway contract, and the address of its corresponding Ethereum token accessible via `l1Address`. For an example implementation, see [TestArbCustomToken](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/test/TestArbCustomToken.sol).
+
+**2. Register Your Token on L1 to Your Token on L2 via the L1CustomGateway Contract**
+This can be done permissionlessly if your L1 token's contract can make an external call to `L1CustomGateway.registerTokenToL2` (see, i.e., [TestCustomTokenL1](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/test/TestCustomTokenL1.sol)), or via an admin call on our end.
```
fredlacs, line 136: Hm I don't like this `or via an admin call on our end.`
We could add a different subsection that is something like `what if my L1 token is already deployed and I can't call `L1CustomGateway.registerTokenToL2`? Then we present the different options: wrap in L1 then trigger custom token, wrap in L2 with a custom implementation, or just deploy L2 custom token and we can admin set it (we could add a note that this is a temporary measure to support already deployed tokens)

fredlacs (changes requested):
Looking good, left a couple comments that we can discuss

fredlacs:
Ah another example we could give: if you want to mint in the L1, you need to deploy your own gateway - our general custom gateway doesn't support it

docs/Bridging_Assets.md
DZGoldman: ah yeah, i like that

DZGoldman: That's a better 3rd option for the list; the current one is week, but just wanted 3 for aesthetic rule-of-3 reasons

docs/Bridging_Assets.md
DZGoldman: "wrap in L2 with a custom implementation" here I'm assuming you mean "bridge via standard gateway, and build a custom wrapper around the L2 arb-standard-token

docs/Bridging_Assets.md
DZGoldman: gave this rewording a stab, lmk what you think

fredlacs (approved):
LGTM

PR #1339, opened 2021-06-30
Pass JSON config string on command line or file in S3 bucket

joshuacolvin0:


packages/arb-util/configuration/configuration.go
```diff
@@ -250,10 +267,10 @@ func ParseNonRelay(ctx context.Context, f *flag.FlagSet) (*Config, *Wallet, *eth
 			err := k.Load(confmap.Provider(map[string]interface{}{
 				"bridge-utils-address":             "0x84efa170dc6d521495d7942e372b8e4b2fb918ec",
 				"feed.input.url":                   []string{"wss://arb1.arbitrum.io/feed"},
+				"node.chain-id":                    "42161",
```
hkalodner, line 270: I don't think this hurts, but this I don't think this is necessary since it's set via a commandline flag

hkalodner (approved):
LGTM. I'm a bit concerned about the difficulty of testing these config changing PRs since they're not covered well by our test harness. We should optimally figure out a scheme for doing validation on them

packages/arb-util/configuration/configuration.go
joshuacolvin0: I'm inclined to leave it as is, that way it matches the definition for rinkeby below.

PR #1343, opened 2021-07-02
Assorted minor log cleanups

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -403,6 +405,11 @@ rocksdb::Status ArbCore::reorgToMessageCountOrBefore(
             tx.checkpointDelete(checkpoint_it->key());
 
             checkpoint_it->Prev();
+            if (!checkpoint_it->status().ok()) {
```
hkalodner, line 408: If checkpoint_it->IsValid() is true, that means there's no error, so I think this can be taken out of the loop and put before the IsValid check after the loop

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
```diff
@@ -136,22 +136,31 @@ void deleteMachineState(ReadWriteTransaction& tx,
     ++segment_counts[parsed_state.pc.pc.segment];
     ++segment_counts[parsed_state.err_pc.pc.segment];
 
-    deleteCode(tx, segment_counts);
+    auto delete_code_status = deleteCode(tx, segment_counts);
 
     if (!delete_static_res.status.ok()) {
-        std::cout << "error deleting static in checkpoint" << std::endl;
+        std::cout << "error deleting static in checkpoint: "
```
hkalodner, line 142: These should probably be `cerr`

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: I was looking at old documentation, isn't actually needed, so removed the code

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1345, opened 2021-07-02
prevent unnecessary execution attempt in  triggerL2ToL1Transaction

DZGoldman:
catches the case when proof data is available but the txn's node isn't confirmed yet


fredlacs (approved):
LGTM

PR #1346, opened 2021-07-02
increase default retryable txn submission fee bump / other retryable tooling fixes

DZGoldman:
max base fee increase per window is (i.e., will be) 150%


fredlacs (commented):
I think a slightly bigger refactor would make sense here. 
Should we make a utility function that handles the bump up?

DZGoldman:
with the %150 increase per period cap, this should be pretty safe; what do you have in mind in particular?

DZGoldman:
- added % increase utility func
- increased default % increase for until the 150% cap is enforced

DZGoldman:
added some additional semi-related fixes that were surfaced here https://github.com/OffchainLabs/arbitrum/pull/1350

packages/arb-ts/src/lib/bridge_helpers.ts
```diff
@@ -865,4 +865,8 @@ export class BridgeHelper {
     const whiteList = Whitelist__factory.connect(whiteListAddress, l1Provider)
     return whiteList.isAllowed(address)
   }
+
+  static percentIncrease(num: BigNumber, increase: BigNumber) {
```
fredlacs, line 869: If we're doing a percent increase of 100%, wouldn't that be `num + num.mul(increase).div(100)`?

fredlacs (changes requested):
One small naming thing that can be misleading, looks good other than that

packages/arb-ts/src/lib/bridge_helpers.ts
DZGoldman: lel true, it's poorly named. I'll change the logic tho (and keep the name)

fredlacs (approved):
LGTM

PR #1349, opened 2021-07-06
Rollup patch

fredlacs:


packages/arb-bridge-eth/contracts/rollup/Rollup.sol
```diff
@@ -158,6 +169,27 @@ contract Rollup is RollupBase {
         require(success, "FAIL_INIT_FACET");
 
         emit RollupCreated(_machineHash);
+        require(isInit(), "INITIALIZE_NOT_INIT");
+    }
+
+    function postUpgradeInit(address newAdminFacet) external {
```
hkalodner, line 175: Should we put `postUpgradeInit` into the admin facet? Also we'll definitely need to upgradeAndCall when we do this upgrade since the init hook takes input

packages/arb-ts/scripts/triggerLogic.ts
```diff
@@ -0,0 +1,97 @@
...
+  const provider = new ethers.providers.JsonRpcProvider('http://127.0.0.1:8545')
+  const signer = new ethers.Wallet(process.env['DEVNET_PRIVKEY']).connect(
+    provider
+  )
+
+  // const rollup = "0xc12ba48c781f6e392b49db2e25cd0c28cd77531a"
+  const delayedBridgeLogic = '0xcb0da32914a683286ed2d4890e8157ffecc9bd06'
+  const rollupLogic = '0xae71755b42d1ef5fb365aeb4a74cb73992dd9fbe'
+  const eventBridgeLogic = '0x5c7355e46d5486583a1cc211701e25004231d9dd'
+
```
hkalodner, line 17: We need to get out of all of this awkward hardcoding of urls and addresses

packages/arb-ts/scripts/triggerLogic.ts
```diff
@@ -0,0 +1,97 @@
...
+    rollupLogic
+  )
+  await initEventBridge.wait()
+
+  const nodeFactory = '0x4442a659aaa27eeddf0a5d6ef7659290e92f678c'
+  const sequencerInbox = '0x0000000000000000000000000000000000000000'
+  const outbox = '0x0000000000000000000000000000000000000000'
+  const challengeFactory = '0x0000000000000000000000000000000000000000'
+
+  const tx = await rollup.initialize(
```
hkalodner, line 42: Initialization of the logic contract shouldn't be necessary now, should it?

hkalodner (changes requested):

packages/arb-bridge-eth/contracts/rollup/Rollup.sol
fredlacs: This is currently on the Rollup dispatch which I believe works well.
I agree this should be triggered in the `upgradeAndCall`, but that is done by the ProxyAdmin, which is different to the Rollup Owner that has access to the admin facet.

packages/arb-ts/scripts/triggerLogic.ts
fredlacs: Agreed, this was a one off script that has now been deleted.

packages/arb-ts/scripts/triggerLogic.ts
fredlacs: Not needed anymore. I now deleted the script

DZGoldman (approved):
All looks good to me - just out of curiosity, why the createNewNode logic moved up to Rollup.sol?

DZGoldman:
(will wait for @hkalodner before merging)

fredlacs:
> All looks good to me - just out of curiosity, why the createNewNode logic moved up to Rollup.sol?

So that we can access it in both admin and user facet. In hindsight, should probably have moved it to `RollupCore` instead.

packages/arb-bridge-eth/contracts/rollup/facets/RollupAdmin.sol
```diff
@@ -242,24 +242,29 @@ contract RollupAdminFacet is RollupBase, IRollupAdmin {
...
-        for (uint256 i = 0; i < stacker.length; i++) {
-            withdrawStaker(stacker[i]);
+    function forceRefundStaker(address[] memory staker) external override whenPaused {
+        for (uint256 i = 0; i < staker.length; i++) {
+            withdrawStaker(staker[i]);
         }
+        emit OwnerFunctionCalled(22);
     }
 
     function forceCreateNode(
```
hkalodner, line 270: It'd be good to talk through the uses of this function relative to standard node creation by a validator

packages/arb-bridge-eth/contracts/rollup/facets/RollupAdmin.sol
```diff
@@ -269,53 +274,77 @@ contract RollupAdminFacet is RollupBase, IRollupAdmin {
...
+        rollupEventBridge.nodeConfirmed(nodeNum);
 
-        // emit NodeConfirmed(
-        //     firstUnresolved,
-        //     afterSendAcc,
-        //     afterSendCount,
-        //     afterLogAcc,
-        //     afterLogCount
-        // );
+        emit NodeConfirmed(nodeNum, afterSendAcc, afterSendCount, afterLogAcc, afterLogCount);
```
hkalodner, line 347: If possible it'd be nice to do some refactoring to reduce the duplication between the regular rollup methods and the admin methods

packages/arb-bridge-eth/test/arbRollup.spec.ts
```diff
@@ -405,4 +436,172 @@ describe('ArbRollup', () => {
...
+    const postCode = await ethers.provider.getCode(challenge.address)
+    expect(postCode).to.equal('0x')
+
+    const challengeA = await rollupAdmin.currentChallenge(
+      await accounts[8].getAddress()
+    )
+    const challengeB = await rollupAdmin.currentChallenge(
+      await accounts[1].getAddress()
+    )
+    const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
```
hkalodner, line 535: Can use `ethers.constants.AddressZero`

packages/arb-bridge-eth/test/arbRollup.spec.ts
```diff
@@ -405,4 +436,172 @@ describe('ArbRollup', () => {
...
+    await expect(
+      rollup
+        .connect(accounts[2])
+        .addToDeposit(await accounts[2].getAddress(), { value: 5 })
+    ).to.be.revertedWith('Pausable: paused')
+
+    await rollupAdmin.resume()
+  })
+
+  it('should allow admin to truncate nodes', async function () {
```
hkalodner, line 458: This test is way broader than the title suggests. Should probably have a more general name. If there's time it'd also be good to consider ways to break this up into smaller chunks

hkalodner (changes requested):

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -8,7 +8,6 @@ import "./IRollupFacets.sol";
 abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
     function initialize(address _stakeToken) public virtual override;
 
-    // TODO: Configure this value based on the cost of sends
```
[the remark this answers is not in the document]
fredlacs: Agreed, added back in

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -427,4 +426,61 @@ contract RollupCore is IRollupCore {
...
+        }
+        uint256 nodeNum = latestNodeCreated() + 1;
+        frame.prevNode.childCreated(nodeNum);
+
+        bytes32 nodeHash =
+            RollupLib.nodeHash(hasSibling, prevHash, frame.executionHash, frame.sequencerBatchAcc);
+        nodeCreated(frame.node, nodeHash);
+
+        targets.rollupEventBridge.nodeCreated(nodeNum, prevNode, deadlineBlock, msg.sender);
+        require(nodeNum == latestNodeCreated(), "NODE_NOT_CREATED");
```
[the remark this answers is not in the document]
fredlacs: This never fails, works well as an `assert` but I don't think its worth the gas cost of reading from storage. 
When adding testing on a separate PR I removed this https://github.com/OffchainLabs/arbitrum/pull/1383/commits/bebf1b48b3b583e932182e91f38f409e0df8a0ba#diff-e93ed756aed574257e780c9fa918b0563e01b4f711bdfa633f6bf2ef196bb776L482

I'll remove it here too

packages/arb-bridge-eth/contracts/rollup/facets/RollupAdmin.sol
fredlacs: I added in a comment in these that make explicit the validations that are skipped

packages/arb-bridge-eth/contracts/rollup/facets/RollupAdmin.sol
fredlacs: Refactored for the `create new node` and `confirm node` to use more shared logic in the RollupCore.

The non-shared code is for the checks that the user rollup does that are skipped in the admin - should be easy to throw them around now.

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -423,64 +471,177 @@ contract RollupCore is IRollupCore {
         _nodes[nodeNum] = INode(0);
     }
 
+    function assertionGasUsed(RollupLib.Assertion memory assertion)
```
hkalodner, line 474: Maybe better in RollupLib?

packages/arb-bridge-eth/contracts/rollup/facets/RollupAdmin.sol
```diff
@@ -288,27 +289,25 @@ contract RollupAdminFacet is RollupBase, IRollupAdmin {
...
+                prevNode: prevNode,
+                sequencerInbox: sequencerBridge,
+                rollupEventBridge: rollupEventBridge,
+                nodeFactory: nodeFactory
+            }),
+            expectedNodeHash
+        );
 
+        // this does not validate msg.sender as a staker
         stakeOnNode(msg.sender, latestNodeCreated(), confirmPeriodBlocks);
```
hkalodner: We should discuss if having this here makes sense

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1351, opened 2021-07-06
Don't wait for broadcast feed connection before starting up

PlasmaPower:


packages/arb-node-core/cmd/arb-relay/arb-relay.go
```diff
@@ -141,19 +141,21 @@ func (ar *ArbRelay) Start(ctx context.Context) (chan bool, error) {
...
-			case <-ctx.Done():
-				return nil, errors.New("ctx cancelled broadcast client connect")
-			case <-time.After(5 * time.Second):
+		go (func() {
+			for {
+				err := client.ConnectWithChannel(ctx, messages)
+				if err == nil {
+					break
+				}
+				logger.Warn().Err(err).
```
hkalodner, line 150: Would be good to include the url in this warning so we know which connection is failing

packages/arb-rpc-node/cmd/arb-node/arb-node.go
```diff
@@ -196,20 +196,22 @@ func startup() error {
 		sequencerFeed = make(chan broadcaster.BroadcastFeedMessage, 1)
 		for _, url := range config.Feed.Input.URLs {
 			broadcastClient := broadcastclient.NewBroadcastClient(url, nil, config.Feed.Input.Timeout)
-			for {
-				err = broadcastClient.ConnectWithChannel(ctx, sequencerFeed)
-				if err == nil {
-					break
+			go (func() {
```
hkalodner, line 199: Worth refactoring into a `waitForFeedConnection` function that can be shared between this and the relay?

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1353, opened 2021-07-07
WIP Fireblocks support

joshuacolvin0:


packages/arb-node-core/cmdhelp/wallet.go
```diff
@@ -27,75 +28,122 @@ import (
...
+			keystore.StandardScryptN,
+			keystore.StandardScryptP,
+		)
+
+		if len(ks.Accounts()) > 0 {
+			account = ks.Accounts()[0]
+		}
+
+		if ks.Unlock(account, walletConfig.Password) != nil {
+			if len(walletConfig.Password) == 0 {
```
PlasmaPower, line 99: I've previously used an empty password with a wallet. I'm not sure if this was supported in the old version either but something to consider.

packages/arb-node-core/cmdhelp/wallet.go
```diff
@@ -27,75 +28,122 @@ import (
...
+				}
+				// Just return original unsigned transaction because fireblocks will handle signing
+				return tx, nil
+			},
 		}
 
-		bytePassword, err := terminal.ReadPassword(int(syscall.Stdin))
+		signer = func(data []byte) ([]byte, error) {
+			// Fireblocks cannot sign arbitrary data
+			return make([]byte, 32), nil
```
PlasmaPower, line 72: We should be careful that we aren't covering up any issues here. It's just the sequencer feed?

packages/arb-node-core/ethbridge/auth.go
```diff
@@ -92,7 +171,7 @@ func (t *TransactAuth) makeContract(ctx context.Context, contractFunc func(auth
 
 		t.auth.Nonce = t.auth.Nonce.Add(t.auth.Nonce, big.NewInt(1))
 		auth.Nonce = t.auth.Nonce
-		addr, tx, _, err = contractFunc(auth)
+		err = t.sendTx(ctx, tx)
```
PlasmaPower, line 174: Is this going to actually effect the nonce change? We might just want to get rid of this code IMO, I've never seen it actually used.

packages/arb-util/configuration/configuration.go
```diff
@@ -567,10 +610,34 @@ func endCommonParse(k *koanf.Koanf) (*Config, *Wallet, error) {
 	}
 	err := k.UnmarshalWithConf("", &out, koanf.UnmarshalConf{DecoderConfig: &decoderConfig})
 	if err != nil {
-
 		return nil, nil, err
 	}
 
+	if len(out.Fireblocks.APIKey) != 0 || len(out.Fireblocks.BaseURL) != 0 ||
+		len(out.Wallet.FireblocksPrivateKey) != 0 || len(out.Fireblocks.SourceAddress) != 0 ||
+		len(out.Fireblocks.SourceId) != 0 || len(out.Fireblocks.SourceType) != 0 {
```
PlasmaPower, line 618: Could you do `if out.Fireblocks != (Fireblocks{})`?

packages/arb-util/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+
+	resp, err := fb.postRequest("/v1/transactions", url.Values{}, body)
+	if err != nil {
+		return nil, err
+	}
+
+	var result CreateTransactionResponse
+	response, err := ioutil.ReadAll(resp.Body)
+	fmt.Println("Response: ", string(response))
```
PlasmaPower, line 329: I'm guessing this was some leftover debugging code. This should probably be moved to a proper logger.

packages/arb-util/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,512 @@
...
+			Msg("error returned when posting fireblocks request")
+		return nil, fmt.Errorf("status '%s' fireblocks response", resp.Status)
+	}
+
+	return resp, nil
+}
+
+func (fb *Fireblocks) signJWT(path string, body []byte) (string, error) {
+	newPath := strings.Replace(path, "[", "%5B", -1)
+	newPath = strings.Replace(newPath, "]", "%5D", -1)
```
PlasmaPower, line 492: Go seems to have a URL escape in the standard library if that's what you're looking for: https://pkg.go.dev/net/url#PathEscape

packages/arb-util/fireblocks/fireblocks.go
```diff
@@ -0,0 +1,512 @@
...
+	now := time.Now().Unix()
+	if body == nil {
+		body = []byte("null")
+	}
+
+	bodyHash := sha256.Sum256([]byte(body))
+	fmt.Println("body: (", string(body), ")[", len(body), "], Hash ", hex.EncodeToString(bodyHash[:]))
```
PlasmaPower, line 499: Should be moved to a proper logger

PlasmaPower (changes requested):

packages/arb-node-core/cmdhelp/wallet.go
joshuacolvin0: The old version always prompted for password if wallet failed to open.  The new version only prompts for password if wallet failed to open and no password (or empty password) is specified.

Empty password can still be used by just pressing return at prompt, or providing a pregenerated wallet.  I think that is a reasonable compromise

packages/arb-util/configuration/configuration.go
joshuacolvin0: The idea is to catch an error such that if any field is set, all the fields should be set.

packages/arb-node-core/ethbridge/auth.go
joshuacolvin0: We ran into nonce issues early on that were mostly tracked down to running multiple validators with same wallet.  I agree that it is good to remove this code, because nonce issues are usually a symptom of something else being wrong, so better to not gloss over it.

packages/arb-util/fireblocks/fireblocks.go
joshuacolvin0: :+1:

packages/arb-util/fireblocks/fireblocks.go
joshuacolvin0: newPath is never used in an actual URL, it is just used in generating signature.  I wanted to duplicate the exact same thing that the other fireblocks API libraries do to ensure I came up with the exact same signature.

packages/arb-util/fireblocks/fireblocks.go
joshuacolvin0: :+1:

packages/arb-node-core/cmdhelp/wallet.go
joshuacolvin0: Correct, just the sequencer feed. I'm going to open a separate PR so that a separate key can be provided for feed signing.

joshuacolvin0: Added `config.Wallet.FeedPrivateKey` so that separate key can be provided for signing feed.

PR #1355, opened 2021-07-07
use L2gateway for address querying in arb-ts

RahulMaganti47:
added functionality to ensure arb-ts looks at l2 router instead of l1

### DIFF
--- a/packages/arb-ts/src/lib/abi/L1ERC20Gateway.d.ts
+++ b/packages/arb-ts/src/lib/abi/L1ERC20Gateway.d.ts
@@ -453,7 +453,7 @@ export class L1ERC20Gateway extends Contract {
 
   counterpartGateway(overrides?: CallOverrides): Promise<string>
 
-  ' ()'(overrides?: CallOverrides): Promise<string>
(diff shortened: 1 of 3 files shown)


DZGoldman (changes requested):
Thanks @RahulMaganti47! Run this through the pre-commit hook, and also make sure it builds without errors (`yarn workspace arb-ts build`; it might just be the linting that's triggering the CI error)

packages/arb-ts/src/lib/abi/L1ERC20Gateway.d.ts
```diff
@@ -453,7 +453,7 @@ export class L1ERC20Gateway extends Contract {
 
   counterpartGateway(overrides?: CallOverrides): Promise<string>
 
-  'counterpartGateway()'(overrides?: CallOverrides): Promise<string>
```
fredlacs, line 456: Accidental change?

RahulMaganti47: Yea my bad

DZGoldman:
merge conflict (also needs a pre-commit pass)

PR #1360, opened 2021-07-09
arb-bridge-eth contracts slither static coverage / review 

DZGoldman:


packages/arb-bridge-eth/contracts/bridge/SequencerInbox.sol
```diff
@@ -102,9 +100,11 @@ contract SequencerInbox is ISequencerInbox, Cloneable {
                     gasPriceL1,
                     messageDataHash
                 );
+            // Can only force-include after the Sequencer-only window has expired.
             require(l1BlockAndTimestamp[0] + maxDelayBlocks < block.number, "MAX_DELAY_BLOCKS");
             require(l1BlockAndTimestamp[1] + maxDelaySeconds < block.timestamp, "MAX_DELAY_TIME");
 
+            // Ensure that the penultimate message on the delayed queue was the last delayed message added to the sequencer queue
```
hkalodner, line 107: I think this is more about making sure the message hash calculated above was actually the last message in the chunk of delayed messages we'll be including

This is what we need for the timestamp check since all messages before it in the delayed inbox came at the same time or earlier

packages/arb-bridge-eth/contracts/bridge/SequencerInbox.sol
```diff
@@ -361,41 +376,68 @@ contract SequencerInbox is ISequencerInbox, Cloneable {
...
+        external
+        view
+        override
+        returns (uint256, bytes32)
+    {
+        return proveInboxContainsMessageImp(proof, _messageCount);
     }
 
-    function proveBatchContainsSequenceNumber(bytes calldata proof, uint256 inboxCount)
+    function proveBatchContainsSequenceNumber(bytes calldata proof, uint256 _messageCount)
```
hkalodner, line 399: Maybe add a deprecated in favor of `proveInboxContainsMessage` comment here

packages/arb-bridge-eth/contracts/bridge/interfaces/ISequencerInbox.sol
```diff
@@ -68,6 +68,11 @@ interface ISequencerInbox {
 
     function getInboxAccsLength() external view returns (uint256);
 
+    function proveInboxContainsMessage(bytes calldata proof, uint256 inboxCount)
+        external
+        view
+        returns (uint256, bytes32);
+
     function proveBatchContainsSequenceNumber(bytes calldata proof, uint256 inboxCount)
```
hkalodner, line 76: I think we can drop this from the interface

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -262,6 +251,7 @@ contract RollupCore is IRollupCore {
         bytes32 afterSendAcc = RollupLib.feedAccumulator(sendsData, sendLengths, beforeSendAcc);
 
         INode node = getNode(nodeNum);
+        // Replay protect:
```
hkalodner, line 254: This is more of a pre-image review to get the accumulators and counts out of the node than replay protection

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -236,7 +259,7 @@ abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
     }
 
     /**
-     * @notice Start a challenge between the given stakers over the node created by the first staker assuming that the two are staked on conflicting nodes
+     * @notice Start a challenge between the given stakers over the node created by the first staker assuming that the two are staked on conflicting nodes. N.B.: challenge creator must be *a* staker, but not necessarily one of the two in the challenge.
```
hkalodner, line 262: I don't think this is the case, though they do have to be on the whitelist

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1371, opened 2021-07-12
bridge-peripherals contracts slither static coverage / review 

DZGoldman:
Currently you have to yarn link `arb-bridge-eth` and `arbos-contracts` to get slither to run in this repo; also you have to remove MMR and the contracts that use it (only stuff in Buddy deployer) since Slither errors on ternaries (and --filter-paths doesn't seem to be working)


packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/ClonableBeaconProxy.sol
```diff
@@ -15,7 +15,7 @@ contract ClonableBeaconProxy is BeaconProxy {
 }
 
 contract BeaconProxyFactory is ProxySetter {
```
fredlacs, line 17: We can't really make breaking changes to this contract.
It wasn't deployed behind a proxy to ensure `type(ClonableBeaconProxy).creationCode` wouldn't break through upgrades

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/aeERC20.sol
```diff
@@ -28,12 +28,12 @@ contract aeERC20 is ERC20PermitUpgradeable, ERC677Token {
     using AddressUpgradeable for address;
 
     function _initialize(
-        string memory name,
-        string memory symbol,
-        uint8 decimals
+        string memory name_,
```
fredlacs, line 31: I think we should keep the underscores at the start (ie `_name`) for function parameters to keep it consistent.

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/aeWETH.sol
```diff
@@ -24,13 +24,13 @@ import "./IWETH9.sol";
 /// @title Arbitrum extended WETH
 contract aeWETH is L2GatewayToken, IWETH9 {
     function initialize(
-        string memory _name,
-        string memory _symbol,
-        uint8 _decimals,
-        address _l2Gateway,
-        address _l1Address
+        string memory name_,
```
fredlacs, line 27: Any reason to place the underscore at the end rather than the start?

packages/arb-bridge-peripherals/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "arb-bridge-peripherals",
   "version": "1.0.0",
   "license": "Apache-2.0",
+  "private": true,
```
fredlacs, line 5: I don't believe we want to make this package private. We can instead create the openzeppelin workspace path in the top level package.json which is already private

packages/arb-bridge-peripherals/package.json
```diff
@@ -32,6 +37,10 @@
     "typechain": "^4.0.0",
     "typescript": "^4.1.5"
   },
+  "dependencies": {
```
fredlacs, line 40: Is there a particular reason to change these to dependencies rather than devDependency?

packages/arb-bridge-peripherals/package.json
```diff
@@ -11,14 +12,18 @@
...
+      "@openzeppelin/contracts",
+      "@openzeppelin/contracts-upgradeable"
+    ]
   },
   "devDependencies": {
     "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@nomiclabs/hardhat-waffle": "^2.0.0",
-    "@openzeppelin/contracts": "3.4.0",
-    "@openzeppelin/contracts-upgradeable": "3.4.0",
-    "@openzeppelin/upgrades-core": "^1.7.6",
```
fredlacs, line 21: Seems like you accidentally removed `upgrades-core` which is used to check storage layouts before upgrades

fredlacs (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/aeERC20.sol
DZGoldman: _name etc. shadows storage var declarations in ERC20; I think the convention generally worth sticking to is _var for internal variables, var_ for params

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/aeWETH.sol
DZGoldman: https://github.com/OffchainLabs/arbitrum/pull/1371/files#r668794284

packages/arb-bridge-peripherals/package.json
DZGoldman: The hoisting didn't seem to be working w/ them as dev dependencies, but I'll take another look

packages/arb-bridge-peripherals/package.json
DZGoldman: this do was done for hoisting. r.e. workspace path, does yarn let you do that? ("foist" a package down to a particular package's node modules?)

packages/arb-bridge-peripherals/package.json
DZGoldman: ah indeed

fredlacs:
No more need to link packages

fredlacs (approved):
LGTM

PR #1373, opened 2021-07-12
Add missing message delivery mutex

PlasmaPower:
This should only matter if the sequencer is connected to a feed with its own txs


hkalodner (approved):
LGTM

PR #1381, opened 2021-07-14
Rollup test

fredlacs:


DZGoldman (approved):
I like it; was thinking maybe we should have the onlyValidator check on forceConfirmNode to preserve that invariant, but prob unnecessary since it's admin anyway. LGTM

PR #1388, opened 2021-07-16
Refactor rollup timing to account for time spent paused

fredlacs:


packages/arb-bridge-eth/contracts/rollup/INode.sol
```diff
@@ -60,7 +60,7 @@ interface INode {
 
     function latestChildNumber() external view returns (uint256);
 
-    function requirePastDeadline() external view;
+    function requirePastDeadline(uint256 blocksSpentPaused) external view;
```
hkalodner, line 63: We might want to consider having the node query from the rollup to make the API simpler

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -13,6 +13,17 @@ abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
         _;
     }
 
+    function blocksSpentPaused() public view override returns (uint256) {
```
hkalodner, line 16: If we pause and unpause repeatedly, this will only capture the length of the last pause

hkalodner (changes requested):
Leaving a couple comments, but I'm gonna need to stare at this more since I'm nervous about  the implementation here. I had a bit different implementation in mind where we record a total time spent paused since the beginning of time variable and then add it to `block.number` anywhere that is used

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
fredlacs: Made it so time paused accumulates through many pauses, and we can reset this timer

PR #1391, opened 2021-07-18
Add tuple inlining (targets more-optimizations2)

PlasmaPower:


joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

packages/arb-avm-cpp/data_storage/src/value/value.cpp
```diff
@@ -43,6 +43,15 @@ struct ValueBeingParsed {
     }
 };
 
+constexpr uint64_t tupleInlineNumerator = 7;
+constexpr uint64_t tupleInlineDenominator = 8;
+bool shouldInlineTuple(const Tuple& tuple,
+                       const std::vector<unsigned char>& secret_hash_seed) {
+    auto hash = tuple.getHashPreImage().secretHash(secret_hash_seed);
+    return tuple.tuple_size() == 0 ||
+           (hash % tupleInlineDenominator) < tupleInlineNumerator;
```
hkalodner, line 52: I wondering whether it would be better to have small tuples be more likely to inline than larger tuples. Likely not a change to make here, but an interesting possible future extension

hkalodner (changes requested):

packages/arb-avm-cpp/avm_values/include/avm_values/tuple.hpp
```diff
@@ -125,6 +125,30 @@ class Tuple {
         return tpl->data[pos];
     }
 
+    [[nodiscard]] std::vector<value>::const_iterator begin() const {
+        if (tpl == nullptr) {
+            return std::vector<value>().begin();
```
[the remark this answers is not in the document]
PlasmaPower: My thinking was that an empty vector is guaranteed not to allocate I believe, so it should always be a null pointer. But you're right, a static is safer. Fixed.

packages/arb-avm-cpp/data_storage/src/value/value.cpp
PlasmaPower: I'm not sure. A lot of trees in ArbOS are two-tuples right? It would definitely be an interesting experiment.

hkalodner: I'm not totally sure. It'd definitely be interesting to do some analysis on the latest machine state on rinkeby or mainnet to get a better idea of the structure of the register

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM, re-approving because all parts of this have been independently already reviewed

PR #1396, opened 2021-07-19
maximum outgoing message tests

DZGoldman:


packages/arb-bridge-eth/test/rolluplib.ts
```diff
@@ -344,7 +345,7 @@ export class RollupContract {
     afterLogCount: BigNumberish
   ): Promise<ContractTransaction> {
     const messageData = ethers.utils.concat(sends)
-    const messageLengths = sends.map(msg => msg.length)
+    const messageLengths = sends.map(msg => (msg.length - 2) / 2)
```
hkalodner, line 348: `ethers.utils.hexDataLength()` would be safer here

packages/arb-bridge-eth/test/arbRollup.spec.ts
```diff
@@ -903,4 +952,46 @@ describe('ArbRollup', () => {
       node.connect(accounts[0]).addStaker(await accounts[1].getAddress())
     ).to.be.revertedWith('ALREADY_STAKED')
   })
+
+  it('should re-initialize', async function () {
```
hkalodner, line 956: At first this sounded like calling initialize again on the same rollup. Maybe instead say "should initialize a fresh rollup chain"

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #1405, opened 2021-07-21
add some inline comments, remove important TODO (done)

DZGoldman:


packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -214,7 +213,7 @@ abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
             CreateNodeDataFrame({
                 arbGasSpeedLimitPerBlock: arbGasSpeedLimitPerBlock,
                 confirmPeriodBlocks: confirmPeriodBlocks,
-                prevNode: latestStakedNode(msg.sender),
+                prevNode: latestStakedNode(msg.sender), // Ensure staker is staked on the previous node
```
fredlacs, line 216: I don't think this comment is actually adding clarity.

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -352,6 +352,7 @@ contract RollupCore is IRollupCore {
 
     /**
      * @notice Reduce the stake of the given staker to the given target
+     * @dev Value deducted is accessible by pulling from _withdrawableFunds mapping
```
fredlacs, line 355: I think the "withdrawableFunds" is self explanatory.

It might be interesting to instead mention the top level function that is callable to access your `_withdrawableFunds`. Also this is an internal function, it might be worth instead adding this comment to the top level function that calls `reduceStakeTo`

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -369,15 +368,18 @@ abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
...
 
+        // Reward the winner with half the remaining stake
         uint256 amountWon = remainingLoserStake / 2;
         increaseStakeBy(winningStaker, amountWon);
         remainingLoserStake = remainingLoserStake.sub(amountWon);
         clearChallenge(winningStaker);
-
+        // Send the other half to the owner address
         increaseWithdrawableFunds(owner, remainingLoserStake);
+        // Turning loser into zombie renders the loser's remaining stake inaccessible
```
fredlacs, line 382: Should we instead add this comment to the `turnIntoZombie` function?
Or in both

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
```diff
@@ -369,15 +368,18 @@ abstract contract AbsRollupUserFacet is RollupBase, IRollupUser {
...
             remainingLoserStake = remainingLoserStake.sub(reduceStakeTo(losingStaker, winnerStake));
         }
 
+        // Reward the winner with half the remaining stake
         uint256 amountWon = remainingLoserStake / 2;
         increaseStakeBy(winningStaker, amountWon);
         remainingLoserStake = remainingLoserStake.sub(amountWon);
         clearChallenge(winningStaker);
-
+        // Send the other half to the owner address
```
fredlacs, line 380: Technically we don't send funds to the owner address. Instead we make them withdraw-able.

fredlacs (changes requested):

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
DZGoldman: ye, maybe "credit to the owner address"

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
DZGoldman: Mmm, i found this hard to find on my first pass. i.e., this is a thing that's important to enforce, and I want it to be clear (somehow) where in the code it is that we enforce it. Maybe there's a better/clearer way?

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
DZGoldman: ye- removing and adding a line to reduceDeposit (generally, wanted to make it more obvious which methods are moving funds in and out of stake and which are slashing)

packages/arb-bridge-eth/contracts/rollup/facets/RollupUser.sol
DZGoldman: I kinda like it just here, as this is the only call that actually slashes

fredlacs (approved):
LGTM

PR #1407, opened 2021-07-22
Update outbox execution docstring for clarity

fredlacs:


packages/arb-bridge-eth/contracts/bridge/Outbox.sol
```diff
@@ -112,7 +112,9 @@ contract Outbox is IOutbox, Cloneable {
     }
 
     /**
-     * @notice Executes a messages in an Outbox entry. Reverts if dispute period hasn't expired
+     * @notice Executes a messages in an Outbox entry.
+     * @dev Reverts if dispute period hasn't expired, since the outbox entry
+     * is only created once the rollup confirms the respective assertion.
```
DZGoldman, line 117: yeah the initial comment was bad (i wrote it). Maybe get rid of the "reverts if" language altogether, and instead go w/ "outbox entry is created only after the dispute period expires and its assertion is confirmed"

(basically just making it clear "this is the thing you have to wait before calling" was the impetus

PR #1416, opened 2021-07-23
remove 677 reference for standard gateway token from contracts, scripts, and docs

DZGoldman:


packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/aeERC20.sol
```diff
@@ -19,12 +19,12 @@
...
 
 import "@openzeppelin/contracts-upgradeable/drafts/ERC20PermitUpgradeable.sol";
-import "./ERC677Token.sol";
+import "./TransferAndCallToken.sol";
 
 /// @title Arbitrum extended ERC20
 /// @notice The recommended ERC20 implementation for Layer 2 tokens
-/// @dev This implements the ERC20 standard with extensions to improve UX (ERC677 & ERC2612)
-contract aeERC20 is ERC20PermitUpgradeable, ERC677Token {
+/// @dev This implements the ERC20 standard with callback hook extenstion
```
fredlacs, line 26: I don't like the idea of calling it a callback hook.
I would describe it as "This implements the ERC20 standard with transferAndCall extenstion/affordances

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/TransferAndCallToken.sol
```diff
@@ -2,11 +2,14 @@
...
-import "./IERC677.sol";
+import "./ITransferAndCall.sol";
 
-// Implementation from https://github.com/smartcontractkit/LinkToken/blob/master/contracts/v0.6/ERC677Token.sol
-
-abstract contract ERC677Token is ERC20Upgradeable, IERC677 {
+// Implementation from https://github.com/smartcontractkit/LinkToken/blob/master/contracts/v0.6/TransferAndCallToken.sol
+/**
+ * @notice based on Implementation from https://github.com/smartcontractkit/LinkToken/blob/master/contracts/v0.6/ERC677Token.sol
+ * Implementation doesn't return a bool on onTokenTransfer, and so doesn't adhere to the proposed 677 standard, thus, tho it's similar to ERC677, we don't refer to it as such.
```
fredlacs, line 10: This sentence is a bit long and confusing.
I would say something like
" The implementation doesn't return a bool on onTokenTransfer. This is similar to the proposed 677 standard, but still incompatible - thus we don't refer to it as such."

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/ITransferAndCall.sol
```diff
@@ -13,7 +13,7 @@ interface IERC677 is IERC20Upgradeable {
     event Transfer(address indexed from, address indexed to, uint256 value, bytes data);
 }
 
-interface IERC677Receiver {
+interface ITransferAndCallReceiver {
```
fredlacs, line 16: I think we should add a comment here about the expected behaviour for reverts / return value.

fredlacs (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/ITransferAndCall.sol
DZGoldman: this sufficient? not sure if there's more we want to say about reverts

fredlacs (approved):
LGTM

PR #1419, opened 2021-07-23
Refactor checkpoint DB record

joshuacolvin0:
* Add L1/L2 block numbers and L2 timestamp to c checkpoint DB record
* Refactor to allow checkpoint to be saved without saved machine


packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -900,13 +900,16 @@ void send(MachineState& m) {
...
+    if (m.context.inboxEmpty()) {
         return InboxBlocked();
     }
 
+    auto next_message = m.context.popInbox();
+
+    if (next_message.message.timestamp > m.last_inbox_timestamp) {
+        m.l1_block_number = next_message.message.block_number;
+        m.last_inbox_timestamp = next_message.message.timestamp;
+    }
```
hkalodner, line 912: Might want a separate check for the block number being greater in case one goes up and the other down

packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
```diff
@@ -971,8 +974,10 @@ BlockReason sideload(MachineState& m) {
         m.context.sideloads.pop_back();
     } else {
         if (m.context.stop_on_sideload && !m.context.first_instruction) {
+            m.l2_block_number = block_num;
             return SideloadBlocked{block_num};
         }
+        m.last_inbox_timestamp = 0;
```
hkalodner, line 980: I don't see why we reset the timestamp here

packages/arb-avm-cpp/cmachine/machine.go
```diff
@@ -149,19 +149,19 @@ func (m *Machine) String() string {
 func makeExecutionAssertion(assertion C.RawAssertion) (*protocol.ExecutionAssertion, []value.Value, uint64, error) {
 	sendsRaw := receiveByteSlice(assertion.sends)
 	logsRaw := receiveByteSlice(assertion.logs)
-	debugPrints, err := protocol.BytesArrayToVals(receiveByteSlice(assertion.debugPrints), uint64(assertion.debugPrintCount))
+	debug_prints, err := protocol.BytesArrayToVals(receiveByteSlice(assertion.debug_prints), uint64(assertion.debug_print_count))
```
hkalodner, line 152: I think `debug_prints` should still be camelCase for golang

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
```diff
@@ -103,34 +103,44 @@ struct MachineOutput {
 };
 
 struct MachineStateKeys {
```
hkalodner, line 105: I think the changes to `MachineStateKeys` would be somewhat simpler by just adding the three new fields (`l1_block_number`, `l2_block_number`, and `last_inbox_timestamp`) to `MachineOutput` instead. That change will propagate to affect some of the refactoring and I think provides a cleaner split since then a checkpoint without a machine is just a `MachineOutput`

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
```diff
@@ -67,26 +60,30 @@ void serializeMachineStateKeys(const MachineStateKeys& state_data,
...
+
+    marshal_uint256_t(state_data.static_hash, state_data_vector);
+    marshal_uint256_t(state_data.register_hash, state_data_vector);
+    marshal_uint256_t(state_data.datastack_hash, state_data_vector);
+    marshal_uint256_t(state_data.auxstack_hash, state_data_vector);
+    state_data.pc.marshal(state_data_vector);
+    state_data.err_pc.marshal(state_data_vector);
 }
 
 MachineStateKeys extractMachineStateKeys(
```
hkalodner, line 81: If this function is meant to handle both checkpoint types, I think it would be cleaner and lead to more safety to have it return a `std::variant<MachineStateKeys, MachineOutput>`

hkalodner, line 81: It'd probably be nice to have a separate serialization and deserialization function for MachineOutput since these are pretty complicated as is and that provides a nice division

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -382,8 +382,7 @@ rocksdb::Status ArbCore::reorgToMessageCountOrBefore(
             std::vector<unsigned char> checkpoint_vector(
                 checkpoint_it->value().data(),
                 checkpoint_it->value().data() + checkpoint_it->value().size());
-            auto checkpoint =
-                extractMachineStateKeys(checkpoint_vector.begin());
+            auto checkpoint = extractMachineStateKeys(checkpoint_vector);
```
hkalodner, line 385: As-is this would fail if the checkpoint wasn't filled it. With the return type change, this would nicely pop out as a place where we could just continue to the next checkpoint if we get one that's just a `MachineOutput`

hkalodner (changes requested):

packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/avm/src/machinestate/machineoperation.cpp
joshuacolvin0: I was thinking of resetting the values when there isn't any sideloads, but I don't think it makes much sense in retrospect, will remove.

packages/arb-avm-cpp/cmachine/machine.go
joshuacolvin0: addressed in later PR

packages/arb-avm-cpp/avm/include/avm/machinestate/machinestate.hpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/value/machine.cpp
joshuacolvin0: :+1:

joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -546,28 +547,37 @@ std::variant<rocksdb::Status, MachineStateKeys> ArbCore::getCheckpointUsingGas(
...
+            it->value().data(), it->value().data() + it->value().size());
+        auto variantkeys = extractMachineStateKeys(saved_value);
+
+        if (std::holds_alternative<MachineStateKeys>(variantkeys)) {
+            // Found checkpoint with machine
+            return std::get<MachineStateKeys>(variantkeys);
         }
+
+        // Checkpoint did not contain machine
+        it->Prev();
```
PlasmaPower, line 574: This needs to go forwards if `after_gas` is true

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -546,28 +547,37 @@ std::variant<rocksdb::Status, MachineStateKeys> ArbCore::getCheckpointUsingGas(
...
+                return rocksdb::Status::NotFound();
+            }
         }
-        return rocksdb::Status::NotFound();
-    }
-    if (after_gas) {
-        it->Next();
         if (!it->status().ok()) {
             return it->status();
         }
```
PlasmaPower, line 562: This check is unnecessary, as if the status is not ok, it'll exit the loop as the iterator isn't valid, and then return the status.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -546,28 +547,37 @@ std::variant<rocksdb::Status, MachineStateKeys> ArbCore::getCheckpointUsingGas(
     marshal_uint256_t(total_gas, key);
     auto key_slice = vecToSlice(key);
     it->SeekForPrev(key_slice);
-    if (!it->Valid()) {
-        if (!it->status().ok()) {
-            return it->status();
+    while (it->Valid()) {
+        if (after_gas) {
+            it->Next();
```
PlasmaPower, line 552: This should be outside the loop, and as mentioned below it should continue forwards if after_gas is true. However, I've just noticed that getCheckpointUsingGas is never called with after_gas is true, so I would just cut out this code and remove the parameter.

hkalodner (approved):
LGTM

PlasmaPower (commented):
LGTM minus some cleanup in `getCheckpointUsingGas`

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: good catch

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: gone :)

PlasmaPower (approved):
LGTM

PR #1428, opened 2021-07-27
ensure custom token can't reset L2 token address for arbCustomGateway

DZGoldman:
looks like my linter picked up a few other things


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1CustomGateway.sol
```diff
@@ -148,6 +148,10 @@ contract L1CustomGateway is L1ArbitrumExtendedGateway, ICustomGateway {
         address _creditBackAddress
     ) public payable virtual returns (uint256) {
         require(address(msg.sender).isContract(), "MUST_BE_CONTRACT");
+
+        // Can't modify previously set L2 address (can "reset" it to same address on the off chance that retryable fails on the first try)
+        if (l1ToL2Token[msg.sender] != address(0)) require(l1ToL2Token[msg.sender] == _l2Address);
```
fredlacs, line 153: I think we should create an intermediate variable in memory to ensure we don't make two reads from state
Also would be cool to add a revert message to the require statement

fredlacs (changes requested):

DZGoldman:
as per discussion, we're going to keep the current behavior but document it more clearly

PR #1430, opened 2021-07-27
Some writing improvements

dionyziz:
- Fix a couple of typos and grammar
- Change `C` variable name to `k` for stability parameter. This removes
  a variable naming conflict from the Fees section. It is also
  consistent with the Bitcoin Backbone paper.
- Split up "correctness" into "safety" and "liveness" corresponding to
  classical BFT literature.
- Fix an incorrect induction parameter from n to lgn. The induction is
  on the tree height, not on the leaves count.


DZGoldman:
@dionyziz sorry this got buried! If you're still around, if you fix the conflict (some of the typos have been fixed elsewhere, but not all) and sign the CLA we can get this merged 🙏

PR #1432, opened 2021-07-29
Properly handle configuration overrides

joshuacolvin0:
The list of configuration possibilities, each item overrides all configurations below it
* Environment variables
* JSON string supplied with `--conf.string`
* Command line options
* Local configuration file
* configuration file stored in S3
* Chain specific parameters


hkalodner (approved):
LGTM

PR #1445, opened 2021-08-01
Sequencer L1 transacting improvements

PlasmaPower:
- Uses the new validator replace by fee method, which with #1444 includes full EIP-1559 support
- Guarantees that on startup or on failover, the sequencer sequences delayed messages before incoming user transactions (note: requires `--wait-to-catch-up` to work properly, as otherwise it won't have the latest delayed messages to begin with).
- Confirms that delayed messages still exist before sequencing them, protecting against the inbox reader missing a reorg
- Publish batches when they fill up, not just on the hour


packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -508,10 +509,27 @@ func (b *SequencerBatcher) deliverDelayedMessages(chainTime inbox.ChainTime) (bo
...
+	_, isSimulatedBackend := b.client.(*ethutils.SimulatedEthClient)
+	var getDelayedAccTarget *big.Int
+	if isSimulatedBackend {
+		// The simulated backend doesn't support querying against old blocks
+		getDelayedAccTarget = chainTime.BlockNum.AsInt()
+	} else {
+		// Confirm that the message wasn't reorganized forwards to a further reorganizable block
+		getDelayedAccTarget = lastConfirmedL1Block
+	}
+	l1DelayedAcc, err := b.inboxReader.GetDelayedAccumulator(ctx, lastDelayedSeqNum, getDelayedAccTarget)
```
hkalodner, line 540: We should discuss whether adding an archive node requirement here is the only way to check this. In particular, could we use an event query instead?

packages/arb-rpc-node/batcher/sequencerBatcher.go
```diff
@@ -739,6 +751,15 @@ func (b *SequencerBatcher) publishBatch(ctx context.Context, dontPublishBlockNum
 		return false, err
 	}
 
+	if publishingAllBatchItems {
+		// Reset the pending gas estimate to gasCostBase,
+		// plus whatever was sequenced while we were building this batch.
+		atomic.AddInt64(&b.pendingBatchGasEstimateAtomic, int64(gasCostBase)-origEstimate)
+	} else {
+		// Since we didn't publish everything, only subtract gas for what we did publish.
+		atomic.AddInt64(&b.pendingBatchGasEstimateAtomic, -int64(estimatedGasCost))
```
hkalodner, line 760: In this case, doesn't the base gas cost of the next batch still need to be incorporated into the total?

hkalodner (changes requested):

packages/arb-rpc-node/batcher/sequencerBatcher.go
PlasmaPower: I believed that geth would keep the last 255 blocks in memory or so. This only queries from 12 blocks ago. Is that an issue? I can update it to use an event query if so.

packages/arb-rpc-node/batcher/sequencerBatcher.go
PlasmaPower: Ah, good call

packages/arb-rpc-node/batcher/sequencerBatcher.go
hkalodner: Ah good point, I hadn't considered that the backwards looking would be tightly bounded. We'll definitely want to verify that our geth node, alchemy, and infura can handle this, but it seems safe

hkalodner: Alchemy, infura, and default geth all look good

hkalodner (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

hkalodner (dismissed):
LGTM

joshuacolvin0 (dismissed):
LGTM

hkalodner (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

PR #1454, opened 2021-08-04
Refactor outbox to surface batch number and index

fredlacs:


PlasmaPower:
```sol
        // With a minimal path, the pair of path and proof length should always identify
        // a unique leaf. The path itself is not enough since the path length to different
        // leaves could potentially be different
        bytes32 uniqueKey = keccak256(abi.encodePacked(path, proof.length));
```
The above is taken from later in the code where outputs are recorded as spent. I don't believe paths are unique, you'd need to use this uniqueKey.

hkalodner (approved):
LGTM

PR #1457, opened 2021-08-05
Validate post upgrade init in rollup so its only callable by proxy admin

fredlacs:


packages/arb-bridge-eth/contracts/libraries/ProxyUtil.sol
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pragma solidity ^0.6.11;
+
+library ProxyUtil {
+    function getProxyAdmin() internal view returns (address admin) {
+        bytes32 slot = 0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
```
hkalodner, line 23: Would be good to have a comment citing where this random string of bytes came from 😁

hkalodner (changes requested):
LGTM

hkalodner (approved):
LGTM

PR #1458, opened 2021-08-06
Metrics

hkalodner:
Use go-ethereum metrics server throughout the node to unify metrics serving between the rpc interface and internals
Enable with `-metrics`

JSON: http://localhost:6070/debug/metrics
Prometheus: http://localhost:6070/debug/metrics/prometheus


joshuacolvin0 (approved):
LGTM

PR #1463, opened 2021-08-08
Gateway exit patch

fredlacs:


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumExtendedGateway.sol
```diff
@@ -32,6 +32,7 @@ interface ITradeableExitReceiver {
 
 abstract contract L1ArbitrumExtendedGateway is L1ArbitrumGateway {
     struct ExitData {
+        bool isExit;
```
hkalodner, line 35: This struct is stored in a mapping so we need to be careful with layout changes. I think this might map into the old structure ok because address is only 20 bytes so struct compaction may put the bool into data that was zero'd before, though I'm not sure whether solidity guarantees that the unused data is zero so we'd definitely need to be careful here

hkalodner: Another option would be to verify that no exits have been redirected to the 0 address and then move back to checking for nonzero `_newTo`

hkalodner (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumExtendedGateway.sol
hkalodner: Had an offline discussion. The plan is to fix any issues with the new data layout as part of the upgrade

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #1485, opened 2021-08-12
Isolated changes for disabling archive support

joshuacolvin0:
Add `--core.no-archive` to disable serving of archive requests (false by default to maintain existing behavior as default)

### DIFF
--- a/packages/arb-avm-cpp/cavm/carbcore.cpp
+++ b/packages/arb-avm-cpp/cavm/carbcore.cpp
@@ -560,15 +560,17 @@ CMachineResult arbCoreGetMachineForSideload(CArbCore* arbcore_ptr,
         auto machine =
             arbcore->getMachineForSideload(block_number, allow_slow_lookup);
         if (!machine.status.ok()) {
+            if (machine.status.IsNotFound() && !allow_slow_lookup) {
+                // Machine not found in memory cache and database lookup
+                // disabled
+                return {nullptr, 1};
+            }
+
             std::cerr << "Failed to load machine for sideload "
                       << machine.status.ToString() << std::endl;
             return {nullptr, 0};
         }
 
-        if (!machine.data) {
-            // Machine not found in memory cache and database lookup disabled
-            return {nullptr, 1};
-        }
         return {static_cast<void*>(machine.data.release()), 0};
     } catch (const std::exception& e) {
         std::cerr << "Exception while loading machine for sideload " << e.what()
--- a/packages/arb-avm-cpp/cavm/carbcore.cpp
+++ b/packages/arb-avm-cpp/cavm/carbcore.cpp
@@ -551,21 +551,28 @@ CMachine* arbCoreTakeMachine(CArbCore* arbcore_ptr,
         arbCore->takeExecutionCursorMachine(*executionCursor).release());
 }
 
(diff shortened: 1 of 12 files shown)


PlasmaPower (dismissed):
LGTM

packages/arb-rpc-node/txdb/txdb.go
```diff
@@ -209,9 +217,9 @@ func (db *TxDB) DeleteLogs(avmLogs []value.Value) error {
 		}
 
 		for i := oldHeight; i > reorgBlockHeight; i-- {
-			db.snapshotCache.Remove(i)
+			db.snapshotLRUCache.Remove(i)
 		}
-		db.snapshotCache.Remove(reorgBlockHeight)
+		db.snapshotLRUCache.Remove(reorgBlockHeight)
```
hkalodner, line 222: I think this is an issue if `cacheConfig.LRUSize == 0` so `db.snapshotLRUCache` is nil

hkalodner: Similar issue in other places where `snapshotLRUCache` is read. One easy solve would be to error if the size is zero

packages/arb-avm-cpp/cavm/carbcore.cpp
```diff
@@ -551,21 +551,28 @@ CMachine* arbCoreTakeMachine(CArbCore* arbcore_ptr,
...
+            arbcore->getMachineForSideload(block_number, allow_slow_lookup);
         if (!machine.status.ok()) {
             std::cerr << "Failed to load machine for sideload "
                       << machine.status.ToString() << std::endl;
-            return nullptr;
+            return {nullptr, 0};
+        }
+
+        if (!machine.data) {
+            // Machine not found in memory cache and database lookup disabled
```
hkalodner, line 569: If the machine isn't found, the status check above with fail since we return `NotFound` in that case. I think we want to special case check for `NotFound` in the case where `allow_slow_lookup` is true

hkalodner (changes requested):

packages/arb-util/configuration/configuration.go
```diff
@@ -160,6 +161,13 @@ type Node struct {
 	WS         WS         `koanf:"ws"`
 }
 
+type NodeCache struct {
+	AllowSlowLookup  bool          `koanf:"allow-slow-lookup"`
```
hkalodner, line 165: This needs to be surfaced as a command line option

hkalodner (changes requested):

packages/arb-rpc-node/txdb/txdb.go
joshuacolvin0: Added checks so snapshotLRUCache is not used if nil.  0 is valid setting when timedcache is added in more-optimizations2 branch

packages/arb-avm-cpp/cavm/carbcore.cpp
joshuacolvin0: :+1:

packages/arb-util/configuration/configuration.go
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1488, opened 2021-08-13
Add check to ensure database with new schema  is not used

joshuacolvin0:
Also ensure that node stops when arbcore main thread exits


hkalodner (approved):
LGTM

PR #1490, opened 2021-08-13
arb-ts: only use networks config values both both networks are configged 

DZGoldman:


DZGoldman:
(one more thing to push)

packages/arb-bridge-peripherals/scripts/deploy_token_bridge_l1.ts
```diff
@@ -32,7 +32,7 @@ const main = async () => {
   const l2SignerAddress = await l2Signer.getAddress()
 
   // parse needed vars
-  const inboxAddress = l1Network.tokenBridge.inbox
+  const inboxAddress = await bridge.l1Bridge.getInbox()
   if (!inboxAddress) throw new Error('Please set inbox address! INBOX_ADDRESS')
```
hkalodner, line 36: I think this check no longer makes sense? Idk if we want to replace with a check for `bridge.l1Bridge` above the `getInbox` call

hkalodner (changes requested):

packages/arb-bridge-peripherals/scripts/deploy_token_bridge_l1.ts
DZGoldman: ye we can just remove it entirely; `.l1Bridge` will always be present, and  `getInbox()` will itself throw when appropriate

hkalodner (approved):
LGTM

PR #1492, opened 2021-08-14
Implement and enable tuple lazy loading (targets more-optimizations2)

PlasmaPower:


hkalodner:
This was merged separately

PR #1497, opened 2021-08-17
Add options for easy profiling setup

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -685,6 +698,9 @@ void ArbCore::operator()() {
     size_t max_message_batch_size = 10;
     uint64_t next_rocksdb_save_timestamp = 0;
     std::filesystem::path save_rocksdb_path(coreConfig.save_rocksdb_path);
+    auto begin_time = std::chrono::steady_clock::now();
+    auto begin_message =
```
hkalodner, line 702: Looks like this is unused

packages/arb-node-core/monitor/monitor.go
```diff
@@ -46,6 +46,17 @@ type Monitor struct {
...
 func NewMonitor(dbDir string, contractFile string, coreConfig *configuration.Core) (*Monitor, error) {
+	if coreConfig.Profile.ResetAllExceptInbox {
+		err := cmachine.ResetAllExceptInbox(dbDir, contractFile)
+		if err != nil {
+			return nil, err
+		}
+
+		if coreConfig.Profile.RunUntil == 0 {
+			return nil, errors.New("database reset except inbox, nothing else to do")
+		}
```
hkalodner, line 57: If `RunUntil` is 0, doesn't the node just sync normally? If that's true this message seems incorrect

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: it is now being used.

packages/arb-node-core/monitor/monitor.go
joshuacolvin0: yup, removed

packages/arb-avm-cpp/cavm/carbstorage.h
```diff
@@ -32,7 +33,10 @@ CArbStorage* createArbStorage(const char* db_path,
                               int32_t lru_cache_size,
                               int32_t debug,
                               int32_t save_rocksdb_interval,
-                              const char* save_rocksdb_path);
+                              const char* save_rocksdb_path,
+                              int64_t profile_reorg_to,
+                              int64_t profile_run_until,
+                              int64_t profile_load_count);
```
PlasmaPower, line 39: In #1600 this is changed into a struct. Likely we'll want to merge this one first but something to be aware of.

packages/arb-avm-cpp/data_storage/src/arbstorage.cpp
```diff
@@ -90,3 +90,43 @@ std::unique_ptr<ReadTransaction> ArbStorage::makeReadTransaction() {
...
+        if (!s.ok()) {
+            std::cerr << "Error deleting columns: " << s.ToString()
+                      << std::endl;
+            return s;
+        }
+        storage.closeDb();
+    }
+    {
+        ArbStorage storage{dbpath, coreConfig};
+        auto s = storage.initialize(arbospath);
```
PlasmaPower, line 123: I know this was in the old code, but why do we initialize ArbStorage here?

packages/arb-avm-cpp/data_storage/src/arbstorage.cpp
```diff
@@ -90,3 +90,43 @@ std::unique_ptr<ReadTransaction> ArbStorage::makeReadTransaction() {
 std::unique_ptr<ReadWriteTransaction> ArbStorage::makeReadWriteTransaction() {
     return std::make_unique<ReadWriteTransaction>(datastorage);
 }
+
+rocksdb::Status resetDBExceptInbox(const std::string& dbpath,
+                                   const std::string& arbospath) {
```
PlasmaPower, line 95: It's a bit weird to me that this is a standalone function instead of being a method on ArbStorage. It's possible the architecture makes that not possible, but I think it should be possible to make this a method that you have to call before initialize, which I think would be preferable. If this involves significant refactoring to change though, leave it be, it's only here for debugging/profiling.

packages/arb-avm-cpp/cavm/carbstorage.h
joshuacolvin0: copied the struct into this PR to reduce conflicts

packages/arb-avm-cpp/data_storage/src/arbstorage.cpp
joshuacolvin0: moved reset into initialize

packages/arb-avm-cpp/data_storage/src/arbstorage.cpp
joshuacolvin0: removed

hkalodner (approved):
LGTM

PlasmaPower (approved):
LGTM

PR #1499, opened 2021-08-17
Delay L1 transacting when the gas price is high

PlasmaPower:


packages/arb-node-core/staker/staker.go
```diff
@@ -114,7 +122,59 @@ func (s *Staker) RunInBackground(ctx context.Context) chan bool {
...
+		}
+	}
+	latestBlockInfo, err := s.client.BlockInfoByNumber(ctx, nil)
+	if err != nil {
+		logger.Warn().Err(err).Msg("error getting latest block")
+		return true
+	}
+	latestBlockNum := latestBlockInfo.Number.ToInt()
+	if s.lastActCalledBlock == nil {
+		s.lastActCalledBlock = latestBlockNum
```
hkalodner, line 144: Seems like `lastActCalledBlock` should also be updated when the validator takes an action, but I don't think it does currently

hkalodner (changes requested):

packages/arb-node-core/staker/staker.go
PlasmaPower: I completely missed this, thanks

PlasmaPower: Specifically, it should be updated not only when the validator takes an action, but also when it doesn't, as the highGasBlocksBuffer is still updated. I've fixed this.

hkalodner (approved):
LGTM

PR #1501, opened 2021-08-18
Add duplicate code point ArbCore test (targets more-optimizations2)

PlasmaPower:

PR #1502, opened 2021-08-18
Fix more-optimizations2 duplicate code point issue and add test

PlasmaPower:
Combines #1501 with a fix for it based on keeping saved code point stubs in memory


hkalodner (commented):
LGTM

PR #1503, opened 2021-08-19
Metrics, Healthcheck, and Config Updates

hkalodner:
- Add proper ready check to node
- Add healthcheck to determine whether node is synced
- Add more internal metrics
- Refactor config to make updates easier


packages/arb-node-core/monitor/inboxReader.go
```diff
@@ -160,10 +180,27 @@ func (ir *InboxReader) getMessages(ctx context.Context) error {
...
 		reorgingSequencer := false
-		if ir.caughtUp {
+
+		localCount, err := ir.db.GetMessageCount()
+		if err != nil {
+			return err
+		}
+
+		// TODO: Lee is this the correct check? It seems like this check should be more
+		// often, but I'm not sure what condition should be used
```
hkalodner, line 202: Noting this for discussion

PR #1508, opened 2021-08-23
arb-upgrades package / config for arb-bridge-peripherals 

DZGoldman:


packages/arb-upgrades/hardhat.config.ts
```diff
@@ -0,0 +1,5 @@
+// use same config from bridge package
+import * as config from 'arb-bridge-eth/hardhat.config'
+import 'hardhat-deploy-ethers'
```
fredlacs, line 3: the hardhat config in`arb-bridge-eth` does `import 'hardhat-deploy'`
is there any reason we use this here but not there? I'm afraid of us running into incompatible dependency version clashes in the future

packages/arb-bridge-peripherals/hardhat.config.ts
```diff
@@ -1,3 +1,34 @@
...
+import * as config from 'arb-bridge-eth/hardhat.config'
+import { task } from 'hardhat/config'
+import { initUpgrades } from 'arb-upgrades'
+
+require('@openzeppelin/hardhat-upgrades')
+
+task('deploy-logic-one', 'deploy one logic')
+  .addParam('contract', 'contract to deploy')
+  .setAction(async (args, hre) => {
+    const { contract } = args
```
fredlacs, line 10: should we explicitly validate this input is valid? or do we get a clean type error when we try to use an incorrect input in `await deployLogic(contract)`

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+//   getStorageLayout,
+//   assertUpgradeSafe,
+//   getContractVersion,
+//   assertStorageUpgradeSafe,
+//   solcInputOutputDecoder,
+//   validate,
+//   RunValidation,
+// } from '@openzeppelin/upgrades-core'
+
+const adminSlot =
```
fredlacs, line 23: we should probably add a comment that references where this value comes from

ie https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.4.0/contracts/proxy/TransparentUpgradeableProxy.sol#L48

same for the beacon and implementation slot

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+) => {
+  const getQueuedUpdates = async (): Promise<{
+    path: string
+    data: QueuedUpdates
+  }> => {
+    const network = await hre.ethers.provider.getNetwork()
+    const path = `${rootDir}/deployments/${network.chainId}_queued-updates.json`
+    try {
+      const jsonBuff = readFileSync(path)
+      return { path, data: JSON.parse(jsonBuff.toString()) as QueuedUpdates }
```
fredlacs, line 57: I believe this is typesafe since we usually `writeFileSync(path, JSON.stringify(queuedUpdatesData))`
I'm not sure how this would behave if the QueuedUpdates json is manually changed to something incorrect

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+    path: string
+    data: CurrentDeployments
+  }> => {
+    const network = await hre.ethers.provider.getNetwork()
+    const path = `${rootDir}/deployments/${network.chainId}_current_deployment.json`
+    try {
+      const jsonBuff = readFileSync(path)
+      return {
+        path,
+        data: JSON.parse(jsonBuff.toString()) as CurrentDeployments,
```
fredlacs, line 79: as for the QueuedUpdates comment above, does this fail in a clean way if the json is manually altered into something incompatible?

fredlacs: I'm afraid of a failure mode similar to something described [here](https://basarat.gitbook.io/typescript/type-system/type-assertion#:~:text=but%20the%20hazard%20here%20is%20the%20same%2C%20if%20you%20forget%20a%20property%20the%20compiler%20will%20not%20complain.%20It%20is%20better%20if%20you%20do%20the%20following%3A)

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+
+    const { path, data: queuedUpdatesData } = await getQueuedUpdates()
+
+    for (const contractName of contractNames) {
+      console.log('Deploying new logic for ', contractName)
+
+      const contractFactory = (
+        await hre.ethers.getContractFactory(contractName)
+      ).connect(signer)
+      // TODO: ensure we're on the right layer?
```
fredlacs, line 129: we can try to `getCode` for an arb-os precompile to see if we are actually in the L2

ie arbsys at `0x0000000000000000000000000000000000000064` returns `0xfe` in L2 but `0x` in L1

fredlacs: we can infer if the contract is intended for L1 or L2 based on the first 2 characters of `ContractNames` (except for StandardArbERC20 which is always L2)

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+      const deployedContract = await newLogic.deployed()
+      const receipt = await deployedContract.deployTransaction.wait()
+
+      const newLogicData: QueuedUpdate = {
+        address: receipt.contractAddress,
+        deployTxn: receipt.transactionHash,
+        arbitrumCommitHash: currentCommit,
+        buildInfo: await getBuildInfoString(contractName),
+      }
+      queuedUpdatesData[contractName] = newLogicData
```
fredlacs, line 140: before making the deployment, should we check if there is an update already queued for the current contract?
if so, im not sure we want to override the update (if so, we should surface some warning about it first)

fredlacs (changes requested):
One thing I want to think more deeply about is how we recover in case a provider or something else fails halfway through `updateImplementations` and if that can lead us to some awkward states.

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+  const found = tree.split('\n').find(diff => diff.includes('contracts'))
+  if (found) {
+    throw new Error(
+      `You have local changes to ${found}. Commit/stash/ get rid of them`
+    )
+  }
+}
+
+export const initUpgrades = (
+  hre: HardhatRuntimeEnvironment,
```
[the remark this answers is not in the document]
DZGoldman: We use artifacts as well; I'd rather just have the whole HRE available, as the package assumes its user is using hardhat

packages/arb-bridge-peripherals/hardhat.config.ts
```diff
@@ -1,3 +1,34 @@
-import * as config from "arb-bridge-eth/hardhat.config"
-// use same config from bridge package
+import * as config from 'arb-bridge-eth/hardhat.config'
+import { task } from 'hardhat/config'
+import { initUpgrades } from 'arb-upgrades'
+
+require('@openzeppelin/hardhat-upgrades')
```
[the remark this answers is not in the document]
DZGoldman: shouldn't be here, removing

packages/arb-bridge-peripherals/deployments/1337_current_deployment.json
```diff
@@ -0,0 +1,30 @@
+{
```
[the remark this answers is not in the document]
DZGoldman: okay, we should think about how the hardhat fork will be handled then

packages/arb-upgrades/index.ts
```diff
@@ -0,0 +1,361 @@
...
+  rootDir: string
```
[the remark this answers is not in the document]
DZGoldman: __direname returns the path in the context that its invoked (i.e., in this case it would always be `/arb-upgrades`

packages/arb-upgrades/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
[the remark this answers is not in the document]
DZGoldman: hmm, I got this from `tools` so there's already an inconsistency. but probably better to use ES2017 here

packages/arb-bridge-peripherals/hardhat.config.ts
DZGoldman: getContractFactory throws a clean error if the contract name is unrecognized

packages/arb-bridge-peripherals/deployments/1337_current_deployment.json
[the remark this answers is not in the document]
fredlacs: uses the current deployment from the chain it forked?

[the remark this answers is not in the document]
DZGoldman: We would need additional affordances in the script for this (i.e., read the data for the chain with the forked chain's id, not the current chain, and don't overwrite to the forked chain's json)

PR #1509, opened 2021-08-23
Tuple lazy loading duplicate code segments fix

PlasmaPower:


hkalodner:
This code was merged separately

PR #1520, opened 2021-08-24
Update node to integrate contracts-review changes

hkalodner:


packages/arb-bridge-eth/contracts/rollup/RollupLib.sol
```diff
@@ -143,8 +147,7 @@ library RollupLib {
         pure
         returns (uint256)
     {
-        require(assertion.afterState.gasUsed > assertion.beforeState.gasUsed, "GAS_USED_UNDERFLOW");
-        return assertion.afterState.gasUsed - assertion.beforeState.gasUsed;
+        return assertion.afterState.gasUsed.sub(assertion.beforeState.gasUsed);
```
hkalodner, line 150: Changed this because a staker test was reverting @PlasmaPower. Interesting that we hd a test that made assertions with 0 gas, but changing this to use safemath and allow 0 seems fine. We should separately check for the value being nonzero if we want that

PR #1554, opened 2021-08-28
Make TransactAuth an interface

PlasmaPower:
Also hopefully fixes rbf code thinking the tx was confirmed too early with fireblocks


joshuacolvin0 (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

PR #1555, opened 2021-08-28
Fix fireblocks config issues

joshuacolvin0:
Add unit test case where fireblocks internal/external wallet maps are missing
Remove private keys from dumped configuration


PlasmaPower (dismissed):
LGTM

hkalodner (dismissed):
LGTM

hkalodner (approved):
LGTM

PR #1564, opened 2021-08-29
Blake2f support: initial

tsahee:
This has initial blake2b_support.

Notes:
* opcode, gas price, behavior taken from arbos emulator
* upgraded boost to 1.71 - it is the default on Ubuntu 20.04, and it adds nice functions to load/store from array of bytes to integer types, both little and big endian
* Test cases are taken from https://eips.ethereum.org/EIPS/eip-152

Known Missing:
* system / full tests with a contract trying to call blake2f precompile.


joshuacolvin0:
no longer needed

PR #1568, opened 2021-08-30
Use transactauth in arb-cli

joshuacolvin0:


PlasmaPower:
A couple overall things:
- I think we'll generally only want one TransactAuth. That one might be wrapping an inner TransactAuth, but that's only visible during construction, not during usage. We shouldn't be accessing the underlying L1 TransactAuth (if it even exists).
- With that in mind, I think we should return to passing around a config with the transactauth, client, and ctx in it.

PR #1577, opened 2021-08-30
Arb ts event filter; fix tryGetProof error type check; fix getOutGoingMesssageState casing update tests

DZGoldman:


hkalodner (approved):
LGTM

PR #1578, opened 2021-08-30
Add new GasRefunder contract and integrate it into SequencerInbox

PlasmaPower:


packages/arb-bridge-eth/contracts/bridge/SequencerInbox.sol
```diff
@@ -187,6 +188,44 @@ contract SequencerInbox is ISequencerInbox, Cloneable {
...
+        emit SequencerBatchDeliveredFromOrigin(
+            startNum,
+            beforeAcc,
+            messageCount,
+            afterAcc,
+            inboxAccs.length - 1
+        );
+
+        if (gasRefunder != IGasRefunder(0)) {
+            gasRefunder.onGasSpent(msg.sender, startGasLeft - gasleft());
```
fredlacs, line 225: Should we do a contract existence check before the external call?
If this accidentally calls an EOA, I believe we won't see a revert

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+    receive() external payable {
+        emit Deposited(msg.sender, msg.value);
+    }
+
+    function withdraw(address payable destination, uint256 amount) external onlyOwner {
+        destination.transfer(amount);
+        emit Withdrawn(msg.sender, destination, amount);
+    }
+
+    function onGasSpent(address payable refundee, uint256 gasUsed) external override {
```
fredlacs, line 76: should we return a boolean indicating if a refund was sent to the consumer?

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+    mapping(address => uint256) public lastContractRefund;
+    uint256 public maxRefundeeBalance;
+    uint256 public extraGasMargin;
+
+    event RefundedGasCosts(
+        address indexed refundee,
+        address indexed contractAddress,
+        uint256 gas,
+        uint256 amountPaid,
+        bool success
```
fredlacs, line 36: Since we have a spare param to index, should we add `success`?
Makes it easy for us to check if there were any failed refunds

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+            if (refundeeBalance > maxRefundeeBalanceCache) {
+                // The refundee is already above their max balance
+                emit RefundGasCostsDenied(refundee, msg.sender, gasUsed, 1);
+                return;
+            } else {
+                refundAmount = maxRefundeeBalanceCache - refundeeBalance;
+            }
+        }
+
+        bool success = refundee.send(refundAmount);
```
fredlacs, line 115:
```suggestion
        bool success = refundee.call{ value: refundAmount }("");
```
https://consensys.github.io/smart-contract-best-practices/recommendations/#dont-use-transfer-or-send

fredlacs: We could also potentially limit the gas sent in this call instead of just the evm `63/64`

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pragma solidity ^0.8.7;
+
+import "./IGasRefunder.sol";
+
+import "@openzeppelin/contracts-0.8/access/Ownable.sol";
```
fredlacs, line 23: We want to use `OwnableUpgradeable` from `https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.3.1/contracts/access/OwnableUpgradeable.sol`.

packages/arb-bridge-eth/package.json
```diff
@@ -42,6 +42,7 @@
   },
   "dependencies": {
     "@openzeppelin/contracts": "3.4.0",
+    "@openzeppelin/contracts-0.8": "npm:@openzeppelin/contracts@^4.3.1",
```
fredlacs, line 45: We should probably use the upgradeable version instead
`https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/tree/v4.3.1`

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
fredlacs: If we do so, we should set the constructor of this contract as an initializer and surface an initializer here that calls this:
https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.3.1/contracts/access/OwnableUpgradeable.sol#L28

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+        uint256 gas,
+        uint256 reason
+    );
+    event Deposited(address sender, uint256 amount);
+    event Withdrawn(address initiator, address destination, uint256 amount);
+    event ContractAllowedSet(address indexed contractAddress, bool allowed);
+    event MaxRefundeeBalanceSet(uint256 balance);
+    event ExtraGasMarginSet(uint256 balance);
+
+    function setContractAllowed(address contractAddress, bool allowed) external onlyOwner {
```
fredlacs, line 52: Could be useful to have the input as an array to enable/disable multiple contracts at once

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+        extraGasMargin = newMargin;
+        emit ExtraGasMarginSet(newMargin);
+    }
+
+    receive() external payable {
+        emit Deposited(msg.sender, msg.value);
+    }
+
+    function withdraw(address payable destination, uint256 amount) external onlyOwner {
+        destination.transfer(amount);
```
fredlacs, line 72:
```suggestion
        bool success = destination.call{ value: amount }("");
        require(success, "TRANSFER_FAIL");
```

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+        emit Withdrawn(msg.sender, destination, amount);
+    }
+
+    function onGasSpent(address payable refundee, uint256 gasUsed) external override {
+        uint256 startGasLeft = gasleft();
+        require(allowedContracts[msg.sender], "NOT_ALLOWED_CONTRACT");
```
fredlacs, line 78: If we change to have a boolean return, we probably want to return false instead of reverting here

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        if (lastContractRefund[msg.sender] == block.number) {
+            // There was already a refund this block, don't refund further
+            emit RefundGasCostsDenied(refundee, msg.sender, gasUsed, 0);
+            return;
+        }
+        lastContractRefund[msg.sender] = block.number;
+
+        // Allow for up to a 2 gwei tip
+        uint256 estGasPrice = block.basefee + 2e9;
```
fredlacs, line 88: should probably make `2e9` settable

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+        // Cache these variables and retrieve them before measuring gasleft()
+        uint256 refundeeBalance = refundee.balance;
+        uint256 maxRefundeeBalanceCache = maxRefundeeBalance;
+
+        // Add in a bit of a buffer for the tx costs not measured with gasleft
+        gasUsed += startGasLeft + extraGasMargin;
+        // Split this up into two statements so that gasleft() comes after the storage load of extraGasMargin
+        gasUsed -= gasleft();
+
+        uint256 refundAmount = estGasPrice * gasUsed;
```
fredlacs, line 102: unlikely to trigger, but should prob add an overflow check

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -0,0 +1,118 @@
...
+        if (tx.gasprice < estGasPrice) {
+            estGasPrice = tx.gasprice;
+        }
+
+        // Cache these variables and retrieve them before measuring gasleft()
+        uint256 refundeeBalance = refundee.balance;
+        uint256 maxRefundeeBalanceCache = maxRefundeeBalance;
+
+        // Add in a bit of a buffer for the tx costs not measured with gasleft
+        gasUsed += startGasLeft + extraGasMargin;
```
fredlacs, line 98: unlikely but should prob add an overflow check

packages/arb-bridge-eth/contracts/bridge/SequencerInbox.sol
fredlacs: Should we add a try catch in case this external call reverts?
Shouldn't be an issue since the `gasRefunder` is trusted code and should only hit a revert if badly configured

fredlacs (changes requested):

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
fredlacs: this solidity version does the check for us

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
fredlacs: this solidity version does the check for us

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -46,22 +49,49 @@ contract GasRefunder is IGasRefunder, Ownable {
...
     event ContractAllowedSet(address indexed contractAddress, bool allowed);
-    event MaxRefundeeBalanceSet(uint256 balance);
-    event ExtraGasMarginSet(uint256 balance);
+    // Current parameter IDs:
+    // 0: maxRefundeeBalance
+    // 1: extraGasMargin
+    // 2: calldataCost
+    // 3: maxGasTip
+    // 4: maxSingleRefund
+    event ParameterSet(uint256 indexed parameter, uint256 value);
```
fredlacs, line 60: We could use an enum here

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
```diff
@@ -27,6 +27,9 @@ contract GasRefunder is IGasRefunder, Ownable {
     mapping(address => uint256) public lastContractRefund;
     uint256 public maxRefundeeBalance;
     uint256 public extraGasMargin;
+    uint256 public calldataCost;
```
fredlacs, line 30: Do we need uint256 for each of these? Can half the storage slots by using uint128

fredlacs: Also I'd prob use a struct to get set these

fredlacs (changes requested):

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
fredlacs: no proxy will be used

packages/arb-bridge-eth/package.json
fredlacs: no proxy will be used

packages/arb-bridge-eth/contracts/validator/GasRefunder.sol
fredlacs: no proxy will be used

DZGoldman (approved):

hkalodner (approved):
LGTM

fredlacs (approved):
LGTM

PR #1584, opened 2021-08-30
docs updates

DZGoldman:


docs/L1_L2_Messages.md
```diff
@@ -89,127 +93,41 @@ If the Retryable Ticket is cancelled or expires before it is redeemed, Callvalue
...
+        address _inbox = inbox;
+        _;
+    }
+```
 
-    event LifetimeExtended(bytes32 indexed txId, uint newTimeout);
-    event Redeemed(bytes32 indexed txId);
-    event Canceled(bytes32 indexed txId);
-}
+_Of note: the (highly) recommended convenience methods `Inbox.depositEth` and `Inbox.createRetryableTicket` handle subtracting the offset for the credit back address and beneficiary address, and using the these reverse-aliased values as the inputs for the create retryable tickets. This way, the beneficiary and credit back addresses as received on L2 will be equal to those provided as the L1 params._
```
tsahee, line 122: I'm not sure "will be equal" is right (as they are different addresses).
will be equivalent?

docs/L1_L2_Messages.md
```diff
@@ -89,127 +93,41 @@ If the Retryable Ticket is cancelled or expires before it is redeemed, Callvalue
...
 
+#### Directly Redeeming / Cancelling Retryables
 
-```
+In the normal, happy case, a retryable ticket is automatically redeemed and executed when it arrives on L2; in this case only a single user-action (publishing the transaction on L1 that creates a retryable ticket) is required. If the transaction is not auto-redeemed (either because no L2 gas was provided or because the L2 retryable ticket reverts for whatever reason), the transaction can now either be redeemed via a signed L2 transaction (probably desireable) or cancelled.
```
tsahee, line 126: Too many parentheses. Perhaps something like this:

In the normal, happy case, a retryable ticket is automatically redeemed and executed when it arrives on L2; in this case no additional user-actions will be required. If an error occurs, e.g. if no L2 gas was provided, the L2 retryable ticket will not be redeemed automatically. In that case, the transaction can later either be redeemed via a signed L2 transaction or cancelled.

docs/Solidity_Support.md
```diff
@@ -17,6 +17,7 @@ Although Arbitrum supports Solidity code, there are differences in the effects o
 - `block.gaslimit` returns the block's ArbGas limit
 - `gasleft` returns the amount of ArbGas remaining
 - `block.number` on a non-Sequencer Arbitrum chain, returns the L1 block number at which the transaction was submitted to the inbox; on a Sequencer Arbitrum chain, returns an "estimate" of the L1 block number at which the Sequencer received the transaction (see [Time in Arbitrum](Time_in_Arbitrum.md))
+- `msg.sender` works the same way it does on Ethereum for normal L2-to-L2 transactions; for L1-to-L2 "retryable ticket" transactions, it will return the address alias of the L1 contract that triggered the message. See [retryable ticket address aliasing](L1_L2_Messages.md#address-aliasing) for more.
```
tsahee, line 20: "the L2 address alias of"?

tsahee (commented):
Do note I am an ESL and missing context.. feel free to ignore anything you don't agree with :)

DZGoldman:
>Do note I am an ESL 
well your english seems to be better than mine, so

fredlacs (approved):
LGTM

PR #1586, opened 2021-08-30
Fix eth_getStorageAt

hkalodner:


joshuacolvin0 (approved):
LGTM

happypsyduck:
In relation to the issue #1607, eth_getStorageAt doesn't accept "0x1" which is ETH-spec but accepts "0x01", which is not ETH-spec. This issue is preventing common development tools (Hardhat and Ganache) from building Arbitrum One forks.

happypsyduck:
See here for equivalent implementation on ETH:
https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L822

PR #1594, opened 2021-09-01
bump maxGas estimate for token deposits + Address alias arb-ts helpers

DZGoldman:


packages/arb-ts/src/lib/bridge.ts
```diff
@@ -288,20 +291,24 @@ export class Bridge {
...
-          0,
-          0,
-          depositCalldata
-        )
-      )[0]
+      BridgeHelper.percentIncrease(
+        (
+          await nodeInterface.estimateRetryableTicket(
+            expectedL1GatewayAddress,
+            ethers.utils.parseEther('0.05').add(estimateGasCallValue),
```
hkalodner, line 298: .05 eth is a pretty large amount, can we make change this to be configurable as well and maybe drop the default if it's safe? We can make it configurable now and drop it separately then

hkalodner (changes requested):

packages/arb-ts/src/lib/bridge.ts
hkalodner: Nm just haven't had coffee 😁

hkalodner (approved):
LGTM

PR #1608, opened 2021-09-03
synchronize outbox semantics for params / events to use "merklePath" 

DZGoldman:


packages/arb-bridge-eth/contracts/bridge/Outbox.sol
```diff
@@ -134,9 +134,9 @@ contract Outbox is IOutbox, Cloneable {
      * @notice Executes a messages in an Outbox entry.
      * @dev Reverts if dispute period hasn't expired, since the outbox entry
      * is only created once the rollup confirms the respective assertion.
-     * @param batchNum Index of OutboxEntry in outboxEntries array
+     * @param batchNum Number of target OutboxEntry
```
fredlacs, line 137: Isn't this actually the `Index of OutboxEntry in outboxEntries array` as the previous doc?
I don't think `Number of target OutboxEntry` is clearer.

Maybe we can call it the identifier of the batch?

fredlacs (dismissed):
LGTM just not sure about one natspec change

packages/arb-bridge-eth/contracts/bridge/Outbox.sol
DZGoldman: yeah not sure why i did this tbh

fredlacs (approved):
LGTM

fredlacs:
> (I force-pushed b/c it was the easiest way to handle the old abis commit don't be mad)

I am mad, but just a bit.

PR #1610, opened 2021-09-05
Solidity testing and refactor

hkalodner:


joshuacolvin0 (approved):
LGTM

packages/arb-bridge-eth/contracts/bridge/Bridge.sol
```diff
@@ -84,7 +103,7 @@ contract Bridge is OwnableUpgradeable, IBridge {
         address destAddr,
         uint256 amount,
         bytes calldata data
-    ) external override returns (bool success, bytes memory returnData) {
+    ) external virtual override returns (bool success, bytes memory returnData) {
```
DZGoldman, line 106: Is this necessary? looks like it isn't overridden by BridgeMock (or is it "preemtive" for future testing?)

DZGoldman (approved):

packages/arb-bridge-eth/contracts/bridge/Bridge.sol
hkalodner: Ah, I think this was an accident. Great catch

DZGoldman (approved):
lgtm

PR #1613, opened 2021-09-07
Output database info on startup

joshuacolvin0:
Also add option to print out files and directories used


packages/arb-util/configuration/configuration.go
```diff
@@ -788,6 +795,15 @@ func endCommonParse(k *koanf.Koanf) (*Config, *Wallet, error) {
 	wallet := out.Wallet
 	out.Wallet = Wallet{}
 
+	if out.FileInfo {
+		fmt.Printf("Database:         %s\n", out.Persistent.DatabasePath)
+		fmt.Printf("Database Backup:  %s\n", out.Core.SaveRocksdbPath)
+		fmt.Printf("Machine:          %s\n", out.Rollup.Machine.Filename)
+		fmt.Printf("Wallet DIrectory: %s\n", wallet.Local.Pathname)
+
+		os.Exit(0)
```
PlasmaPower, line 804: Should we just always print this info instead of having a flag to print it?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -448,6 +448,14 @@ rocksdb::Status ArbCore::reorgToMessageCountOrBefore(
         std::get<std::unique_ptr<MachineThread>>(std::move(setup));
     auto& output = new_machine->machine_state.output;
 
+    if (use_latest) {
+        std::cerr << "Loaded valid checkpoint, gas: " << output.arb_gas_used
+                  << ", messages: " << output.fully_processed_inbox.count
+                  << ", steps: " << output.total_steps
+                  << ", logs: " << output.log_count << ", sends"
+                  << output.send_count << "\n";
```
PlasmaPower, line 456: Somewhat unrelated, but we might want to add this info to the "Last checkpoint gas used" log statement.

PlasmaPower (commented):
Generally LGTM but maybe some small tweaks would improve it

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Yes.  I already added additional info in more-optimizations branch, will add to that branch to reduce conflicts.

packages/arb-util/configuration/configuration.go
joshuacolvin0: Updated it to print in plaintext and exit if flag is given, and do json log if log is not given.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -486,8 +486,9 @@ rocksdb::Status ArbCore::reorgToMessageCountOrBefore(
         std::cerr << "Loaded valid checkpoint, gas: " << output.arb_gas_used
                   << ", messages: " << output.fully_processed_inbox.count
                   << ", steps: " << output.total_steps
-                  << ", logs: " << output.log_count << ", sends"
-                  << output.send_count << "\n";
+                  << ", logs: " << output.log_count
+                  << ", sends: " << output.send_count
+                  << ", gas: " << output.arb_gas_used << "\n";
```
PlasmaPower, line 491: This seems to double print gas used (it's printed at the start too).

packages/arb-util/configuration/configuration.go
```diff
@@ -810,9 +814,17 @@ func endCommonParse(k *koanf.Koanf) (*Config, *Wallet, error) {
...
 
 		os.Exit(0)
+	} else {
+		logger.
+			Info().
+			Str("Database", out.Persistent.DatabasePath).
+			Str("Database Backup", out.Core.SaveRocksdbPath).
+			Str("Machine", out.Rollup.Machine.Filename).
+			Str("Wallet Directory", wallet.Local.Pathname).
+			Msg("downloading machine")
```
PlasmaPower, line 827: Is "downloading machine" an accurate summary here? Does this only get called on first boot?

PlasmaPower (commented):
A couple other minor things

packages/arb-util/configuration/configuration.go
joshuacolvin0: cut and paste error

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: fixed

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1139,6 +1210,26 @@ uint256_t ArbCore::maxCheckpointGas() {
     }
 }
 
+std::variant<rocksdb::Status, MachineStateKeys> ArbCore::getMaxCheckpoint(
```
hkalodner, line 1213: Maybe just return the variant and then use the function we have to get a `MachineOutput` out of the variant since both member types have one

packages/arb-node-core/cmd/arb-validator/arb-validator.go
```diff
@@ -95,6 +96,17 @@ func startup() error {
...
+		storage, err := cmachine.NewArbStorage(config.GetValidatorDatabasePath(), &config.Core)
+		if err != nil {
+			return err
+		}
+		storage.PrintDatabaseMetadata()
+		storage.CloseArbStorage()
+
+		return nil
+	}
+
```
hkalodner, line 109: Maybe add this to the `cmdhelp` to dedup?

hkalodner (changes requested):

packages/arb-avm-cpp/cmachine/storage.go
```diff
@@ -119,6 +119,10 @@ func NewArbStorage(dbPath string, coreConfig *configuration.Core) (*ArbStorage,
 	return returnVal, nil
 }
 
+func (s *ArbStorage) PrintDatabaseMetadata() {
+	C.printDatabaseMetadata(s.c)
```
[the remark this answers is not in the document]
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-node-core/cmd/arb-validator/arb-validator.go
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -226,9 +226,14 @@ void ArbCore::printDatabaseMetadata() {
...
                   << "\n";
     } else {
-        printMachineOutputInfo(
-            "last checkpoint",
-            std::get<MachineStateKeys>(checkpoint_result).output);
+        auto output =
+            getMachineOutput(std::get<CheckpointVariant>(checkpoint_result));
+        if (std::holds_alternative<rocksdb::Status>(checkpoint_result)) {
+            std::cout << std::get<rocksdb::Status>(checkpoint_result).ToString()
+                      << "\n";
```
hkalodner, line 233: We already checked whether `checkpoint_result` was a status above

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -999,6 +1011,17 @@ ArbCore::loadLastMatchingMachine(
     return rocksdb::Status::NotFound();
 }
 
+std::unique_ptr<MachineThread> ArbCore::getMachineThreadFromSimpleCheck(
```
hkalodner, line 1014: I think the code would actually be more readable with this inlined since it looks like it's only used in one place

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -159,6 +157,81 @@ ValueLoader ArbCore::makeValueLoader() const {
...
 }
 
+void ArbCore::printDatabaseMetadata() {
+    ReadTransaction tx(data_storage);
+    auto schema_version_result = schemaVersion(tx);
+    std::cout << "schema version:                   ";
+    if (!schema_version_result.status.ok()) {
+        std::cout << schema_version_result.status.ToString() << "\n";
+    } else {
+        std::cout << schema_version_result.data << "\n";
```
PlasmaPower, line 167: Maybe a templated helper function to print out a generic DbResult could be helpful? The machine output would still be manual but everything else could be collapsed a lot, and inlined without assigning to a variable each time even.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -896,6 +970,14 @@ ArbCore::loadLastMatchingMachine(
     ReadWriteTransaction& tx,
     std::unique_ptr<rocksdb::Iterator>& checkpoint_it,
     ValueCache& value_cache) {
+    // First check if last saved machine matches
+    auto mach = combined_machine_cache.checkSimpleMatching(check_output);
+    if (mach.machine != nullptr) {
+        // Found machine in cache
+        return std::make_unique<MachineThread>(mach.machine->machine_state);
```
PlasmaPower, line 977: Do we have a guarantee that this is past the last checkpoint?

packages/arb-avm-cpp/data_storage/src/readtransaction.cpp
```diff
@@ -36,6 +36,7 @@ rocksdb::Status ReadTransaction::createRocksdbCheckpoint(
     }
 
     status = checkpoint->CreateCheckpoint(checkpoint_dir);
+    delete checkpoint;
```
PlasmaPower, line 39: Ooh nice catch :D

PlasmaPower (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Every time machine stops, one of two things happens:
1. machine added to basic cache, and last*_cache is  cleared if new machine is newer
2. last_cache is updated with new machine

Only after cache is updated will the decision to create new checkpoints is done.

If reorg happens, out of date last*_cache are cleared.

Given all of that, checkSimpleMatching is guaranteed to always b at or past the last checkpoint, or empty.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -51,6 +51,16 @@ constexpr auto log_inserted_key = std::array<char, 1>{-60};
 constexpr auto send_inserted_key = std::array<char, 1>{-62};
 constexpr auto schema_version_key = std::array<char, 1>{-64};
 constexpr auto logscursor_current_prefix = std::array<char, 1>{-120};
+
+template <class T>
+void printResult(const char* msg, ValueResult<T> result) {
```
hkalodner, line 56: `const ValueResult<T>& result` to save the copy

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Added extra check to only use last cache if it is more up to date than database

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PlasmaPower (approved):
LGTM

PR #1615, opened 2021-09-08
fix handle upper_bound return value

zhiqiangxu:
upper_bound returns map::end when not found


PlasmaPower:
This code is correct, in the case where it returns the end, we want to go back one to the element before the end. Since we're looking for the element before the found one, the only invalid case is when the first element is returned and there's no previous element (and if the map is empty .begin() == .end()).

PR #1616, opened 2021-09-08
Fix replace by fee now that auth doesn't have a gas price

PlasmaPower:
This check isn't necessary, it's just a fail fast mechanism. There are better checks about the gas price increase later in the method.


hkalodner (approved):
LGTM

PR #1605, opened 2021-09-03
Update deps

hkalodner:
- Update NPM dependencies
- Fix typescript compilation

PR #1619, opened 2021-09-09
Fix arbts type issues

DZGoldman:
some minor regressions from https://github.com/OffchainLabs/arbitrum/pull/1605; unlovely that these don't get caught during compile time :/


hkalodner (approved):
LGTM

PR #1638, opened 2021-09-13
handle dead logo uris / check additional source for token list gen

DZGoldman:


packages/arb-ts/token_lists/generate_json.ts
```diff
@@ -25,21 +25,72 @@ import { TokenInfo, TokenList, schema } from '@uniswap/token-lists'
...
+const getL1Name = (ERC20: ERC20) => {
+  switch (ERC20.address) {
+    case '0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2':
+      return 'Maker'
+    case '0x431ad2ff6a9C365805eBaD47Ee021148d6f7DBe0':
+      return 'dForce'
+    default:
+      return ERC20.name()
+  }
+}
```
fredlacs, line 74: I think we should actually parse these as bytes32 strings instead of hardcoding specific addresses.
https://github.com/ethers-io/ethers.js/blob/ce8f1e4015c0f27bf178238770b1325136e3351a/packages/strings/src.ts/bytes32.ts#L21
```suggestion
const getL1Name = async (ERC20: ERC20) => {
   const name = await ERC20.name()
   if(name.length == 66) return ethers.utils.parseBytes32String(name)
   else return name
  }
}
```
same for the symbol

fredlacs: also instead of `name.length == 66` we could `arrayify(name).length == 32` to express it in bytes

packages/arb-ts/token_lists/generate_json.ts
```diff
@@ -63,12 +114,13 @@ const gen = async () => {
     '0x4Dbd4fc535Ac27206064B68FfCf827b0A60BAB3f',
     '0xEDA6eFE5556e134Ef52f2F858aa1e81c84CDA84b',
     '0xe54942077Df7b8EEf8D4e6bCe2f7B58B0082b0cd',
+    '0x99D8a9C45b2ecA8864373A26D1459e3Dff1e17F3',
```
fredlacs, line 117: Should we have a personalized message for this one poiting to anyswap's website?
Could open an issue and add this as a feature later

packages/arb-ts/token_lists/generate_json.ts
```diff
@@ -106,29 +155,25 @@ const gen = async () => {
...
-        : await l1TokenContract.name()
-    const l1Symbol =
-      l1Address === '0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2'
-        ? 'MKR'
-        : await l1TokenContract.symbol()
+    const l1Name = await getL1Name(l1TokenContract)
+    const l1Symbol = await getL1Symbol(l1TokenContract)
     const l1Decimals = await l1TokenContract.decimals()
-
     const l2Symbol =
```
fredlacs, line 173: if we refactor the getSymbol logic to use `parseBytes32String`, then it should work here too

fredlacs (changes requested):

PR #1640, opened 2021-09-14
Fix broadcaster receive data

joshuacolvin0:
Broadcastclient had previously been modified to reliably receive data.  The broadcast server data didn't have these changes.  The server doesn't receive any data from client, but under heavy traffic some corner cases can be reached, so better to fix the receive code fully.


PlasmaPower (approved):
LGTM

PR #1642, opened 2021-09-14
Use actual queue processing for the sequencer txQueue

PlasmaPower:
This looks like a huge diff but most of it is moving identical code from SendTransaction to consumeQueue. The main change to be aware of within that move is that resultQueue now tracks transactions which haven't been sent results yet, so every time we send a result we pop an item from the queue.


hkalodner:
Superseded by https://github.com/OffchainLabs/arbitrum/pull/1648

PR #1662, opened 2021-09-20
add inbox address to the Rinkeby arb-ts networks

mahsamoosavi:


packages/arb-ts/src/lib/networks.ts
```diff
@@ -30,6 +30,7 @@ export interface Network {
   confirmPeriodBlocks?: number
   blockTime?: number //seconds
   rpcURL: string
+  inbox: string
```
fredlacs, line 33: I believe this is more appropriate in the `tokenBridge: TokenBridge` object.
Even better could be creating a new `ethBridge` object inside Network.
We can then include addresses for inbox/outbox/rollup/sequencer inbox etc

fredlacs: The type would be `ethBridge: EthBridge | undefined` since there are no ethBridge contracts deployed when in a L2 network

packages/arb-ts/src/lib/networks.ts
```diff
@@ -125,6 +126,7 @@ export const networks: Networks = {
     confirmPeriodBlocks: 6545, // TODO
     blockTime: 15,
     rpcURL: process.env['RINKEBY_RPC'] as string,
+    inbox: '0x578BAde599406A8fE3d24Fd7f7211c0911F5B29e',
```
fredlacs, line 129: Lets take this opportunity to also add in the values for mainnet!

fredlacs (changes requested):

fredlacs:
closed in favor of https://github.com/OffchainLabs/arbitrum/pull/1664

PR #1664, opened 2021-09-20
Add ethbridge networks

mahsamoosavi:
Changes include:

- ethbridge object with addresses for inbox/outbox/rollup/sequencer inbox is added to the Network   
- addresses are set for all the networks (including mainnet)


fredlacs (approved):
LGTM

PR #1668, opened 2021-09-22
Arbts multicall

fredlacs:


DZGoldman (approved):
love it, LGTM pending conflicts

fredlacs:
This contains a breaking change to the `init` function so we should prob do a major version bump?

packages/arb-ts/src/lib/bridge.ts
```diff
@@ -850,4 +858,40 @@ export class Bridge {
       this.l2Provider
     )
   }
+
+  public async getTokenBalanceBatch(
+    userAddr: string,
+    tokenAddrs: Array<string>,
+    targetNetwork: 'L1' | 'L2'
+  ) {
```
hkalodner, line 866: Can we add typing to the return value here

hkalodner (changes requested):

packages/arb-ts/src/lib/bridge.ts
fredlacs: This was added in the follow up PR
https://github.com/OffchainLabs/arbitrum/pull/1673

fredlacs: https://github.com/OffchainLabs/arbitrum/pull/1673/commits/62248797b99b70552b0a24f93bc9bb4372140a48

hkalodner (approved):
LGTM

PR #1671, opened 2021-09-23
Upgrade 45

RachelFranks:
Adds support for ArbOS upgrade 45 by
- Adding expected estimation field to `TestArbOSFees`


hkalodner (approved):
LGTM

PR #1673, opened 2021-09-24
Refactor bridge

fredlacs:


packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -89,11 +82,25 @@ export class L1Bridge {
...
-      await this.getAndUpdateL1TokenData(l1Address)
+  public async setSigner(newSigner: Signer) {
+    const newL1Provider = newSigner.provider
+    if (newL1Provider === undefined) {
+      throw new Error('Signer must be connected to an Ethereum provider')
     }
-    return this.l1Tokens
+    // check chainId to ensure its still in the same network.
+    const prevNetwork = await this.l1Provider.getNetwork()
+    const newNetwork = await newL1Provider.getNetwork()
```
hkalodner, line 92: Let's pipeline these promises

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -102,82 +109,60 @@ export class L1Bridge {
...
-        // If this will throw if not an ERC20, which is what we *want*.
-        const ethERC20TokenContract = await ERC20__factory.connect(
-          erc20L1Address,
-          this.l1Signer
-        )
-        const [balance] = await ethERC20TokenContract.functions.balanceOf(
-          walletAddress
-        )
+    if (!this.contractExists(erc20L1Address))
+      throw new Error(`No ERC20 at ${erc20L1Address} `)
```
hkalodner, line 122: Re: the comment below before `ERC20__factory.connect`, is this necessary?

hkalodner: Actually the balanceOf call will definitely throw

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -102,82 +109,60 @@ export class L1Bridge {
...
-        this.l1Signer
-      )
-      const [balance] = await ethERC20TokenContract.functions.balanceOf(
-        walletAddress
-      )
-      tokenData.ERC20.balance = balance
+    const decimals = await ethERC20TokenContract.functions
+      .decimals()
+      .then(([res]) => res)
+      .catch(() => 18)
```
hkalodner, line 146: Would be good to pipeline all of these calls, creating promises and then `await`ing

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -102,82 +109,60 @@ export class L1Bridge {
...
-        tokenData.ERC20.allowed = allowance.gte(MIN_APPROVAL.div(2))
-      }
+    const name = await ethERC20TokenContract.functions
+      .name()
+      .then(([res]) => res)
+      .catch(() => symbol + '_Token')
+    const allowanceLimit = BigNumber.from(
+      '0xffffffffffffffffffffffff'
+    ) /** for ERC20s that cap approve at 96 bits  */
+    const allowed = await allowance.gte(allowanceLimit.div(2))
```
hkalodner, line 155: Doesn't seem like this should need an await?

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -84,6 +78,29 @@ export class L2Bridge {
...
   }
 
+  public async setSigner(newSigner: Signer) {
+    const newL2Provider = newSigner.provider
+    if (newL2Provider === undefined) {
+      throw new Error('Signer must be connected to an (Arbitrum) provider')
+    }
+    // check chainId to ensure its still in the same network.
+    const prevNetwork = await this.l2Provider.getNetwork()
+    const newNetwork = await newL2Provider.getNetwork()
```
hkalodner, line 88: Pipeline processing of these

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -118,75 +135,43 @@ export class L2Bridge {
...
-    for (const l1Address in this.l2Tokens) {
-      const l2Address = this.l2Tokens[l1Address]?.ERC20?.contract.address
-      if (l2Address) {
-        await this.getAndUpdateL2TokenData(l1Address, l2Address)
-      }
-    }
-    return this.l2Tokens
+  public async contractExists(contractAddress: string): Promise<boolean> {
+    const contractCode = await this.l2Provider.getCode(contractAddress)
+    return !(contractCode.length > 2)
```
hkalodner, line 140: I might be missing a JS thing here, but doesn't this return true if length is less than or equal to 0?

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -118,75 +135,43 @@ export class L2Bridge {
...
-    } else {
-      const arbERC20TokenContract = await StandardArbERC20__factory.connect(
-        l2ERC20Address,
-        this.l2Signer
-      )
-      const [balance] = await arbERC20TokenContract.functions.balanceOf(
-        walletAddress
+    if (!this.contractExists(l2ERC20Address))
+      throw new Error(
+        `Corresponding ArbERC20 for ${erc20L1Address} not yet deployed (would be at ${l2ERC20Address})`
```
hkalodner, line 151: Possibly unnecessary check

hkalodner: Though maybe we want it for the clearer error messages?

hkalodner: If that's the case we could use a try catch thing where we only check for existence if a call reverts

hkalodner (changes requested):

packages/arb-ts/src/lib/l1Bridge.ts
fredlacs: yea, not necessary, only if we wanted to surface the reason for a throw more accurately

packages/arb-ts/src/lib/l1Bridge.ts
fredlacs: Pipelined (also took the opportunity to pipeline in Bridge.init)

packages/arb-ts/src/lib/l1Bridge.ts
fredlacs: how about we multicall them

packages/arb-ts/src/lib/l1Bridge.ts
fredlacs: oops

packages/arb-ts/src/lib/l2Bridge.ts
fredlacs: this function was removed since its not needed anymore

packages/arb-ts/src/lib/l2Bridge.ts
fredlacs: Yea the idea was for clearer error messages, but I don't think its worth it here

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -30,14 +30,26 @@ import { Inbox__factory } from './abi/factories/Inbox__factory'
...
 //TODO handle address update / lowercase
 
+// typing magic from https://stackoverflow.com/a/57364353
+export type Await<T> = T extends {
+  then(onfulfilled?: (value: infer U) => unknown): unknown
+}
+  ? U
+  : T
+
 export interface L1TokenData {
```
DZGoldman, line 53: I think along with this refactor, we can/should just turn this interface into 

```ts
export interface L1TokenData {
    contract: ERC20
    balance: BigNumber
    allowed: boolean
    symbol: string
    decimals: number
    name: string
}

File: packages/arb-ts/src/lib/l2Bridge.ts [hunk 9]
```diff
@@ -36,31 +36,27 @@ import {
   ARB_RETRYABLE_TX_ADDRESS,
 } from './precompile_addresses'
 import { Network } from './networks'
+import { ArbMulticall2__factory } from './abi/factories/ArbMulticall2__factory'
+import { MulticallFunctionInput, BridgeHelper } from './bridge_helpers'
 
 export interface L2TokenData {
```
DZGoldman, line 42: Same as above comment r.e. L1TokenData: should probably just get rid of the ERC/CUSTOM layer of the interface

packages/arb-ts/src/lib/l2Bridge.ts
```diff
@@ -118,75 +139,34 @@ export class L2Bridge {
...
-      if (l2Address) {
-        await this.getAndUpdateL2TokenData(l1Address, l2Address)
-      }
-    }
-    return this.l2Tokens
-  }
-
-  public async getAndUpdateL2TokenData(
+  public async getL2TokenData(
     erc20L1Address: string,
```
DZGoldman, line 143: this param is currently usused

DZGoldman: It might (indeed) make more sense to use the L2 address as the (only) input for this method (but currently elsewhere, i.e., the tests, it's still assumed to be the L1 address)

PR #1682, opened 2021-09-28
Refactor function parameter types in deposit

fredlacs:


packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -264,27 +275,34 @@ export class L1Bridge {
...
+    }: DepositParams,
     overrides: PayableOverrides = {}
   ): Promise<ContractTransaction> {
     const destination = destinationAddress || (await this.getWalletAddress())
     const data = defaultAbiCoder.encode(
       ['uint256', 'bytes'],
       [maxSubmissionCost, '0x']
     )
+
+    if (overrides.value)
```
DZGoldman, line 295: should we also throw if the value is BigNumber.Zero?

DZGoldman (approved):
ltgm

PR #1695, opened 2021-10-01
expose estimateGasDeposit method in arb-ts

DZGoldman:


packages/arb-ts/src/lib/l1Bridge.ts
```diff
@@ -263,7 +292,14 @@ export class L1Bridge {
     )
   }
 
-  public async deposit(
+  public async estimateGasDeposit(
```
fredlacs, line 295: We could also include this in `public async getDepositTxParams(` instead

PR #1696, opened 2021-10-01
Fix code locking in new DB

PlasmaPower:
- Remove unnecessary locking in RunningCode when only the parent needs accessed
- Invert lock order of commitCode / fillInCode to lock RunningCode before CoreCode, which makes sense given that the former contains the latter


joshuacolvin0 (approved):
LGTM

hkalodner (approved):
LGTM

PR #1701, opened 2021-10-04
Ethcall overrides

tsahee:
Missing Tests:
* Trying to call arbos_test functions from anywhere not zero and failing.


packages/arb-evm/arbos/arbos.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 var ARB_SYS_ADDRESS = ethcommon.HexToAddress("0x0000000000000000000000000000000000000064")
+var ARB_TEST_ADDRESS = ethcommon.HexToAddress("0x0000000000000000000000000000000000000069")
```
hkalodner, line 12: Tiny thing, but these are otherwise sorted

packages/arb-rpc-node/snapshot/snapshot.go
```diff
@@ -371,6 +371,46 @@ func (s *Snapshot) GetStorageAt(account common.Address, index *big.Int) (*big.In
 	return arbos.ParseGetStorageAtResult(res.ReturnData)
 }
 
+func (s *Snapshot) SetNonce(account common.Address, nonce uint64) error {
+	_, err := s.basicCall(arbos.SetNonceData(account, nonce), common.NewAddressFromEth(arbos.ARB_TEST_ADDRESS))
```
hkalodner, line 375: I think we want the `checkValidResult` check here like in `GetBalance` to catch if the method fails for some reason. Especially with that, there's enough code here that I think a common helper method that takes just `data []byte` would be good

packages/arb-evm/arbos/arbostest.go
```diff
@@ -0,0 +1,72 @@
...
+	return append(setCodeABI.ID, args...)
+}
+
+func SetStorageData(address common.Address, storage map[common.Hash]common.Hash) []byte {
+	args, err := setStorageABI.Inputs.Pack(address, storage)
+	if err != nil {
+		panic(err)
+	}
+	return append(setStorageABI.ID, args...)
+}
```
hkalodner, line 64: This function doesn't appear to map to anything in the `ArbosTest` file. I'm guessing this should go to `setState` which looks like it expects a byte slice formatted as a list of 32 byte keys and values in https://github.com/OffchainLabs/arb-os/blob/143a45ec68b10cdcf3b7c3dc3c5170b9a31bc0b3/stdlib/storageMap.mini#L262

hkalodner (changes requested):

packages/arb-rpc-node/dev/ethcall_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/offchainlabs/arbitrum/packages/arb-evm/arboscontracts"
+	"github.com/offchainlabs/arbitrum/packages/arb-rpc-node/arbostestcontracts"
+	"github.com/offchainlabs/arbitrum/packages/arb-rpc-node/web3"
+	"github.com/offchainlabs/arbitrum/packages/arb-util/common"
+	"github.com/offchainlabs/arbitrum/packages/arb-util/protocol"
+	"github.com/offchainlabs/arbitrum/packages/arb-util/test"
+)
+
+func expectHex(t *testing.T, result []byte, resultErr error, expected string) {
+	test.FailIfError(t, resultErr)
```
hkalodner, line 26: If you add `t.Helper()` at the top of this function, test failures will give you the line number of the caller if this fails rather than the line in `expectHex`

packages/arb-rpc-node/dev/ethcall_test.go
```diff
@@ -0,0 +1,180 @@
...
+	if ethcommon.BytesToHash(result) != ethcommon.HexToHash(expected) {
+		t.Fatal("Unexpected ", expected, "Got", ethcommon.BytesToHash(result))
+	}
```
hkalodner, line 29: I think this only checks the first 32 bytes. It'd be safer to either do a `bytes.Equal` on two byte slices or do a length check on the inputs to make sure they're the expected length

packages/arb-rpc-node/snapshot/snapshot.go
```diff
@@ -372,42 +389,57 @@ func (s *Snapshot) GetStorageAt(account common.Address, index *big.Int) (*big.In
...
-	_, err := s.basicCall(arbos.SetBalanceData(account, ballance), common.NewAddressFromEth(arbos.ARB_TEST_ADDRESS))
+	res, err := s.basicAddMessage(arbos.SetBalanceData(account, ballance), common.NewAddressFromEth(arbos.ARB_TEST_ADDRESS))
 	if err != nil {
 		return err
 	}
+	if err := checkValidResult(res); err != nil {
+		return err
+	}
 	return nil
 }
```
hkalodner: I think these new methods could be simpler by defining 
```golang
func (s *Snapshot) addArbTestMessage(data []byte) error {
	res, err := s.basicAddMessage(data), common.NewAddressFromEth(arbos.ARB_TEST_ADDRESS))
	if err != nil {
		return err
	}
	if err := checkValidResult(res); err != nil {
		return err
	}
	return nil
}
```
and then this just becomes
```go
func (s *Snapshot) SetBalance(account common.Address, balance *big.Int) error {
    return addArbTestMessage(arbos.SetBalanceData(account, balance))
}
```

Also, typo in `balance` 🙂

PR #1713, opened 2021-10-07
chore: add custom gateway lazy mint on withdrawal

fredlacs:


packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/ICustomToken.sol
```diff
@@ -23,6 +23,12 @@ interface ArbitrumEnabledToken {
     function isArbitrumEnabled() external view returns (uint8);
 }
 
+interface L1MintableToken {
```
DZGoldman, line 26: should we have this inherit from ICustomToken

gzeoneth:
should we keep accounting on the bridgeminted token so that we burn them when they are bridged back?

PR #1725, opened 2021-10-13
Fix MachineThread::nextAssertion when running synchronously

PlasmaPower:
If running the machine synchronously, `machine_thread` is null, and therefore shouldn't be joined.


hkalodner (approved):
LGTM

PR #1729, opened 2021-10-18
Add logic for L2 token to revert on getters not available in L1

fredlacs:


packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/BytesParser.sol
```diff
@@ -19,32 +19,51 @@
...
-            return DebugPrint.bytes32string(input.toBytes32(0));
+            // TODO: can validate anything other than length and being null terminated?
+            if (input[31] != bytes1(0x00)) return (false, res);
+            else success = true;
+
+            // here we assume its a null terminated Bytes32 string
+            // https://github.com/ethereum/solidity/blob/5852972ec148bc041909400affc778dee66d384d/test/libsolidity/semanticTests/externalContracts/_stringutils/stringutils.sol#L89
+            // https://github.com/Arachnid/solidity-stringutils
+            uint256 len = 32;
+            while (input[len - 1] == bytes1(0x00)) {
```
DZGoldman, line 50: sanity checking my understanding here - we could never even hit the out bounds error here b/c the empty bytes array check is handled above?

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/BytesParser.sol
```diff
@@ -19,32 +19,51 @@
...
+            for (uint8 i = 0; i < len; i++) {
+                foo[i] = input[i];
+            }
+            // we can't just do `res := input` because of the null values in the end
+            // TODO: can we instead use a bitwise AND? build it dynamically with the length
+            assembly {
+                res := foo
+            }
         } else {
             // TODO: try catch to handle error
```
DZGoldman, line 64: what would we want the error handling behavior to be here (and above)? simply returning success false instead of throwing so the standardtoken intializing/parsing doesn't revert?

DZGoldman (commented):
looks good - left 2 comments, one checking my understanding, one clarifying whether or not this is ready as is

packages/arb-bridge-peripherals/contracts/tokenbridge/test/TestERC20.sol
```diff
@@ -29,3 +30,47 @@ contract TestERC20 is aeERC20 {
...
+
+contract Bytes32ERC20WithMetadata is Bytes32ERC20 {
+    bytes32 public name = 0x4d616b6572000000000000000000000000000000000000000000000000000000;
+    bytes32 public symbol = 0x4d4b520000000000000000000000000000000000000000000000000000000000;
+    // TODO: what if this overflows?
+    uint8 public decimals = 18;
+
+    // no totalSupply field
+    // uint256 public totalSupply;
+}
```
hkalodner, line 76: For testing purposes might be good to have these set by a constructor so that we can ex. add an all 0s case

PR #1730, opened 2021-10-18
Tracing frontend implementation

joshuacolvin0:


packages/arb-avm-cpp/cavm/carbcore.h
```diff
@@ -112,6 +112,12 @@ CMachine* arbCoreTakeMachine(CArbCore* arbcore_ptr,
 CMachineResult arbCoreGetMachineAtBlock(CArbCore* arbcore_ptr,
                                         uint64_t block_number,
                                         int allow_slow_lookup);
+CMachineResult arbCoreGetMachineForSideload(CArbCore* arbcore_ptr,
+                                            uint64_t block_number,
+                                            int allow_slow_lookup);
+CExecutionCursorResult arbCoreGetExecutionCursorAtBlock(CArbCore* arbcore_ptr,
```
hkalodner, line 115: Now that with have this, maybe we can remove `arbCoreGetMachineAtBlock`?

packages/arb-avm-cpp/cmachine/arbcore.go
```diff
@@ -526,3 +526,20 @@ func (ac *ArbCore) GetMachineAtBlock(blockNumber uint64, allowSlowLookup bool) (
 
 	return WrapCMachine(cMachineResult.machine), nil
 }
+
+func (ac *ArbCore) GetExecutionCursorAtBlock(blockNumber uint64, allowSlowLookup bool) (core.ExecutionCursor, error) {
+	CallowSlowLookup := 0
+	if allowSlowLookup {
+		CallowSlowLookup = 1
+	}
```
hkalodner, line 534: I think we have a boolToInt method now?

packages/arb-avm-cpp/cmachine/arbcore.go
```diff
@@ -526,3 +526,20 @@ func (ac *ArbCore) GetMachineAtBlock(blockNumber uint64, allowSlowLookup bool) (
...
+	cExecutionCursorResult := C.arbCoreGetExecutionCursorAtBlock(ac.c, C.uint64_t(blockNumber), C.int(CallowSlowLookup))
+
+	if cExecutionCursorResult.slow_error == 1 {
+		return nil, errors.Errorf("missing trie node 0000000000000000000000000000000000000000000000000000000000000000 (path )")
```
hkalodner, line 538: Probably should make this error a variable rather than including it inline

packages/arb-avm-cpp/cmachine/arbcore.go
joshuacolvin0: :+1:

packages/arb-avm-cpp/cmachine/arbcore.go
joshuacolvin0: :+1:

PR #1737, opened 2021-10-19
deploy new standard token script

DZGoldman:


packages/arb-ts/README.md
```diff
@@ -31,3 +31,25 @@ Defaults to `rinkArby`, for custom network use `--network` flag.
...
+
+3. `cd packages/arb-ts`
+
+4. Set `PRIVKEY` environmental variable to the key of the account from which you'll be deploying (account should have some balance of the token you're bridging).
+
+5. `yarn bridgeStandardToken`
+
+Required CL params:
+`networkID`:number — Chain ID of L1 network
+`l1_token_address`:string — address of L1 token to be bridged
```
fredlacs, line 51: i think its cleaner if we use camel case all over

```suggestion
`networkID`:number — Chain ID of L1 network
`l1TokenAddress`:string — address of L1 token to be bridged
```

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
+import { instantiateBridge } from './instantiate_bridge'
+import { ERC20__factory } from '../src/lib/abi/factories/ERC20__factory'
+import dotenv from 'dotenv'
+import args from './getCLargs'
+import { constants, BigNumber, utils } from 'ethers'
+
+dotenv.config()
```
fredlacs, line 7: why do we ask users to set `PRIVKEY` env variable in the readme if we're using dotenv?

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const l1Token = ERC20__factory.connect(l1TokenAddress, bridge.l1Provider)
+  // TODO: check blacklist
+
+  /* Looks like an L1 token: */
+  try {
+    await l1Token.symbol()
+    await l1Token.name()
+    await l1Token.decimals()
+    await l1Token.balanceOf(walletAddress)
```
fredlacs, line 38: we can multicall these

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const { l1_token_address: l1TokenAddress, amount } = args
+const amountBigNum = BigNumber.from(amount)
+
+const main = async () => {
+  const { bridge } = await instantiateBridge(privKey, privKey)
+  const walletAddress = await bridge.l1Bridge.getWalletAddress()
+
+  const l1Token = ERC20__factory.connect(l1TokenAddress, bridge.l1Provider)
+  // TODO: check blacklist
```
fredlacs, line 31: We could add a disclaimer about what isn't bridgeable and a link to the docs and force the user to say "yes" they have read / are aware / know what they are doing etc

packages/arb-ts/scripts/deployStandard.ts
fredlacs: i think that if you use the L1 bridge's "get token data" it should work
https://github.com/OffchainLabs/arbitrum/blob/2c2cd9872e959c77be45a72af075c67379658cad/packages/arb-ts/src/lib/l1Bridge.ts#L133

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+    await l1Token.balanceOf(walletAddress)
+  } catch (err) {
+    console.warn(`${l1TokenAddress} doesn't look like an L1 ERC20 token`)
+    throw err
+  }
+  /* check that you have some eth */
+  const walletBal = await bridge.l1Provider.getBalance(walletAddress)
+  if (walletBal.eq(constants.Zero)) {
+    throw new Error(`${walletAddress} has no Ether to pay for gas`)
+  }
```
fredlacs, line 47: is this a useful check? ethers will surface "not enough funds for gas" if the user doesnt have enough balance

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+  /* check token bal */
+  const tokenBal = await l1Token.balanceOf(walletAddress)
```
fredlacs, line 49: we already check `await l1Token.balanceOf(walletAddress)` we can keep that value instead of querying again

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+  if (tokenBal.lte(amountBigNum)) {
```
fredlacs, line 50:
```suggestion
  if (tokenBal.lt(amountBigNum)) {
```

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+    )
+  }
+
+  /* check token not yet deployed */
+  const l2TokenAddress = await bridge.l1Bridge.getERC20L2Address(l1TokenAddress)
+  if (l2TokenAddress === constants.AddressZero) {
+    throw new Error(`${l1TokenAddress} can't be bridged`)
+  }
+
+  if ((await bridge.l2Provider.getCode(l2TokenAddress)).length > 2) {
```
fredlacs, line 62: we do this "getCode" then check if none in multiple places. we could have a util function instead

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+    throw new Error(
+      `${l1TokenAddress} already deployed on L2 at ${l2TokenAddress}`
+    )
+  }
+  /* check allowance */
+
+  const isAllowed = await (async () => {
```
fredlacs, line 69: can't we use the allowance logic from https://github.com/OffchainLabs/arbitrum/blob/2c2cd9872e959c77be45a72af075c67379658cad/packages/arb-ts/src/lib/l1Bridge.ts#L133

packages/arb-ts/scripts/deployStandard.ts
DZGoldman: I don't follow - dotenv just let's you load env variables from .env if you want

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+    )
+  }
+  const seqNum = seqNums[0]
+
+  const userTxnHash = await bridge.calculateL2RetryableTransactionHash(seqNum)
+  console.log(
+    `Waiting for L2 txn; this takes ~10 minutes; waiting for L2 txn hash: ${userTxnHash}`
+  )
+
+  await bridge.waitForRetryableReceipt(seqNum, 2)
```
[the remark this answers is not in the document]
DZGoldman: yes, I'd rather do that change separate from this PR tho since it'll involve touching a lot of other things

packages/arb-ts/scripts/deployStandard.ts
```diff
@@ -0,0 +1,131 @@
...
+    )
+  }
+  const depositParams = {
+    erc20L1Address: l1TokenAddress,
+    amount: amountBigNum,
+  }
+  /* check for required gas */
+  const gasNeeded = await bridge.estimateGasDeposit(depositParams)
+
+  if (gasNeeded.gt(walletBal)) {
```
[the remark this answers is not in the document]
DZGoldman: naw both are L1

PR #1747, opened 2021-10-21
Handle missing cursor properly

joshuacolvin0:
If archive support is not enabled, handle missing execution cursor properly


PlasmaPower (approved):
LGTM

PR #1764, opened 2021-10-27
remove old token list generation scripts

DZGoldman:
in favor of https://github.com/OffchainLabs/arb-token-lists


fredlacs (approved):
LGTM

PR #1779, opened 2021-10-29
Typechain gen

fredlacs:
this PR relies on changes from https://github.com/OffchainLabs/arb-os/pull/611


packages/arb-ts/scripts/cleanCompileContracts.ts
```diff
@@ -0,0 +1,27 @@
...
+// this breaks on browsers, we can instead do a dynamic require to surface the error cleanly?
+import { execSync } from 'child_process'
+
+const cleanCompileContracts = async () => {
+  const arbosPath = getPackagePath('arbos-precompiles')
+  const ethBridgePath = getPackagePath('arb-bridge-eth')
+  const peripheralsPath = getPackagePath('arb-bridge-eth')
+
+  console.log('Clean building arbos')
+  const stdout1 = execSync(`cd ${arbosPath} && yarn clean:build`)
```
hkalodner, line 11: Maybe simpler to use`yarn workspace arb-precompiles clean:build` in the package.json instead of needing this script?

packages/arb-ts/scripts/genAbi.ts
```diff
@@ -0,0 +1,47 @@
...
+  console.log('building arbos')
+  const stdout1 = execSync(`cd ${arbosPath} && yarn build`)
+  console.log('building ethbridge')
+  const stdout2 = execSync(`cd ${ethBridgePath} && yarn build`)
+  console.log('building peripherals')
+  const stdout3 = execSync(`cd ${peripheralsPath} && yarn build`)
+
+  console.log('Done compiling')
+
+  const allFiles = glob(cwd, [
```
hkalodner, line 22: We can re-use the path vars from above, ex. `arbosPath`

hkalodner (changes requested):

packages/arb-ts/scripts/cleanCompileContracts.ts
hkalodner: Nevermind, this is so it works when checked out without the monorepo

packages/arb-ts/package.json
```diff
@@ -41,15 +43,20 @@
...
     "@nomiclabs/hardhat-ethers": "^2.0.2",
-    "@types/prompts": "^2.0.14",
+    "@typechain/ethers-v5": "^7.2.0",
+    "arb-bridge-eth": "0.7.2",
+    "arb-bridge-peripherals": "1.0.0",
+    "arbos-precompiles": "1.0.1",
     "dotenv": "^10.0.0",
     "ethers": "^5.1.0",
-    "prompts": "^2.4.2"
+    "ts-node": "^10.2.1"
```
DZGoldman, line 52: can't this stay in dev dependencies? i.e., only additionally used in install hooks, no?

fredlacs: we use ts-node in the psotinstall hook. if this is consumed as a lib, the devdependency won't be installed.
an alternative is to actually compile it with tsc and run with node

DZGoldman:
successful build here https://github.com/OffchainLabs/arb-token-bridge/pull/211

packages/arb-ts/src/index.ts
```diff
@@ -16,17 +16,6 @@
 /* eslint-env node */
 'use strict'
 
-export { RollupCreator__factory } from './lib/abi/factories/RollupCreator__factory'
-export { Inbox__factory } from './lib/abi/factories/Inbox__factory'
```
DZGoldman, line 20: does dis mean I can't import these externally from arb-ts anymore?

package.json
```diff
@@ -66,6 +66,7 @@
   },
   "resolutions": {
     "eth-sig-util/ethereumjs-abi": "^0.6.8-1",
-    "typechain/web3-v1/web3": "^1.0.0"
+    "typechain/web3-v1/web3": "^1.0.0",
+    "ansi-regex": "^5.0.1"
```
DZGoldman, line 70: what's this for?

DZGoldman (approved):
lgtm - all my comments are very minor. I think we can PR this right into master (or just merge from arb-ts-next right into master; arb-ts-next has nothing new yet)

package.json
fredlacs: yarn audit was complaining about the version of this dependency, so i forced a bump up this way

packages/arb-ts/src/index.ts
fredlacs: you can import it as in https://github.com/OffchainLabs/arb-token-bridge/pull/211/commits/2e0c13cc62154265f9551673f8e3ea17b05e973a

DZGoldman (approved):

fredlacs:
closed in favour of https://github.com/OffchainLabs/arbitrum/pull/1906

PR #1785, opened 2021-10-30
Fix memory leaks in unit tests

joshuacolvin0:


PlasmaPower:
We should consider just putting `closeArbStorage` in the `ArbStorage` destructor.

PR #1802, opened 2021-11-07
Add full vim version to dist and strip symbols from rocksdb

joshuacolvin0:
Also add back in procps tools and make beginning of dist-base docker file the same as cpp-base docker file.


hkalodner (approved):
LGTM

PR #1799, opened 2021-11-05
Upgrade 49

RachelFranks:
Upgrades ArbOS to the 49th version, which includes
- https://github.com/OffchainLabs/arb-os/pull/640

In summary, this upgrade corrects the behavior of modexp

PR #1804, opened 2021-11-09
Upgrade 50

RachelFranks:
Upgrades ArbOS to the 50h version, which includes
- https://github.com/OffchainLabs/arb-os/pull/638
- https://github.com/OffchainLabs/arb-os/pull/639

In summary, this upgrade corrects gas estimation's behavior when the gas price is zero


hkalodner (approved):
LGTM

hkalodner:
Merged accidentally, but superseded by https://github.com/OffchainLabs/arbitrum/pull/1799 so no need to revert

PR #1812, opened 2021-11-10
Update Catch2 to 2.13.7 to fix build w/ newer libc

Tristan-Wilson:
2.13.7 is the latest minor/patch revision on the 2.x major version of
Catch2. This is to fix the build on distros with newer versions of libc
like Fedora 35, which were seeing the same issue as
https://github.com/OpenImageDenoise/oidn/issues/106


hkalodner (approved):
LGTM

PR #1819, opened 2021-11-11
Apply clang-tidy recommendations

joshuacolvin0:


packages/arb-avm-cpp/avm_values/src/codepoint.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include <iostream>
 
-Operation::Operation(OpCode opcode_, value immediate_)
+Operation::Operation(OpCode opcode_, const value& immediate_)
     : opcode(opcode_), immediate(std::make_unique<value>(immediate_)) {}
```
hkalodner, line 26: Rather than making this `const` let's `std::move(immediate_)` instead

packages/arb-avm-cpp/data_storage/include/data_storage/datastorage.hpp
```diff
@@ -85,9 +85,9 @@ class Transaction {
         : datastorage(std::move(datastorage_)),
           transaction(std::move(transaction_)) {}
 
-    rocksdb::Status commit() { return transaction->Commit(); }
+    rocksdb::Status commit() const { return transaction->Commit(); }
 
-    rocksdb::Status rollback() { return transaction->Rollback(); }
+    rocksdb::Status rollback() const { return transaction->Rollback(); }
```
hkalodner, line 90: I kinda like these being non-const for semantic reasons

packages/arb-avm-cpp/cavm/cexecutioncursor.h
```diff
@@ -18,7 +18,7 @@
 
 #include "ctypes.h"
 
-#include <stdint.h>
```
hkalodner, line 21: These should work as c headers for the cgo build which doesn't include `cstdint`

hkalodner (changes requested):

packages/arb-avm-cpp/cavm/cexecutioncursor.h
joshuacolvin0: whoops, I knew that :)

packages/arb-avm-cpp/data_storage/include/data_storage/datastorage.hpp
joshuacolvin0: sounds good

hkalodner (approved):
LGTM

PR #1835, opened 2021-11-18
Preserve up to two batches worth of feed messages

joshuacolvin0:
To better work around synchronization race conditions, preserver two batches worth of sequencer feed messages to send to clients on connect/reconnect to better handle synchronization race conditions


hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

PR #1840, opened 2021-11-19
fix createRetryableTicket / depositEth address alias description in docs

DZGoldman:
A user caught this / was confused; still could probably be explained better, but currently it's just wrong.


fredlacs (approved):
LGTM
requires update to the docs repo

PR #1848, opened 2021-11-22
Clean up constructors and assignment operators of unions

hkalodner:
Based on recommendation from https://en.cppreference.com/w/cpp/language/union, unions should handle assignment to complex types via placement new


PlasmaPower (approved):
LGTM. I was unaware of this syntax, but this probably removes undefined behavior from freeing the old contents of the field.

PR #1855, opened 2021-11-24
Use TCMalloc instead of JEMalloc

joshuacolvin0:
getting better memory performance with TCMalloc, probably because of the large number of threads that are created and destroyed.


packages/arb-avm-cpp/CMakeLists.txt
```diff
@@ -83,16 +91,58 @@ elseif (${AVM_FUZZER} MATCHES "honggfuzz")
...
+if(ENABLE_GPROF)
+    message(STATUS "Using GPROF")
+endif()
+
+if (ENABLE_GPROF)
+    find_package(Gperftools)
+    if(GPERFTOOLS_FOUND)
+        message(STATUS "Using GPROF")
+        include_directories(${GPERFTOOLS_INCLUDE_DIR})
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ALT_MALLOC_FLAGS} -fno-omit-frame-pointer -g -pg")
```
hkalodner, line 113: I think we don't need the `-fno-omit-frame-pointer` anymore since we always apply that globally

packages/arb-avm-cpp/CMakeLists.txt
```diff
@@ -83,16 +91,58 @@ elseif (${AVM_FUZZER} MATCHES "honggfuzz")
...
 
+set(ALT_MALLOC_FLAGS "-fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free")
+
+if(ENABLE_GPROF)
+    message(STATUS "Using GPROF")
+endif()
+
+if (ENABLE_GPROF)
+    find_package(Gperftools)
+    if(GPERFTOOLS_FOUND)
```
hkalodner, line 110: Should we error if this isn't the case?

hkalodner: To do that, we can just have a `REQUIRED` on the `find_package

packages/arb-avm-cpp/CMakeLists.txt
```diff
@@ -83,16 +91,58 @@ elseif (${AVM_FUZZER} MATCHES "honggfuzz")
...
+        set(arb_node_LINKER_LIBS ${arb_node_LINKER_LIBS} ${GPERFTOOLS_PROFILER})
+        set(CMAKE_LINK_LIBRARY_FILE_FLAG "${CMAKE_LINK_LIBRARY_FILE_FLAG} -g -pg")
+        set(ENABLE_JEMALLOC 0)
+        set(ENABLE_TCMALLOC 1)
+    endif(GPERFTOOLS_FOUND)
+endif (ENABLE_GPROF)
+
+if (ENABLE_TCMALLOC)
+    find_package(Gperftools)
+    if(GPERFTOOLS_FOUND)
```
hkalodner, line 124: Also should have Gperftools required

packages/arb-avm-cpp/CMakeLists.txt
```diff
@@ -83,16 +91,58 @@ elseif (${AVM_FUZZER} MATCHES "honggfuzz")
...
+        include_directories(${GPERFTOOLS_INCLUDE_DIR})
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ALT_MALLOC_FLAGS}")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ALT_MALLOC_FLAGS}")
+        set(arb_node_LINKER_LIBS ${arb_node_LINKER_LIBS} ${GPERFTOOLS_TCMALLOC})
+        set(ENABLE_JEMALLOC 0)
+    endif(GPERFTOOLS_FOUND)
+endif (ENABLE_TCMALLOC)
+
 if (ENABLE_JEMALLOC)
+    if(GPERFTOOLS_FOUND)
```
hkalodner, line 135: Probably should just explicitly fail if ENABLE_TCMALLOC and ENABLE_JEMALLOC are both true

hkalodner (changes requested):

packages/arb-avm-cpp/CMakeLists.txt
joshuacolvin0: :+1:

packages/arb-avm-cpp/CMakeLists.txt
joshuacolvin0: :+1:

packages/arb-avm-cpp/CMakeLists.txt
joshuacolvin0: :+1:

packages/arb-avm-cpp/CMakeLists.txt
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1873, opened 2021-12-02
Remove Node contract in favour of struct

yahgwai:
Closes https://github.com/OffchainLabs/App_Engineering_-_Docs__PM/issues/86


packages/arb-bridge-eth/contracts/rollup/IRollupCore.sol
```diff
@@ -71,7 +73,16 @@ interface IRollupCore {
         uint256 finalBalance
     );
 
-    function getNode(uint256 nodeNum) external view returns (INode);
+    /**
+     * @notice Get the Node properties for the given index. 
+     * We can't expose the full Node publicly as it contains a mapping 
+     */
+    function getNodeProps(uint256 nodeNum) external view returns (NodeProps memory);
```
fredlacs, line 80: Any reason to change getNode to getNodeProps?
This doesn't need to be a breaking change. Could instead flatten the 2 node structs into 1, or call the outer node "NodeWithStaker" and inner "Node" to keep previous interface

packages/arb-bridge-eth/contracts/rollup/IRollupCore.sol
```diff
@@ -71,7 +73,16 @@ interface IRollupCore {
...
+
+    /**
+     * @notice Check if the specified node has been staked on by the provided staker
+     */
+    function nodeHasStaker(uint256 nodeNum, address staker) external view returns(bool);
```
fredlacs, line 85: `returns(bool)` missing a whitespace in there.
Do you have the precommit linter configured?

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -17,131 +17,146 @@
...
 
-    /// @notice Hash of the data that can be challenged
-    bytes32 public override challengeHash;
+    // Hash of the data that will be committed if this node is confirmed
+    bytes32 confirmData;
 
-    /// @notice Hash of the data that will be committed if this node is confirmed
-    bytes32 public override confirmData;
+    // Index of the node previous to this one
+    uint256 prev;
```
fredlacs, line 35: we could instead define `prev` as a library function that actually gets the previous node struct.
we could instead rename this to prevNum or something

fredlacs: a bit gimmicky, but could be neat

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -17,131 +17,146 @@
...
 
-    /// @notice Deadline at which a child of this node can be confirmed
-    uint256 public override noChildConfirmedBeforeBlock;
+    // Number of stakers staked on this node. This includes real stakers and zombies
+    uint256 stakerCount;
 
-    /// @notice Number of stakers staked on this node. This includes real stakers and zombies
-    uint256 public override stakerCount;
+    // Address of the rollup contract to which this node belongs
+    address rollup;
```
fredlacs, line 47: I don't believe we need this anymore

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -17,131 +17,146 @@
...
+    // All other Node data
+    NodeProps props;
+}
 
-    /// @notice The number of the latest child of this node to be created
-    uint256 public override latestChildNumber;
+/**
+ * @notice Utility functions for NodeProps
+ */
+library NodePropsLib {
```
fredlacs, line 67: Any reason not to have these functions in the NodeLib? we already have the typing to be sure if a given param is node or nodeprop

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -17,131 +17,146 @@
...
+    ) internal pure returns (Node memory) {
+        NodeProps memory props;
+        props.stateHash = _stateHash;
+        props.challengeHash = _challengeHash;
+        props.confirmData = _confirmData;
+        props.prev = _prev;
+        props.deadlineBlock = _deadlineBlock;
+        props.noChildConfirmedBeforeBlock = _deadlineBlock;
+
+        Node memory node;
```
fredlacs, line 134: why do we return this empty memory node? seems error prone. we can prob just drop the return value here

fredlacs (changes requested):

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: It has the props on it - we store the struct returned from this function later on

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: It's because the extend the NodeProps struct - the first arg of all these functions is NodeProps

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: Yep, I'll get rid

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: We actually do quite a bit of node number comparing with it, so I think it's good to keep it there. Will rename to prevNum

packages/arb-bridge-eth/contracts/rollup/IRollupCore.sol
yahgwai: Had NodeWithStaker originally, happy to use that. I think I'm in favour of going with your idea of completely separating the stakers into their own mapping

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: Merged now that we've removed stakers

packages/arb-bridge-eth/contracts/rollup/IRollupCore.sol
yahgwai: We now have a public getNode and an internal getNodeStorage.

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -21,57 +21,63 @@ pragma experimental ABIEncoderV2;
...
 /**
- * @notice Utility functions for NodeProps
+ * @notice Utility functions for Node
  */
-library NodePropsLib {
+library NodeLib {
     using SafeMath for uint256;
 
+    /**
+     * @notice Initialise a Node
```
fredlacs, line 52: initialize 🏴󠁧󠁢󠁥󠁮󠁧󠁿

packages/arb-bridge-eth/contracts/rollup/Node.sol
```diff
@@ -21,57 +21,63 @@ pragma experimental ABIEncoderV2;
...
+     * @param _stateHash Initial value of stateHash
+     * @param _challengeHash Initial value of challengeHash
+     * @param _confirmData Initial value of confirmData
+     * @param _prevNum Initial value of prevNum
+     * @param _deadlineBlock Initial value of deadlineBlock
+     */
+    function initialize(
```
fredlacs, line 59: If we wanted to initialise without copying to memory, we could probably pass in a storage pointer of an empty struct then write to it, ie
```solidity
function initialize(
    Node storage self,
    ...
)
```

There also is another notation for creating structs which I particularly like, but it prob has the same gas cost
```solidity
Node memory node = Node({
    stateHash: _stateHash,
    ...
});
```

packages/arb-bridge-eth/contracts/rollup/RollupCore.sol
```diff
@@ -96,27 +96,26 @@ abstract contract RollupCore is IRollupCore, Cloneable, Pausable {
     mapping(address => uint256) private _withdrawableFunds;
 
     /**
-     * @notice Get the Node the given node index
+     * @notice Get a storage refernce to the Node for the given node index
```
fredlacs, line 99:
```suggestion
     * @notice Get a storage reference to the Node for the given node index
```

fredlacs (changes requested):
Two small typos and a potential gas optimisation

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: Yeah, I think we'll only save a couple 100 gas doing that. We also may need an extra local var in one of the calling functions, and I'm not sure we have the stack space. Maybe lets leave this for a later improvement?

Regarding the second notation, I prefer it too, however it wouldn't let me use it without specifying all the parameter and there's a bunch we dont actually need to initialise.

packages/arb-bridge-eth/contracts/rollup/Node.sol
yahgwai: fixed

fredlacs (approved):
LGTM

PR #1874, opened 2021-12-02
Upgrade 50

RachelFranks:
This PR upgrades ArbOS the 50th version by integrating the following PRs
- https://github.com/OffchainLabs/arb-os/pull/644
- https://github.com/OffchainLabs/arb-os/pull/643
- https://github.com/OffchainLabs/arb-os/pull/642
- https://github.com/OffchainLabs/arb-os/pull/645

The integration work includes
- No longer overwriting the gas price during estimation for gas price 0
- A test for gas price 0 estimation


hkalodner (approved):
LGTM

PR #1875, opened 2021-12-02
Create validator wallet lazily

PlasmaPower:


joshuacolvin0 (approved):
LGTM

packages/arb-node-core/ethbridge/validator.go
```diff
@@ -53,29 +53,40 @@ func init() {
...
-		auth:          auth,
-		rollupAddress: rollupAddress,
+		con:               con,
+		address:           address,
+		onWalletCreated:   onWalletCreated,
+		client:            client,
+		auth:              auth,
+		rollupAddress:     rollupAddress,
+		walletFactoryAddr: walletFactoryAddr,
+		// TODO: rollupFromBlock
```
hkalodner, line 83: Do we want to resolve this todo?

packages/arb-node-core/ethbridge/validator.go
```diff
@@ -93,6 +104,27 @@ func (v *ValidatorWallet) executeTransaction(ctx context.Context, tx *types.Tran
...
+		addr, err := CreateValidatorWallet(ctx, v.walletFactoryAddr, v.rollupFromBlock, v.auth, v.client)
+		if err != nil {
+			return err
+		}
+		v.address = &addr
+		if v.onWalletCreated != nil {
+			v.onWalletCreated(addr)
+		}
+	}
+	if v.con == nil {
```
hkalodner, line 118: I think we might as well do this unconditionally if we're setting the address

packages/arb-node-core/ethbridge/validator.go
```diff
@@ -93,6 +104,27 @@ func (v *ValidatorWallet) executeTransaction(ctx context.Context, tx *types.Tran
 	})
 }
 
+func (v *ValidatorWallet) createWalletIfNeeded(ctx context.Context) error {
+	if v.address == nil {
```
hkalodner, line 108: Nice to flatten this with an early return

hkalodner (changes requested):

packages/arb-node-core/staker/staker.go
```diff
@@ -186,11 +186,20 @@ func (s *Staker) Act(ctx context.Context) (*arbtransaction.ArbTransaction, error
...
+	}
+	if walletAddress != nil {
+		var err error
+		rawInfo, err = s.rollup.StakerInfo(ctx, walletAddressOrZero)
+		if err != nil {
+			return nil, err
+		}
 	}
-	latestStakedNode, latestStakedNodeHash, err := s.validatorUtils.LatestStaked(ctx, s.wallet.Address())
+	latestStakedNode, latestStakedNodeHash, err := s.validatorUtils.LatestStaked(ctx, walletAddressOrZero)
```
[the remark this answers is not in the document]
PlasmaPower: It returns the latest node and its hash atomically. I've added a comment with that info.

packages/arb-node-core/ethbridge/validator.go
PlasmaPower: Resolved, thanks

packages/arb-node-core/ethbridge/validator.go
PlasmaPower: Fixed as part of the flattening

packages/arb-node-core/ethbridge/validator.go
PlasmaPower: I've flattened this somewhat. It's still not perfectly flat, as there might be a case where we've generated an address but don't have a con yet, but it's better than before :)

hkalodner (approved):
LGTM

PR #1877, opened 2021-12-03
Gracefully shutdown database

joshuacolvin0:
ensure that database is not in use when shutting it down.


packages/arb-avm-cpp/data_storage/include/data_storage/datastorage.hpp
```diff
@@ -56,18 +56,41 @@ class DataStorage {
     size_t next_column_to_flush{0};
     std::vector<uint8_t> secret_hash_seed;
 
+    class shutting_down_exception : public std::exception {};
+
+    class ConcurrentCounter {
+       public:
+        ConcurrentCounter() {
+            if (shutting_down) {
+                throw shutting_down_exception();
```
PlasmaPower, line 65: I think if this condition is hit the destructor will still decrement, which would be bad.

PlasmaPower: You might consider initializing the counter via a helper method which does the first shutting_down check, then calls a private constructor, which increments the counter and then does the second shutting_down check.

joshuacolvin0: Thanks, that's a lot cleaner

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1077,257 +1077,275 @@ void ArbCore::operator()() {
...
+            }
         }
 
-        if (!machineIdle() || message_data_status != MESSAGES_READY) {
-            // Machine is already running or no new messages, so sleep for a
-            // short while
-            std::this_thread::sleep_for(
-                std::chrono::milliseconds(coreConfig.idle_sleep_milliseconds));
-        }
+        std::cerr << "Exiting main ArbCore thread" << std::endl;
```
PlasmaPower, line 1341: Let's move this down to under the try-catch so we can search logs for it consistently

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -25,8 +25,14 @@
 #include <iostream>
 #include <string>
 #include <thread>
+#include <utility>
 
 DataStorage::DataStorage(const std::string& db_path) {
+    std::atomic_init(&shutting_down, false);
+    std::atomic_init(&concurrent_database_access_counter, 0);
```
PlasmaPower, line 32: Could we initialize these as fields in the constructor, like `DataStorage::DataStorage(const std::string& db_path) : shutting_down(false) {`? I was previously unaware `std::atomic_init` existed, but I don't think it's necessary here.

PlasmaPower (commented):
LGTM, just a couple minor things. To review a diff without whitespace (very useful for the ArbCore changes): https://github.com/OffchainLabs/arbitrum/pull/1877/files?w=1

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
joshuacolvin0: good catch

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -220,3 +260,46 @@ rocksdb::Status DataStorage::updateSecretHashSeed() {
...
+    }
+    return ConcurrentCounter(storage);
+}
+
+ConcurrentCounter::ConcurrentCounter(ConcurrentCounter&& other) noexcept
+    : storage(other.storage) {
+    other.storage = nullptr;
+}
+
+ConcurrentCounter& ConcurrentCounter::operator=(
```
hkalodner, line 299: Maybe even simpler to just delete this method? Seems like we probably don't need assignment

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -220,3 +260,46 @@ rocksdb::Status DataStorage::updateSecretHashSeed() {
...
+    : storage(_storage) {
+    storage->concurrent_database_access_counter++;
+    if (storage->shutting_down) {
+        // Destructor will take care of decrementing counter
+        throw DataStorage::shutting_down_exception();
+    }
+}
+
+ConcurrentCounter::~ConcurrentCounter() {
+    if (storage == nullptr) {
```
hkalodner, line 278: Would be nice to have a comment saying that this makes it so the destructor of a counter that's been moved won't run

packages/arb-avm-cpp/cavm/cmachine.h
```diff
@@ -64,7 +64,8 @@ CMachine* machineClone(CMachine* m);
 CStatus machineCurrentStatus(CMachine* m);
 CBlockReason machineIsBlocked(CMachine* m, int newMessages);
 
-RawAssertion executeAssertion(CMachine* m, const CMachineExecutionConfig* c);
+RawAssertionResult executeAssertion(CMachine* m,
```
hkalodner, line 67: I don't think it's critical, but I think many of the cgo methods could fail if the db is shutting down so  not sure if we want to do anything for those. The current generic exception handling behavior could be good enough

hkalodner (changes requested):
Naming suggestion, maybe `tryDBLockShared` for the lock method assuming it's relatively easy to change with automated tooling, otherwise it wouldn't be worth it

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -220,3 +260,46 @@ rocksdb::Status DataStorage::updateSecretHashSeed() {
...
+        return;
+    }
+    auto new_counter = storage->concurrent_database_access_counter.fetch_sub(1);
```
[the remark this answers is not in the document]
PlasmaPower: Hmm, I think we'd want the prefix decrement here actually. I forgot that returned a value.

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -220,3 +260,46 @@ rocksdb::Status DataStorage::updateSecretHashSeed() {
     }
     return status;
 }
+
+ConcurrentCounter DataStorage::getCounter() const {
+    return ConcurrentCounter::Get(this);
```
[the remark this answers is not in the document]
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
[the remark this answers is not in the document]
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
joshuacolvin0: Doesn't hurt anything, and good to have implementation to make sure nothing bad happens in the future.

packages/arb-avm-cpp/cavm/cmachine.h
joshuacolvin0: Added special handling in executeAssertion to reduce extraneous error messages when shutting down while under load.

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
```diff
@@ -274,31 +277,22 @@ ConcurrentCounter::ConcurrentCounter(const DataStorage* _storage)
...
-                        // where assert is disabled
-}
-
-ConcurrentCounter ConcurrentCounter::Get(const DataStorage* storage) {
-    if (storage->shutting_down) {
-        throw DataStorage::shutting_down_exception();
-    }
-    return ConcurrentCounter(storage);
+    storage->concurrent_database_access_counter--;
+    assert(storage->concurrent_database_access_counter.load() >= 0);
```
hkalodner, line 286: No need for two reads, we can just get the return value from the `--` line

hkalodner (changes requested):
I think `DbLock` or `DbSharedLock` for the class and then `tryDbLockShared` or even just `tryLockShared` instead of `getCounter` so then the acquisition lines look like
```cpp
auto locker = transaction->datastorage->tryLockShared();
```

packages/arb-avm-cpp/data_storage/src/datastorage.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1887, opened 2021-12-09
Add parameter for inbox-reader-delay

joshuacolvin0:


packages/arb-util/configuration/configuration.go
```diff
@@ -194,19 +194,23 @@ type Forwarder struct {
 	RpcMode   string `koanf:"rpc-mode"`
 }
 
+type InboxReader struct {
+	DelayBlocks int64 `koanf:"delay-blocks"`
+	Paranoid    bool  `koanf:"paranoid-inbox-reader"`
```
hkalodner, line 199: This should just be `paranoid`

hkalodner (changes requested):
Looks good other than comment

packages/arb-util/configuration/configuration.go
joshuacolvin0: good catch

hkalodner (approved):
LGTM

PR #1893, opened 2021-12-13
Refactor L1ToL2Message to read/write pattern

yahgwai:


packages/arb-ts/src/lib/L1ToL2Message/L1ToL2Message.ts
```diff
@@ -31,86 +29,123 @@ export interface L1ToL2MessageReceipt {
...
-    super()
-    this.initSignersAndProviders(signersAndProviders)
-    this.arbRetryableActions = new RetryableActions(signersAndProviders)
-    this.l1TxnHash = l1TxnHash
-  }
-  static async initFromL1Txn(
-    signersAndProviders: SignersAndProviders,
-    l1Txn: string | TransactionReceipt,
+export class L1ToL2Message {
+  protected static getMessageOrThrow(
```
DZGoldman, line 42: Unused - should this be used inside of fromL1Receipt / fromL1ReceiptAll?

packages/arb-ts/src/lib/L1ToL2Message/L1ToL2Message.ts
```diff
@@ -31,86 +29,123 @@ export interface L1ToL2MessageReceipt {
...
-    l1Txn: string | TransactionReceipt
-  ): Promise<L1ToL2Message[]> {
-    const l1TxnReceipt = await getTxnReceipt(
-      l1Txn,
-      signersAndProviders.l1Provider
-    )
-    const l1TxnHash = l1TxnReceipt.transactionHash
+  public static async fromL1ReceiptAll<T extends Provider | Signer>(
+    l2SignerOrProvider: T,
+    l1TxnReceipt: TransactionReceipt
```
DZGoldman, line 69: We could have a "from L1 Hash" version of all of these that just wraps around them and uses getTxnReceipt w/ an L1 provider... maybe that's overkill?

packages/arb-ts/src/lib/L1ToL2Message/RetryableActions.ts
```diff
@@ -1,63 +0,0 @@
-import {
```
DZGoldman, line 1: Since we can instantiate from the l1 receipt of the L2 ticket creation receipt, I think I'm okay with this just being removed

DZGoldman (approved):
I'm sold on this - (only v minor comment, feel free to merge into arb-ts-next--retryables when ready)

packages/arb-ts/src/lib/L1ToL2Message/L1ToL2Message.ts
yahgwai: yeah, i was before, but i think it's better just to inline. Removed

packages/arb-ts/src/lib/L1ToL2Message/L1ToL2Message.ts
yahgwai: We'd still need 3 methods I think: fromL1Hash, fromL1HashAll and fromL2Ticket. I think getting the receipt for a hash is a very common thing that people using ethers will be useful, so I think I'd rather leave them as they are.

PR #1895, opened 2021-12-14
Save database checkpoint on SIGUSR1

joshuacolvin0:
Check if directory exists every time.  Cannot check on startup because SIGUSR1 may never be used to trigger database checkpoint save.

### DIFF
--- a/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
+++ b/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
@@ -29,6 +29,7 @@
 #include <data_storage/value/value.hpp>
 #include <data_storage/value/valuecache.hpp>
 
+#include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


PlasmaPower:
Hmm.. we should consider doing this from the Go side so the signal doesn't need targeted at the core thread. This works too, but I believe it requires you target the SIGUSR1 at the core thread (I could be wrong).

joshuacolvin0:
> Hmm.. we should consider doing this from the Go side so the signal doesn't need targeted at the core thread. This works too, but I believe it requires you target the SIGUSR1 at the core thread (I could be wrong).

Worked really well during testing, just had to run `ps` to get PID and send SIGUSR1 to it.

PlasmaPower:
I might be wrong then. If that worked during testing this sounds good to me.

PlasmaPower (approved):
LGTM

PR #1897, opened 2021-12-14
Added first outline of l2tol1message refactoring

yahgwai:


packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+    return logs.map(
+      log => (iface.parseLog(log).args as unknown) as L2ToL1EventResult
+    )
+  }
+
+  private getOutboxAddr(network: Network, batchNumber: BigNumber) {
+    // CHRIS: add the old network to the networks object, and look it up here
+    // CHRIS: null check? this shouldn't be possible
+    // CHRIS: disable linting by just using the batchnumber here
+    batchNumber
```
DZGoldman, line 101: This is where we'll want to use the batch number to check for the appopriate outbox, correct?

packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+ */
+export type L2ToL1MessageReaderOrWriter<
+  T extends SignerOrProvider
+> = T extends Provider ? L2ToL1MessageReader : L2ToL1MessageWriter
+
+export class L2ToL1Message {
+  public static fromBatchNumber<T extends SignerOrProvider>(
+    l1SignerOrProvider: T,
+    outboxAddress: string,
+    batchNumber: BigNumber,
```
DZGoldman, line 143: Maybe we make a special `L2MessageId` interface that just combines these two (like we did with MessageID in L1ToL2Messages?

packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+    const eventTopic = iface.getEventTopic(event)
+
+    const topics = filter.topics ? filter.topics : []
+    // CHRIS: keep the warn?
+    // if (!filter.fromBlock && !filter.toBlock)
+    // console.warn('Attempting to query from 0 to block latest')
+    const logs = await l2Provider.getLogs({
+      address: ARB_SYS_ADDRESS,
+      topics: [eventTopic, ...topics],
+      fromBlock: filter.fromBlock || 0,
```
DZGoldman, line 183: I think these default to 0 / latest if undefined anyway, not sure why I did this

DZGoldman: Probably just having the param be a special thing that's just {fromBlock, toBlock, topics} makes more sense, since the address fields won't take effect anyway (I think @fredlacs has pointed this out)

packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+  ) {
+    const batch = await this.getL2ToL1MessageLogs(l2Provider, {
+      topics: [null, null, hexZeroPad(batchNumber.toHexString(), 32)],
+    })
+
+    const indexItem = batch.filter(b => b.indexInBatch.eq(indexInBatch))
+    if (indexItem.length === 1) {
+      return indexItem[0]
+    } else if (indexItem.length > 1) {
+      // CHRIS: warn, error?
```
DZGoldman, line 205: def "666" error here, since that would be an invalid state hehe. I think even if not found (below) also return error, but not sure

packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+    } catch (e) {
+      // CHRIS: discuss all the console logs going on in here and elsewhere, should we keep them?
+      // CHRIS: this error needs updating. also 666?
+      console.warn('666: error in getOutGoingMessageState:', e)
+      return OutgoingMessageState.NOT_FOUND
+    }
+  }
+
+  public async waitUntilOutboxEntryCreated(retryDelay = 500): Promise<void> {
+    // CHRIS: should we even be doing this? what's the use case? it could be a long wait
```
DZGoldman, line 315: agreed this is a weird method; no harm in keeping, but could get rid

packages/arb-ts/src/lib/message/L2ToL1Message.ts
```diff
@@ -0,0 +1,374 @@
...
+      return indexItem[0]
+    } else if (indexItem.length > 1) {
+      // CHRIS: warn, error?
+    }
+
+    return null
+  }
+}
+
+export class L2ToL1MessageReader extends L2ToL1Message {
```
DZGoldman, line 212: Now that we have this nice abstraction, something like "l1BlocksUntilConfirmable" would be a cool (doesn't have to be in this PR)

DZGoldman (commented):
Left some comments — i really like how the L1toL2 / L2toL1 message symmetry is taking shape

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -0,0 +1,395 @@
...
+      })
+      return L1ToL2Message.fromL2Ticket(
+        l2SignerOrProvider,
+        ticketCreationHash,
+        mn
+      )
+    })
+  }
+
+  // CHRIS: should we offer this select function?
```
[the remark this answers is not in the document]
yahgwai: Kept

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -0,0 +1,395 @@
...
+    if (!ticketCreationReceipt) {
+      return L1ToL2MessageStatus.NOT_YET_CREATED
+    }
+    if (ticketCreationReceipt.status === 0) {
+      return L1ToL2MessageStatus.CREATION_FAILED
+    }
+    if (await this.isExpired()) {
+      return L1ToL2MessageStatus.EXPIRED
+    }
+    // we could sanity check that autoredeem failed, but we don't need to
```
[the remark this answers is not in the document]
yahgwai: Added your comment

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -0,0 +1,395 @@
...
+  public async status(): Promise<L1ToL2MessageStatus> {
+    const userTxnReceipt = await this.l2Provider.getTransactionReceipt(
+      this.userTxnHash
+    )
+
+    const ticketCreationReceipt = await this.getTicketCreationReceipt()
+    return this.receiptsToStatus(userTxnReceipt, ticketCreationReceipt)
+  }
+
+  // CHRIS: this wait function should be broken up?
```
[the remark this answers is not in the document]
yahgwai: Hmm, I see why they're together now, since the autoredeem receipt wont really ever be available until after the ticket creation receipt.

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -0,0 +1,395 @@
...
+    )
+    return arbRetryableTx.getBeneficiary(this.userTxnHash)
+  }
+}
+
+export class L1ToL2MessageWriter extends L1ToL2MessageReader {
+  public constructor(
+    private readonly l2Signer: Signer,
+    l2TicketCreationTxnHash: string,
+    // CHRIS: do we really need this? it's never used
```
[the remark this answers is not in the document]
yahgwai: I think we could get rid, but I decided to keep it since it's part of the underlying data emitted by the event - so could have general use

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -0,0 +1,395 @@
...
+  public async redeemSafe(
+    waitTimeForL2Receipt = 900000
+  ): Promise<ContractTransaction> {
+    // CHRIS: we need way better messaging about what;s going on here
+    // CHRIS: we should go over the user flow to see if this is the best way
+    // CHRIS: also for the cancel flow
+    // CHRIS: also check the execute flow on the L2ToL1Message - we dont check status in there, should we?
+    console.log('waiting for retryable ticket...', this.userTxnHash)
+    const result = await this.wait(waitTimeForL2Receipt)
+    if (result.ticketCreationReceipt?.status == 0) {
```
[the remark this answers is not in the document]
yahgwai: updated to check `result.status === L1ToL2MessageStatus.NOT_YET_CREATED`

packages/arb-ts/src/lib/message/L2ToL1Message.ts
yahgwai: Yep - I need to get the old outbox addresses from somewhere

packages/arb-ts/src/lib/message/L2ToL1Message.ts
yahgwai: I've updated this in the asset-bridger branch

packages/arb-ts/src/lib/message/L2ToL1Message.ts
yahgwai: now throwing error

packages/arb-ts/src/lib/message/L2ToL1Message.ts
yahgwai: I got rid for now

packages/arb-ts/src/lib/message/L1ToL2Message.ts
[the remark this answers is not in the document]
DZGoldman: chat hopefully cleared things up — as is, error message "l2 txn failed" is off (should be "not yet created")

PR #1907, opened 2021-12-20
arb-ts v2: new networks handling strategy (/ applied to retryable ticket creator)

DZGoldman:

PR #1925, opened 2021-12-29
Asset bridgers

yahgwai:
Left the bridge classes commented out for now - have a small number of items to discuss and remove.


packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -128,41 +136,39 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
+      log => iface.parseLog(log).args as DepositInitiatedEvent['args']
     )
   }
+
+  /**
+   * Replaces the wait function with one that returns an L1TransactionReceipt
+   * @param contractTransaction
+   * @returns
+   */
+  public static swivelWait = (
```
DZGoldman, line 158: wow this name!

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+   * Get the corresponding L1 for the provided L2 token
+   * @param erc20L1Address
+   * @param l1Provider
+   * @returns
+   */
+  public getL1ERC20Address(
+    erc20L2Address: string,
+    l2Provider: Provider
+  ): Promise<string> {
+    const arbERC20 = StandardArbERC20__factory.connect(
```
DZGoldman, line 331: Ala above, user L2GatewayToken__factory instead

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+    return arbERC20.functions.l1Address().then(([res]) => res)
+  }
+
+  /**
+   * Whether the token has been disabled on the router
+   * @param l1TokenAddress
+   * @param l1Provider
+   * @returns
+   */
+  public async getL1TokenIsDisabled(
```
DZGoldman, line 344: style thing: i prefer just "l1TokenIsDisabled", (idk)

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+    l1Provider: Provider
+  ): Promise<boolean> {
+    const l1GatewayRouter = L1GatewayRouter__factory.connect(
+      this.l2Network.tokenBridge.l1GatewayRouter,
+      l1Provider
+    )
+
+    return (
+      (await l1GatewayRouter.l1TokenToGateway(l1TokenAddress)) ===
+      '0x0000000000000000000000000000000000000001'
```
DZGoldman, line 355: we could change precompile_addresses.ts to something like constants.ts and move this and the alias offset there, so it serves as a bit of a reference/spec

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+      }))
+
+    return await multiCaller.call(callInputs)
+  }
+}
+
+/**
+ * A token-gateway pair
+ */
+interface TokenGateway {
```
DZGoldman, line 616: This confused me below - maybe name it TokenToGatewayPairing or something?

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+    const l1GatewayRouter = L1GatewayRouter__factory.connect(
+      this.l2Network.tokenBridge.l1GatewayRouter,
+      l1Signer.provider
+    )
+
+    return await l1GatewayRouter.functions.setGateways(
+      tokenGateways.map(tG => tG.tokenAddr),
+      tokenGateways.map(tG => tG.gatewayAddr),
+      0,
+      l2GasPrice,
```
DZGoldman, line 712: Since we're hardcoding max gas to 0, we may as well also leave out l2GasPrice / hardcode that to 0 as well (I think gas price of 0 is fine here @fredlacs ?)

Basically this method creates a retryable ticket, but doesn't autoredeem it / requires manual L2 redemption; maybe we want the name to reflect that (i.e., "createSetGatewaysRetryable Ticket").

Fun little thing would be to have another method that wraps around it, takes in an L2 signer as well, calls it, waits for the L1ToL2Message, and redeems it on L2 directly.

(Alternatively, we could just actually try to estimate the maxGas required, but that's annoying)

packages/arb-ts/src/lib/message/L1ToL2MessageCreator.ts
```diff
@@ -78,7 +78,7 @@ export class L1ToL2MessageCreator {
     )
     const l2ChainID = (await l2Provider.getNetwork()).chainId.toString()
     const rec = await this.createRetryableTicketFromGasParams(
-      gasParams,
+      { ...gasParams, l2CallValue },
```
DZGoldman, line 81: A bit confused as to what happened here — was it just erroring at the static checker level before this?

packages/arb-ts/src/lib/assetBridger/assetBridger.ts
```diff
@@ -0,0 +1,58 @@
...
+import { L1ContractTransaction } from '../message/L1ToL2Message'
+import { L2ContractTransaction } from '../message/L2ToL1Message'
+
+import { L2Network } from '../utils/networks'
+
+/**
+ * Base for bridging assets from l1 to l2 and back
+ */
+export abstract class AssetBridger<DepositParams, WithdrawParams> {
+  public constructor(public readonly l2Network: L2Network) {}
```
DZGoldman, line 29: we could derive the L1Network directly here and store that as well (handy info to have)

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+
+  /**
+   * Get the address of the l1 gateway for this token
+   * @param erc20L1Address
+   * @param l1Provider
+   * @returns
+   */
+  public async getL1GatewayAddress(
+    erc20L1Address: string,
+    l1Provider: Provider
```
DZGoldman, line 132: Revisiting this general q of how to handle signers and providers: I'm probably generally overthinking this, but we should be consistent whatever we end up choosing; i.e., if we're gonna go with this pattern of requiring signers/providers to be passed in as params in each method, we should update L1ToL2MessageCreator accordingly (which currently sets up the signer in the constructor).  

One thing that feels slightly off to me is that we require a network object be passed into the constructor, but then ask for a signer/provider when calling a method, effectively assuming that the network & provider correspond. 

Something that might be useful (for me anyway) would be to see what a simple script that instantiates one of these classes (i.e., TokenApproveParams) looks like (without, i.e., using instantiateBridge) and see if there's some utility helpers we could include to it neaten up the flow.

DZGoldman: (See, i.e., https://github.com/OffchainLabs/arbitrum/pull/1907/files)

packages/arb-ts/src/lib/message/L1ToL2Message.ts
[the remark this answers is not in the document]
yahgwai: :recycle:

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
```diff
@@ -0,0 +1,719 @@
...
+   * Get the L2 token contract at the provided address
+   * @param l1Provider
+   * @param l1TokenAddr
+   * @returns
+   */
+  public getL2TokenContract(
+    l2Provider: Provider,
+    l2TokenAddr: string
+  ): StandardArbERC20 {
+    return StandardArbERC20__factory.connect(l2TokenAddr, l2Provider)
```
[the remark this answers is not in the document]
yahgwai: Updated

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
yahgwai: updated

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
yahgwai: updated

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
yahgwai: added a constants file and moved the address offset + this

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
yahgwai: woops, yep. Renamed to `TokenAndGateway`

packages/arb-ts/src/lib/message/L1ToL2MessageCreator.ts
yahgwai: No, I'm just adding in the l2CallValue. It was part of the return from `estimateGasValuesL1ToL2Creation` but I removed that as it wasnt generated in there I think

packages/arb-ts/src/lib/assetBridger/assetBridger.ts
yahgwai: fetch l1Network in the constructor

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
DZGoldman: Ooo maybe for now the best option is include maxGas a param to the method instead of hardcoding it to 0, and we can consume it as we please (this is basically admin only, so not too concerned for now; I have bigger plans r.e. helpers to register custom tokens etc)

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
DZGoldman: as discussed, we'll revisit this, but can ship for beta in its current form 👍

DZGoldman (approved):
Looking good - as discussed, happy to merge (barring merge conflict resolution, etc) for a beta publish and iterate from there.

packages/arb-ts/src/lib/assetBridger/tokenBridger.ts
yahgwai: added the maxgas param for now

PR #1927, opened 2021-12-30
Abort executeassertion when ctx cancelled

joshuacolvin0:
* Add atomic flag to abort machine execution
* Use goroutine to run assertion so that abort can be sent if ctx cancelled
* Pass ctx down as appropriate


hkalodner (approved):
LGTM

packages/arb-avm-cpp/cmachine/machine.go
```diff
@@ -194,12 +203,34 @@ func (m *Machine) ExecuteAssertionAdvanced(
...
+		// Still need to clean up result, so wait until returned
+		assertionResult = <-resultChan
+	case assertionResult = <-resultChan:
+	}
+
+	// Make sure result is cleaned up properly
+	executionAssertion, values, steps, err := makeExecutionAssertion(assertionResult.assertion)
+
+	if aborted {
+		return nil, nil, 0, errors.New("aborted")
```
PlasmaPower, line 227: Returning `ctx.Err()` here should give a normal context canceled error, which is probably a bit cleaner.

packages/arb-avm-cpp/avm/src/machine.cpp
```diff
@@ -63,14 +63,18 @@ void MachineExecutionConfig::setSideloadsFromBytes(
...
+
 Assertion Machine::run() {
     uint256_t start_steps = machine_state.output.total_steps;
     uint256_t start_gas = machine_state.output.arb_gas_used;
 
     bool has_gas_limit = machine_state.context.max_gas != 0;
     BlockReason block_reason = NotBlocked{};
     uint256_t initialConsumed = machine_state.getTotalMessagesRead();
-    while (true) {
+    while (!is_aborted) {
```
PlasmaPower, line 77: In addition to only checking this every 100 iterations or so, `is_aborted.load(std::memory_order_relaxed)` should be faster than the default seq_cst load and we don't need any memory order guarantees.

PlasmaPower (commented):
LGTM, just a couple minor comments

packages/arb-avm-cpp/cmachine/machine.go
joshuacolvin0: :+1:

packages/arb-avm-cpp/avm/src/machine.cpp
joshuacolvin0: :+1:

PlasmaPower (approved):
LGTM

PR #1889, opened 2021-12-10
Upgrade 52

RachelFranks:
Upgrades ArbOS to the 52nd version via the following PRs

- https://github.com/OffchainLabs/arb-os/pull/650
- https://github.com/OffchainLabs/arb-os/pull/631
- https://github.com/OffchainLabs/arb-os/pull/648
- https://github.com/OffchainLabs/arb-os/pull/654
- https://github.com/OffchainLabs/arb-os/pull/636
- https://github.com/OffchainLabs/arb-os/pull/603

In summary, this upgrade introduces a series of performance enhancements targeted at lowering compute gas costs.

PR #1936, opened 2022-01-04
Gas estimate tests sometimes fail due to underflow

yahgwai:
If `unfundedGas` > `fundedGas` then we have an underflow in these tests. This sometimes caused the tests to fail.


RachelFranks:
Nice catch! I ran into this when building `upgrade-52`, which we should have merged soon
- #1889

yahgwai:
Ok, cool, I'll close this in favour of https://github.com/OffchainLabs/arbitrum/pull/1889

PR #1937, opened 2022-01-04
Clean up log output while syncing

hkalodner:
- Change existing "Adding logs" line to debug
- Remove found empty block warn since it isn't useful or relevant
- Add "sync update" log that prints the l1 block, l2 block, and timestamp of the latest block processed by the system

The new setup will output logs with the same frequency as the old one, but with much more useful information


joshuacolvin0 (approved):
LGTM

PR #1943, opened 2022-01-07
Cache machine at last block when sequencer

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/combinedmachinecache.cpp
```diff
@@ -16,21 +16,42 @@
...
 }
 
 void CombinedMachineCache::lru_add(std::unique_ptr<Machine> machine) {
     std::unique_lock lock(mutex);
 
+    if (last_machine != nullptr &&
+        last_machine->machine_state.arb_gas_remaining <
+            machine->machine_state.arb_gas_remaining) {
+        last_machine = nullptr;
+    }
```
hkalodner, line 43: Probably good to add a helper method for this

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/combinedmachinecache.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #1945, opened 2022-01-07
Cache penultimate machine as well as last machine

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/combinedmachinecache.cpp
```diff
@@ -21,11 +21,16 @@ void CombinedMachineCache::checkLastMachine(uint256_t& arb_gas_used) {
...
+    if (last_last_machine != nullptr &&
+        last_last_machine->machine_state.output.arb_gas_used < arb_gas_used) {
+        last_last_machine = nullptr;
+    }
 }
 
 void CombinedMachineCache::lastAdd(std::unique_ptr<Machine> machine) {
     std::unique_lock lock(mutex);
 
+    last_last_machine = std::move(last_machine);
```
hkalodner, line 33: I think we should only do this if `last_machine` is non-empty

hkalodner (dismissed):
LGTM

packages/arb-avm-cpp/data_storage/src/combinedmachinecache.cpp
joshuacolvin0: :+1:

hkalodner (dismissed):
LGTM

hkalodner (approved):
LGTM

PR #1953, opened 2022-01-12
Cleanup SIGUSR1 signal handling

joshuacolvin0:
* Move signal handler from C++ to Go code for more consistency.
* Ensure that current machine checkpoint is saved to database before rocksdb checkpoint is created
* Only perform database pruning when saving checkpoint to avoid spending too much time pruning database.


hkalodner (approved):
LGTM

PR #1965, opened 2022-01-13
Arb ts v2: miscellaneous things 

DZGoldman:


packages/arb-ts/src/lib/utils/lib.ts
```diff
@@ -1,2 +1,61 @@
+import { ArbTsError } from '../errors'
+import { JsonRpcProvider } from '@ethersproject/providers'
+
 export const wait = (ms: number): Promise<void> =>
   new Promise(res => setTimeout(res, ms))
+
+export interface ArbTransactionReceipt {
```
yahgwai, line 7: Can this interface extend ethers.TransactionReceipt?

yahgwai (approved):

PR #1968, opened 2022-01-14
add looksLikeEthDeposit method to L1TransactionReceipt

DZGoldman:
Weird exception to how retryables are normally used, so nice to have this detector

### DIFF
--- a/packages/arb-ts/src/lib/message/L1ToL2Message.ts
+++ b/packages/arb-ts/src/lib/message/L1ToL2Message.ts
@@ -183,14 +183,19 @@ export class L1TransactionReceipt implements TransactionReceipt {
     )
   }
 
+  /**
+   * Check if tx is a direct call to a depositEth function (i.e., on the Inbox contract)
+   * @returns
+   */
   public async looksLikeEthDeposit(
     l1SignerOrProvider: SignerOrProvider
   ): Promise<boolean> {
     const l1Provider =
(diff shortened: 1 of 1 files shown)


fredlacs (approved):
LGTM

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -183,6 +183,16 @@ export class L1TransactionReceipt implements TransactionReceipt {
     )
   }
 
+  public async looksLikeEthDeposit(
```
yahgwai, line 186: Maybe add a comment explaining that this looks like a direct call to a specific function

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -183,6 +183,16 @@ export class L1TransactionReceipt implements TransactionReceipt {
...
+    l1SignerOrProvider: SignerOrProvider
+  ): Promise<boolean> {
+    const l1Provider =
+      SignerProviderUtils.getProviderOrThrow(l1SignerOrProvider)
+    const txRes = await l1Provider.getTransaction(this.transactionHash)
+    const ethDeposit_FUNCTION_SIG = '0x0f4d14e9'
```
yahgwai, line 192: Short comment to show what the plain text function sig is

yahgwai (approved):

PR #1970, opened 2022-01-15
Add extra logging when reorg detected in getMessagesImpl

joshuacolvin0:
* Make sure the getMessagesImpl only returns NotFound when reorg is detected
* Add logging when reorg is detected
* Always print extra checkpoint information when reorg occurs

### DIFF
--- a/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
+++ b/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
@@ -2048,8 +2048,7 @@ ValueResult<std::vector<RawMessageInfo>> ArbCore::getMessagesImpl(
                 delayed_msg_it->Next();
             }
 
-            if (!delayed_msg_it->status().IsNotFound() &&
-                !delayed_msg_it->status().ok()) {
+            if (!delayed_msg_it->status().ok()) {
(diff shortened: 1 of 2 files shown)


PlasmaPower:
A rocksdb iterator's status can never be NotFound from my understanding. When it runs out of items, it returns false from Valid, but ok from status.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -973,23 +973,20 @@ rocksdb::Status ArbCore::reorgCheckpoints(
...
+                  << ", L2 block: "
+                  << nearest_machine->machine_state.output.l2_block_number
+                  << ", log count: "
+                  << nearest_machine->machine_state.output.log_count
+                  << ", timestamp: "
+                  << std::put_time(
+                         std::localtime((time_t*)&nearest_machine->machine_state
+                                            .output.last_inbox_timestamp),
+                         "%c")
+                  << std::endl;
```
PlasmaPower, line 989: Given the current issue, maybe we should put a messages read count in here? Other than that, LGTM

joshuacolvin0: excellent idea

PR #1973, opened 2022-01-15
Fix initial cache seeding

joshuacolvin0:
On startup, node should load initial snapshot in the past and populate cache appropriately.


hkalodner (approved):
LGTM

PR #1991, opened 2022-01-20
Arb-ts updates required for arb-token-list

yahgwai:


packages/arb-ts/package.json
```diff
@@ -19,7 +19,7 @@
     "postinstall": "$npm_execpath run gen:abi",
     "prepublishOnly": "yarn build && yarn format",
     "preversion": "yarn lint",
-    "build": "./scripts/builder",
+    "build": "./scripts/builder && cp -R ./src/lib/abi/*.d.ts ./dist/lib/abi",
```
DZGoldman, line 22: why is this necessary?

yahgwai: Apparently tsc doesn't copy d.ts files, it only creates them. https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build

So I'm manually copying them as part of the build

PR #1993, opened 2022-01-21
Add tracing flags to dev node 

fredlacs:


packages/arb-rpc-node/cmd/arb-dev-node/arb-dev-node.go
```diff
@@ -103,6 +103,8 @@ func startup() error {
 	initialL1Height := fs.Uint64("l1height", 0, "initial l1 height")
 	rollupStr := fs.String("rollup", "", "address of rollup contract")
 	chainId64 := fs.Uint64("chainId", 68799, "chain id of chain")
+	tracingEnable := fs.Bool("node.rpc.tracing.enable", false, "enable tracing api")
```
hkalodner, line 106: I think it makes sense to have this on by default for the dev node. Arguably we might not even need an option for it, but it can't hurt

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #2002, opened 2022-01-25
Added batch info for arb receipt

yahgwai:
Also did a bit of organising.

Have the following open questions:
1. How to choose a delta in `getBatch` - and are we setting start/end block correctly?
2. Are any of the following fields nullabe?:
```
// l1SequenceNumber: bigNumber,
// parentRequestId: hash,
// indexInParent: number,
// arbType: number,
// arbSubType: number,
// l1BlockNumber: number,
// returnData: data, - could be empty
// returnCode: number, - never empty
// feeStats: {
//     prices: {
//         l1Transaction: bigNumber,
//         l1Calldata: bigNumber,
//         l2Storage: bigNumber,
//         l2Computation: bigNumber,
//     }
//     unitsUsed: {
//         l1Transaction: bigNumber,
//         l1Calldata: bigNumber,
//         l2Storage: bigNumber,
//         l2Computation: bigNumber,
//     }
//     paid: {
//         l1Transaction: bigNumber,
//         l1Calldata: bigNumber,
//         l2Storage: bigNumber,
//         l2Computation: bigNumber,
//     }
// }
```
3. The following dont have docs, and I couldnt find much about them. What do they mean?
```
// parentRequestId: string
//   indexInParent: number
//   arbType: number
//   arbSubType: number
// SenderNotPermitted = 13,
```


packages/arb-ts/src/lib/utils/arbProvider.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  public receipt(value: any): ArbTransactionReceipt {
+    return super.receipt(value) as ArbTransactionReceipt
+  }
+}
+
+const getBatch = async (
+  seqNum: BigNumber,
+  l1BlockNumber: number,
```
DZGoldman, line 129: Can we rename this param so it's clear that the caller of getBatch is responsible for providing an l1BlockNumber that corresponds to the receipt the sequencer gave for the tx with the provided seqnum?

(Maybe "l2TxnSeqNum" and "l2TxL1BlockNumber"

packages/arb-ts/src/lib/utils/arbProvider.ts
```diff
@@ -0,0 +1,246 @@
...
+  l1Provider: providers.Provider
+): Promise<Omit<BatchInfo, 'confirmations'> | null> => {
+  const ethNetwork = await getL1Network(l1Provider)
+  const arbNetwork = await getL2Network(ethNetwork.partnerChainIDs[0])
+  const inbox = SequencerInbox__factory.connect(
+    arbNetwork.ethBridge.sequencerInbox,
+    l1Provider
+  )
+
+  const delta = 1000
```
DZGoldman, line 139: I don't we need any delta here; i.e., the actual current l1 Block Number will always be greater than (or maybe equal to) the l1BlockNumber that the Sequencer assigns (since the Sequencer has a delayed view of the L1); the batch at which the transaction ultimately gets submitted will happen at an even greater l1 block number still; thus, I don't see any need to buffer negative/past direction.

packages/arb-ts/src/lib/utils/arbProvider.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  const delayBlocks = (await inbox.maxDelayBlocks()).toNumber()
+
+  const startBlock = l1BlockNumber - delta
+  const delayedBlockMax = l1BlockNumber + delayBlocks
+  const currentBlock = await l1Provider.getBlockNumber()
+
+  const endBlock =
```
DZGoldman, line 147: can use `Math.min`

packages/arb-ts/src/lib/utils/arbProvider.ts
```diff
@@ -0,0 +1,246 @@
...
+    arbNetwork.ethBridge.sequencerInbox,
+    l1Provider
+  )
+
+  const delta = 1000
+
+  const delayBlocks = (await inbox.maxDelayBlocks()).toNumber()
+
+  const startBlock = l1BlockNumber - delta
+  const delayedBlockMax = l1BlockNumber + delayBlocks
```
DZGoldman, line 144: This upper bound only applies to transactions submitted in batches by the sequencer ( they indeed can't be included after `l1BlockNumber + delayBlocks`). Transactions that are force-included, however, can be force-included after any arbitrary amount of time. So since this `getBatch`method wants to cover `DelayedInboxForcedEvent`, as is, it would have to search past `l1BlockNumber + delayBlocks`.

DZGoldman: There's a few options here: one is to do the initial event log query only on `SequencerBatchDelivered` and `SequencerBatchDelivered` (using the `delayedBlockMax ` var as is), and then iff those return nil, _and_ the current L1 block is past the point of l1BlockNumber + delayBlocks, we do a second query for `DelayedInboxForced`. Unlovely to have to do 2 event queries, but  the 2nd query actually would only happen for a very unusual case.

DZGoldman:
@yahgwai  Addresses question 1 in a comment; for 2 & 3, I'm not sure off the bat myself; play to look seems to be `/packages/arb-rpc-node/web3/eth.go`, tho it may be easiest to look in the node-eng team here.

(generally looking good — I have more to review still / want to make sure I didn't mess anything in the reorganisation diff, but left some comments)

packages/arb-ts/src/lib/utils/arbProvider.ts
yahgwai: renamed, and added comment

packages/arb-ts/src/lib/utils/arbProvider.ts
yahgwai: removed

packages/arb-ts/src/lib/utils/arbProvider.ts
yahgwai: Updated

yahgwai:
I've some fields that we dont need to surface at the moment - we can investigate their nullability and document them properly later

DZGoldman (approved):
Little thing we can do is add the max_delay values to the network config (so those are basically constant / admin updated) to save on queries. Doesn't have to happen here tho; LGTM

packages/arb-ts/src/lib/dataEntities/arbTransaction.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * The l1 block number that would be used for block.number calls
+   * that occur within this transaction.
+   * See https://developer.offchainlabs.com/docs/time_in_arbitrum
+   */
+  l1BlockNumber: number
+}
+
+export interface FeeStat {
```
fredlacs, line 32: I think this name is confusing, I'd instead call it FeeComponents, or ResourcesCharged, idk
But FeeStats and FeeStat right now is confusing

packages/arb-ts/src/lib/dataEntities/arbTransaction.ts
```diff
@@ -0,0 +1,174 @@
...
+   */
+  returnCode: ReturnCode
+  /**
+   * Arbitrum fee breakdown
+   */
+  feeStats: FeeStats
+  /**
+   * Batch info, populated if an l1 provider is present when fetching the receipt
+   */
+  l1InboxBatchInfo: BatchInfo | null
```
fredlacs, line 167: I believe this is incorrect. l1InboxBatchInfo is only returned if the L2 node explicitly enables it (which most don't and won't)

fredlacs: context https://github.com/OffchainLabs/arbitrum/releases/tag/v1.0.0

packages/arb-ts/src/lib/message/L1Transaction.ts
```diff
@@ -0,0 +1,213 @@
...
+    this.byzantium = tx.byzantium
+    this.type = tx.type
+    this.status = tx.status
+  }
+
+  /**
+   * Get the numbers of any messages created by this transaction
+   * @returns
+   */
+  public getMessageNumbers(): BigNumber[] {
```
fredlacs, line 85: hm I'd argue its cleaner if we change this to `getInboxMessages` and return the event.
Its pretty easy for the consumer to then do: `receipt.getInboxMessages().messageNum`

packages/arb-ts/src/lib/message/L1Transaction.ts
```diff
@@ -0,0 +1,213 @@
...
+    const iface = Inbox__factory.createInterface()
+    const messageDelivered = iface.getEvent('InboxMessageDelivered')
+    const messageDeliveredFromOrigin = iface.getEvent(
+      'InboxMessageDeliveredFromOrigin'
+    )
```
fredlacs, line 90: Is there a particular reason to listen to the inbox event here?
Given this is somewhat general I'd argue in favour for us to instead to check for the Bridge event.
It also includes the message number `messageIndex`
https://github.com/OffchainLabs/arbitrum/blob/005badd0f7beff05c222bf9d6d301564aa2b8f10/packages/arb-bridge-eth/contracts/bridge/interfaces/IBridge.sol#L22-L29

packages/arb-ts/src/lib/message/L1Transaction.ts
```diff
@@ -0,0 +1,213 @@
...
+   * If the messageIndex is supplied the message at that index will be returned.
+   * If no messageIndex is supplied a message will be returned if this transaction only created one message
+   * All other cases throw an error
+   * @param l2SignerOrProvider
+   */
+  public async getL1ToL2Message<T extends SignerOrProvider>(
+    l2SignerOrProvider: T,
+    messageNumberIndex?: number
+  ): Promise<L1ToL2MessageReaderOrWriter<T>>
+  public async getL1ToL2Message<T extends SignerOrProvider>(
```
fredlacs, line 146: Is this needed?
It doesn't actually have a performance advantage compared to `getL1ToL2Messages` 
Just adds some message count validation then returns the desired index

packages/arb-ts/src/lib/message/L1Transaction.ts
```diff
@@ -0,0 +1,213 @@
...
+      )
+
+    return allL1ToL2Messages[messageIndex || 0]
+  }
+
+  /**
+   * Get any deposit events created by this transaction
+   * @returns
+   */
+  public getDepositEvents(): DepositInitiatedEvent['args'][] {
```
fredlacs, line 173: I think we should make it more explicit that this refers to an asset deposit
```suggestion
  public getTokenDepositEvents(): DepositInitiatedEvent['args'][] {
```

packages/arb-ts/src/lib/message/L1Transaction.ts
fredlacs: would be consistent with
`public getDepositEvents(): DepositInitiatedEvent['args'][]`

packages/arb-ts/src/lib/message/L2Transaction.ts
```diff
@@ -0,0 +1,183 @@
...
+    const arbReceipt = await getRawArbTransactionReceipt(
+      l2Provider,
+      this.transactionHash,
+      l1Provider
+    )
+
+    // Data is made available in batches, if the batch info is
+    // available then so is the tx data
+    return !!arbReceipt?.l1InboxBatchInfo
+  }
```
fredlacs, line 166: I don't believe this is correct (mentioned in previous comment about `l1InboxBatchInfo`)
We need to instead do lookups against the L1 provider.
I agree `isDataAvailable` is the most precise way of describing this, but I don't think its the most intuitive for devs consuming the lib

packages/arb-ts/src/lib/message/L2Transaction.ts
```diff
@@ -0,0 +1,183 @@
...
+   * Replaces the wait function with one that returns an L2TransactionReceipt
+   * @param contractTransaction
+   * @returns
+   */
+  public static monkeyPatchWait = (
+    contractTransaction: ContractTransaction
+  ): L2ContractTransaction => {
+    const wait = contractTransaction.wait
+    contractTransaction.wait = async (confirmations?: number) => {
+      const result = await wait(confirmations)
```
fredlacs, line 178: I don't think it makes sense to wait for anything more than 1 confirmation.
I think we should throw an error if the user sets confirmations > 1

packages/arb-ts/src/lib/message/L2Transaction.ts
fredlacs: Will need to use the getBatch method from the arb provider

packages/arb-ts/src/lib/dataEntities/arbTransaction.ts
yahgwai: Updated to FeeStatComponents

packages/arb-ts/src/lib/dataEntities/arbTransaction.ts
yahgwai: I think it is correct, in getArbTransactionReceipt we use an l1 provider to fetch this info, if the provider is there.

packages/arb-ts/src/lib/message/L1Transaction.ts
yahgwai: Updated

packages/arb-ts/src/lib/message/L1Transaction.ts
yahgwai: yeah, defo, updated

packages/arb-ts/src/lib/message/L1Transaction.ts
yahgwai: Yeah, I kinda agree. @DZGoldman mentioned that this is the most normal way of consuming the messages, and so was useful to have.

packages/arb-ts/src/lib/message/L1Transaction.ts
yahgwai: Updated

packages/arb-ts/src/lib/message/L2Transaction.ts
yahgwai: I kinda like the name - dyou have any other suggestions? In the comment we mention that this is referring specifically to the data being available on L1. It's also a more general principle rather than referring to batches.

packages/arb-ts/src/lib/message/L2Transaction.ts
yahgwai: Ignoring it for now

packages/arb-ts/src/lib/dataEntities/arbTransaction.ts
fredlacs: Ah I misinterpreted it, all good

packages/arb-ts/src/lib/message/L1Transaction.ts
fredlacs: We never use this internally, and externally its an extra couple of lines of validation which we let the consumer decide on

PR #2008, opened 2022-01-26
Old db duplicate transactions in batch check

PlasmaPower:
Avoids "invalid nonce" issue when duplicate transactions are attempted to be put in the same block


joshuacolvin0 (approved):
LGTM

BMItr (approved):
LGTM 👌 
please undraft & merge asap after CI is green 🟢 
thx for the nice work!

PR #2009, opened 2022-01-26
Check for duplicate transactions in a batch

PlasmaPower:
Avoids "invalid nonce" issue when duplicate transactions are attempted to be put in the same block. This also matches geth's error message.


joshuacolvin0 (dismissed):
LGTM

joshuacolvin0 (approved):
LGTM

PR #2013, opened 2022-01-28
Added force-inclusion tests

yahgwai:
Also some small changes to execution proof test to get it compile.


packages/arb-bridge-eth/test/sequencerInbox.spec.ts
```diff
@@ -0,0 +1,447 @@
...
+import { initializeAccounts } from './utils'
+import { Event } from '@ethersproject/contracts'
+import { Interface } from '@ethersproject/abi'
+import { BridgeInterface, MessageDeliveredEvent } from '../build/types/Bridge'
+import { Signer } from 'ethers'
+
+const mineBlocks = async (signer: Signer, count: number) => {
+  for (let index = 0; index < count; index++) {
+    await (
+      await signer.sendTransaction({
```
fredlacs, line 39: I was checking out if we use [`provider.send()`](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider-send) to mint blocks with [`evm_mine`](https://hardhat.org/hardhat-network/reference/#evm-mine) instead.
Seems like just as slow for large number of blocks, but the cool thing is that it lets us decide the timestamp of the next block.
We could use that to test the diff ways that block num vs timestamp can stray and how the inbox reacts to that

packages/arb-bridge-eth/test/sequencerInbox.spec.ts
```diff
@@ -0,0 +1,447 @@
...
+      delayedTx3.l1BlockTimestamp,
+      delayedTx3.deliveredMessageEvent.messageIndex,
+      delayedTx3.l1GasPrice,
+      delayedTx3.senderAddr,
+      delayedTx3.deliveredMessageEvent.messageDataHash,
+      delayedTx3.delayedAcc
+    )
+  })
+
+  it('cannot include before max delay', async () => {
```
fredlacs, line 416: cases covered here are solid - i'd just add some logic to check for the block timestamp as there is also the require statement that throws `MAX_DELAY_TIME`
the timestamp distance between blocks isn't spec'ed and is instead left for each node to decide, but we could add a test that covers if for example each block skews by `n seconds`

fredlacs (changes requested):
Looks good, but I think we should also cover the timestamp in this testing

packages/arb-bridge-eth/test/sequencerInbox.spec.ts
yahgwai: updated to use `evm_mine`

packages/arb-bridge-eth/test/sequencerInbox.spec.ts
yahgwai: add a test for delay time

packages/arb-bridge-eth/test/sequencerInbox.spec.ts
```diff
@@ -444,4 +443,40 @@ describe('SequencerInbox', async () => {
...
+      BigNumber.from(10),
+      '0x1010'
+    )
+
+    const maxDelayBlocks = (await sequencerInbox.maxDelayBlocks()).toNumber()
+    // mine a lot of blocks - but use a short time per block
+    // this should mean enough blocks have passed, but not enough time
+    await mineBlocks(maxDelayBlocks + 1, 5)
+
+    await forceIncludeMessages(
```
fredlacs, line 468: hm we could maybe simplify this interface by just expecting the delayedTx object (could use something like `ReturnType<sendDelayedTx>` (not sure if this is valid syntax, but general idea)

fredlacs (dismissed):
LGTM
Flagged 2 small nitpicks but shouldn't be blockers to merging imo

packages/arb-bridge-eth/test/sequencerInbox.spec.ts
yahgwai: Hmm, I think i wont - the interfaces dont match up exactly I kinda want the force include params to mirror the names in the smart contract

fredlacs: yup sounds fair

fredlacs (approved):
LGTM

PR #2014, opened 2022-01-28
apply default gas percent increase to gas price, not maxgas 

DZGoldman:


packages/arb-ts/src/lib/bridge.ts
```diff
@@ -395,26 +399,22 @@ export class Bridge {
 
     let maxGas =
       retryableGasArgs.maxGas ||
-      BridgeHelper.percentIncrease(
```
fredlacs, line 398: I think we should still percent increase the maxGas, as well as gasPrice

fredlacs (changes requested):
Looks good to add the increase to the gasprice, but i think we should also keep it for the maxGas

fredlacs:
We could also add an optional parameter to override this percent increase if you want to be able to better control this in consumers (ie the UI)

PR #2015, opened 2022-01-28
address table class for arb-ts v2

DZGoldman:
mvp


yahgwai (commented):
Would be good to have a chat with you about this, I'm not sure we need it. The class is a thin wrapper around the contract, so I think it would be good to discuss where we hand over responsibility to the underlying contracts, and where arb-ts fits into that.

fredlacs:
closin this for now, we can add in the other approach discussed in a different PR

PR #2022, opened 2022-02-02
Arb ts next  eth deposit bug

yahgwai:


packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -251,9 +252,19 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     if (retryableCreationReceipt.status === 0) {
       return L1ToL2MessageStatus.CREATION_FAILED
     }
+
+    // we have a retryableCreationReceipt, in that case we should have
+    // an autoRedeemReceipt if the transaction contained data, if it didn't
```
fredlacs, line 257:
```suggestion
    // an autoRedeemReceipt if the transaction contained `maxGas > 0 && gasPriceBid > 0`, if it didn't
```

fredlacs (approved):
LGTM, only one small note on when autoredeem is done

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -306,15 +322,18 @@ export class L1ToL2MessageReader extends L1ToL2Message {
       ? await this.getL2TxReceipt()
       : undefined
 
+    const autoRedeemReceiot = await this.getAutoRedeemReceipt()
```
DZGoldman, line 325: typo

packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -251,9 +252,19 @@ export class L1ToL2MessageReader extends L1ToL2Message {
     if (retryableCreationReceipt.status === 0) {
       return L1ToL2MessageStatus.CREATION_FAILED
     }
+
+    // we have a retryableCreationReceipt, in that case we should have
+    // an autoRedeemReceipt if the transaction contained data, if it didn't
+    // we would expect an autoRedeemReceipt and the presence of a retryableCreationReceipt
+    // means the tx has already been redeemed.
+    if (!autoRedeemReceipt) {
```
DZGoldman, line 260: for the  `maxGas === 0 || gasPriceBid === 0` case, we also won't get an autoredeem receipt, so we can't assume it's been "redeemed"; maybe just sticking with returning `NOT_YET_REDEEMED` for eth deposits, and requiring the consumer to understand the eth deposits are a special case, is best? unsure

DZGoldman: optimally we could detect there's there was no encoded call-data and thus the ticket is "irredeemable," and we could return some other terminal, but not "REDEEMED" status, like, say "COMPLETED." But not sure we have a nice way of doing that detection at this point...

DZGoldman (changes requested):
left a comment — we should discuss

packages/arb-ts/src/lib/message/L1ToL2Message.ts
DZGoldman: or maybe say an "isRedeemable" flag that defaults to true, which if false returns status COMPLETED where appropriate (does not spark joy, just thinking)

yahgwai: yeah, how do we detect that there's no call data though?

yahgwai: would it be decoding this data: https://github.com/OffchainLabs/arbitrum/blob/e710c33f4f819bc7a8d01c981a1988a26fdba680/packages/arb-bridge-eth/contracts/bridge/Inbox.sol#L250

yahgwai: @fredlacs  actually pointed out that no call data doesnt mean it's irredeemable, since it could be calling a fallback function on a contract

yahgwai:
Closed in favour of https://github.com/OffchainLabs/arbitrum/pull/2037

PR #2037, opened 2022-02-08
Arb-ts v2 eth deposit status info

yahgwai:


packages/arb-ts/src/lib/message/L1ToL2Message.ts
```diff
@@ -237,54 +237,108 @@ export class L1ToL2MessageReader extends L1ToL2Message {
...
+   * @returns The wait result contains a complete, a status, and optionally the l2TxReceipt.
+   * If complete is true then this message is in the terminal state for it's type. For eth deposits complete is
+   * true if the status is FUNDS_DEPOSITED, EXPIRED or REDEEMED. For all other messages complete is true
+   * only if the status is REDEEMED.
+   */
+  public async wait(
+    isOnlyEthDeposit: boolean,
+    timeout = 900000,
+    confirmations?: number
+  ): Promise<L1ToL2MessageWaitResult & { complete: boolean }> {
```
fredlacs, line 316:
```suggestion
  public async wait(
    isEthDeposit: boolean,
    confirmations?: number,
    timeout?: number
  ): Promise<L1ToL2MessageWaitResult & { complete: boolean }> {
```

A couple things here, i think we should handle timeouts the same way ethers does, [this](https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/providers/src.ts/base-provider.ts#L936) is the reference I found for it, that is to either have a user timeout or wait indefinitely.

fredlacs: I'm also thinking it might be worth having `confirmations` as the first parameter since its usually the expected first param.
That is because someone consuming this in javascript might mis-use and mess it up

```suggestion
  public async wait(
    confirmations = 1,
    isEthDeposit = false,
    timeout = 0
  ): Promise<L1ToL2MessageWaitResult & { complete: boolean }> {
```
Maybe something like this would work

fredlacs (changes requested):

packages/arb-ts/src/lib/message/L1ToL2Message.ts
yahgwai: The problem is I dont think we should have a default for `isEthDeposit`, since otherwise people wont be forced to think about it, and will get the wrong completion status.

yahgwai: Re the timeout, lets have a chat about it. In this case we have a strong expectation of when we think a transaction will be included, so we want to be explain that to the user. That was the purpose of the default before, we could explain this in a comment?

yahgwai: Re-ordered the params in `wait` and `waitForStatus`

DZGoldman (approved):
Looks good, and looks like Fred's comments have been addressed — I'll let @fredlacs do the final approve / merge

PR #2062, opened 2022-02-17
Reconnect ethclient on repeated error

hkalodner:
This patch should hopefully help with issues where ethclient connections get stuck in a broken state

### DIFF
--- a/packages/arb-util/ethutils/client.go
+++ b/packages/arb-util/ethutils/client.go
@@ -78,7 +78,7 @@ func NewRPCEthClient(url string) (*RPCEthClient, error) {
 func (r *RPCEthClient) reconnect() error {
 	r.Lock()
 	defer r.Unlock()
-	if r.errCount < maxErrCount {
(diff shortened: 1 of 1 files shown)


PlasmaPower (dismissed):
LGTM

packages/arb-util/ethutils/client.go
```diff
@@ -78,12 +78,17 @@ func NewRPCEthClient(url string) (*RPCEthClient, error) {
 func (r *RPCEthClient) reconnect() error {
 	r.Lock()
 	defer r.Unlock()
+	if r.errCount < maxErrCount {
```
PlasmaPower, line 81: Hmm.. technically, this (and the set below) need to also be atomic as the error count is atomically written to outside the mutex.

hkalodner: Ah whoops. Good catch, I hadn't been thinking. Fix pushed

PlasmaPower (approved):
LGTM

PR #2063, opened 2022-02-17
Fork node

hkalodner:
This PR includes a number of different changes that came up as I was working. Happy to split it into parts if it makes review/merging easier

- Add a new utility `arb-fork-node`
- Use `ValueLoader` when marshaling to support lazy-loaded core thread
- Add support for pretty printing logs (not currently enabled for node but easy to add from this)

Fork node is started like
```bash
arb-fork-node --dbdir=path/to/db
```

The fork node should be run against the chain state db of a standard mainnet or rinkeby node. It will reset back to the last real state each time it's run. It will also seed test accounts with eth and add the first address of the addresses as an owner

This can be used for testing the effect of arbos upgrades on the full chain among other use cases


packages/arb-rpc-node/cmd/arb-dev-node/arb-dev-node.go
```diff
@@ -96,10 +95,9 @@ func startup() error {
 	walletbalance := fs.Int64("walletbalance", 100, "amount of funds in each wallet (Eth)")
 	arbosPath := fs.String("arbos", "", "ArbOS version")
 	enableFees := fs.Bool("with-fees", false, "Run arbos with fees on")
-	dbDir := fs.String("dbdir", "", "directory to load dev node on. Use tempory if empty")
+	dbDir := fs.String("dbdir", "", "directory to load dev node on. Use temporay if empty")
```
PlasmaPower, line 98: Nice catch, but I'm afraid it's actually "temporary" :laughing:

packages/arb-rpc-node/cmd/arb-fork-node/arb-fork-node.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	type forkInfo struct {
+		LastMessage int64 `json:"last_block"`
+	}
+	var fork forkInfo
+
+	forkFile := filepath.Join(*dbDir, "fork.json")
+	forkData, err := os.ReadFile(forkFile)
+	if err == nil {
+		if err := json.Unmarshal(forkData, &fork); err != nil {
```
PlasmaPower, line 133: Are we sure we want to reorg the chain every time? It seems useful to be able to preserve state across runs. Maybe we could add a flag to preserve the existing state? Or are we thinking people could just start up the dev node on the same db afterwards to continue their chain?

PlasmaPower (commented):
Generally LGTM. We can probably remove the separate log value cache I introduced in arbcore.cpp now.

packages/arb-rpc-node/cmd/arb-dev-node/arb-dev-node.go
hkalodner: 🤣

packages/arb-rpc-node/cmd/arb-fork-node/arb-fork-node.go
hkalodner: Hmm fair. I was generally just modeling after hardhat's mainnet fork mode, but seems nicer to at least have optional persistance. For now I think just adding a `--persist-state` flag which defaults to off makes sense. For simplicity though I think it makes sense to redo all the initialization message every time which just means your balance will just increase since otherwise it could get into weird inconsistent states

PlasmaPower (dismissed):
LGTM

PR #2071, opened 2022-02-23
Arb ts next  docs

yahgwai:
The docs I've added are just an overview, I think we should add 2 more things for better docs:
1. API reference - we could possibly autogen this from the typescript comments?
2. Recipes - Like short concise versions of the tutorials, without any of the key setup boiler plate, to show how our main classes interact.


DZGoldman:
Looks like this includes the changes from https://github.com/OffchainLabs/arbitrum/pull/2037 but there may be some reconciling necessary? (or maybe this could just be merged in lieu of 2037)?

DZGoldman (approved):
md/ docs part of this looks good (we'll talk next steps for docs in general, as discussed)

DZGoldman (approved):
lgtm

PR #2073, opened 2022-02-23
Check parent code for future segments

PlasmaPower:
I believe this change is safe, as we shouldn't ever try to query for a future code segment, but if we did it could result in e.g. a future code segment being exposed to an eth_call on an old block.


hkalodner (approved):
LGTM

PR #2077, opened 2022-02-24
Arb ts next better wait

yahgwai:
Simplifies deposit by removing the need to get the messages before waiting for the l2 transaction


fredlacs (approved):
LGTM

PR #2078, opened 2022-02-24
Arb ts next force include incremental

yahgwai:
To avoid large ethLog requests


packages/arb-ts/src/lib/inbox/inbox.ts
```diff
@@ -183,6 +184,38 @@ export class InboxTools {
       }
     )
 
+    if (events.length !== 0) return events
+    else if (maxedSearchRangeBlocks === maxSearchRangeBlocks) return []
+    else {
+      return await this.getEventsAndIncreaseRange(
+        bridge,
+        searchRangeBlocks * 2,
```
fredlacs, line 192: we could make the `* 2` a default parameter like "growthRate" that defaults to 2
```suggestion
        searchRangeBlocks * 2,
```

packages/arb-ts/src/lib/inbox/inbox.ts
```diff
@@ -183,6 +184,38 @@ export class InboxTools {
...
+    const bridge = Bridge__factory.connect(
+      this.l2Network.ethBridge.bridge,
+      this.l1Provider
+    )
+
+    // events dont become eligible until they pass a delay
+    // find a block range which will emit eligible events
+    const events = await this.getEventsAndIncreaseRange(
+      bridge,
+      100,
```
fredlacs, line 215: why do we hardcode 100 here?
could be a function param with default value like in `searchRangeBlocks`
If we do the `growthRate` change for `getEventsAndIncreaseRange` we should also surface it here with a default value since this function is the user facing one

fredlacs (changes requested):

packages/arb-ts/src/lib/inbox/inbox.ts
yahgwai: updated

packages/arb-ts/src/lib/inbox/inbox.ts
```diff
@@ -183,6 +185,47 @@ export class InboxTools {
...
+   * Find the event of the latest message that can be force include
+   * @param maxSearchRangeBlocks The max range of blocks to search in.
+   * Defaults to 3 * 6545 ( = ~3 days) prior to the first eligble block
+   * @param startSearchRangeBlocks The start range of block to search in.
+   * Moves incrementally up to the maxSearchRangeBlocks. Defaults to 100;
+   * @param rangeMultiplier The multiplier to use when increasing the block range
+   * Defaults to 2.
+   * @returns Null if non can be found.
+   */
+  public async getForceIncludeableEvent(
```
DZGoldman, line 210: sp: "Includable"

packages/arb-ts/src/lib/inbox/inbox.ts
```diff
@@ -152,24 +149,29 @@ export class InboxTools {
...
+    bridge: Bridge,
+    searchRangeBlocks: number,
+    maxSearchRangeBlocks: number,
+    rangeMultiplier: number
+  ): Promise<FetchedEvent<MessageDeliveredEvent>[]> {
     const eFetcher = new EventFetcher(this.l1Provider)
 
     // events dont become eligible until they pass a delay
     // find a block range which will emit eligible events
+    const maxedSearchRangeBlocks = Math.min(
```
DZGoldman, line 169: personal pref, what about `cappedSearchRangeBlocks` (to avoid max/maxed)

DZGoldman (approved):
looks g — just little spelling / naming suggestions

PR #2084, opened 2022-02-28
Do not log errors interacting with already closed broadcaster connection

joshuacolvin0:
Broadcaster works with multiple threads, so it is possible that a connection was closed in one thread but another thread is still trying to send to it.


hkalodner (changes requested):
Would probably be nice to add a utility method for this check since it's in a bunch of places. Otherwise looks great

hkalodner (approved):
LGTM

PR #2100, opened 2022-03-03
Fix watchtower validator creating stake

PlasmaPower:
- [x] add a test that a watchtower with no stake doesn't crash


packages/arb-node-core/staker/staker.go
```diff
@@ -304,7 +305,8 @@ func (s *Staker) Act(ctx context.Context) (*arbtransaction.ArbTransaction, error
 
 	// Don't attempt to create a new stake if we're resolving a node,
 	// as that might affect the current required stake.
-	creatingNewStake := rawInfo == nil && s.builder.TransactionCount() == 0
+	wouldCreateStake := rawInfo == nil && s.builder.TransactionCount() == 0
+	creatingNewStake := wouldCreateStake && effectiveStrategy >= StakeLatestStrategy
```
hkalodner, line 309: I think we also want to avoid underfunded defensive validators from failing here. Optimally they'd be able to function without a stake and then block waiting for funding if they don't have enough when they want to challenge

hkalodner (changes requested):

packages/arb-node-core/staker/staker.go
PlasmaPower: This does that - `DefensiveStrategy < StakeLatestStrategy`, and their effective strategy would only reach `StakeLatestStrategy` and report an issue when a fault exists

hkalodner: Ah of course

joshuacolvin0 (approved):
LGTM

PR #2101, opened 2022-03-04
Improve crossplatform support and typechain generation 

fredlacs:
This PR relies on the changes from https://github.com/OffchainLabs/arb-os/pull/676

We now rely more on the high level functions provided by node and avoid environment variables in the npm scripts to allow for better Windows support (which uses %ENV% instead of $ENV).

The postinstall hooks are removed, and instead arb-ts now compiles the needed packages and specifies where to have the artifacts written to


yahgwai (approved):

yahgwai (approved):

DZGoldman (approved):
lgtm - note that arb-ts will need another version bump now (can do it directly on arb-ts-next)

fredlacs:
lets wait until https://github.com/OffchainLabs/arb-os/pull/676 is approved before mergin this in

gzeoneth:
So I just found that mocha is still broken after [e7b5bfd](https://github.com/OffchainLabs/arbitrum/pull/2101/commits/e7b5bfd54b25731c30a9a8884e4008858f920142) and then noticed that there are some package resolution conflicts. I removed the hardcoded resolution of yargs-parser in package.json, do another `yarn install` and it seems to have fixed it. I am not sure if that would break other stuff tho.

PR #2106, opened 2022-03-07
Fix arb-ts integration test

gzeoneth:
Some changes to make `yarn test:integration` pass with ETH funded L1 and L2 wallets. ERC20 deposited from L1 -> L2 is transfer back to the funding address for withdrawal test case. This created some test case dependency but should be ok for testing purpose.


DZGoldman (approved):
Looks good — once we get the custom token mint-on-l2 merged in, the withdraw test can use one of those tokens (i.e., mint on L2 and withdraw); fine as is for now since the withdraw test gets skipped if no tokens are funded.
lgtm

PR #2107, opened 2022-03-07
L1GatewayRouter refund address entry point

gzeoneth:
Add optional refund address entry point to gateway router smart contracts. One concern is that without changing `L1ArbitrumMessenger.sol` the `excessFeeRefundAddress` and `callValueRefundAddress` must be set to the same address, which allow that address to cancel the retryable to break the L1L2 atomicity.

### DIFF
--- a/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
+++ b/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
@@ -190,46 +190,15 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
         uint256 _gasPriceBid,
         bytes calldata _data
     ) public payable virtual override returns (bytes memory res) {
-        require(isRouter(msg.sender), "NOT_FROM_ROUTER");
-        // This function is set as public and virtual so that subclasses can override
-        // it and add custom validation for callers (ie only whitelisted users)
-        address _from;
-        uint256 seqNum;
-        bytes memory extraData;
-        {
-            uint256 _maxSubmissionCost;
-            if (super.isRouter(msg.sender)) {
-                // router encoded
-                (_from, extraData) = GatewayMessageHandler.parseFromRouterToGateway(_data);
-            } else {
-                _from = msg.sender;
-                extraData = _data;
-            }
-            // user encoded
-            (_maxSubmissionCost, extraData) = abi.decode(extraData, (uint256, bytes));
-            // the inboundEscrowAndCall functionality has been disabled, so no data is allowed
-            require(extraData.length == 0, "EXTRA_DATA_DISABLED");
-
-            require(_l1Token.isContract(), "L1_NOT_CONTRACT");
-            address l2Token = calculateL2TokenAddress(_l1Token);
-            require(l2Token != address(0), "NO_L2_TOKEN_SET");
-
-            _amount = outboundEscrowTransfer(_l1Token, _from, _amount);
-
-            // we override the res field to save on the stack
-            res = getOutboundCalldata(_l1Token, _from, _to, _amount, extraData);
-
-            seqNum = createOutboundTx(
-                _from,
-                _amount,
-                _maxGas,
-                _gasPriceBid,
-                _maxSubmissionCost,
-                res
-            );
-        }
-        emit DepositInitiated(_l1Token, _from, _to, seqNum, _amount);
-        return abi.encode(seqNum);
+        return outboundTransferCustomRefund(
+            _l1Token,
+            _to,
+            _to,
+            _amount,
+            _maxGas,
+            _gasPriceBid,
+            _data
+        );
     }
 
     /**
--- a/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
+++ b/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
@@ -232,6 +232,69 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
         return abi.encode(seqNum);
     }
 
+    /**
+     * @notice Deposit ERC20 token from Ethereum into Arbitrum. If L2 side hasn't been deployed yet, includes name/symbol/decimals data for initial L2 deploy. Initiate by GatewayRouter.
+     * @param _l1Token L1 address of ERC20
+     * @param _to account to be credited with the tokens in the L2 (can be the user's L2 account or a contract)
+     * @param _refundTo account to be credited with the excess gas refund in the L2 (can be the user's L2 account or a contract)
+     * @param _amount Token Amount
+     * @param _maxGas Max gas deducted from user's L2 balance to cover L2 execution
+     * @param _gasPriceBid Gas price for L2 execution
+     * @param _data encoded data from router and user
+     * @return res abi encoded inbox sequence number
+     */
+    //  * @param maxSubmissionCost Max gas deducted from user's L2 balance to cover base submission fee
+    function outboundTransferCustomRefund(
+        address _l1Token,
+        address _to,
+        address _refundTo,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) public payable virtual override returns (bytes memory res) {
+        require(isRouter(msg.sender), "NOT_FROM_ROUTER");
+        // This function is set as public and virtual so that subclasses can override
+        // it and add custom validation for callers (ie only whitelisted users)
+        address _from;
+        uint256 seqNum;
+        bytes memory extraData;
+        {
+            uint256 _maxSubmissionCost;
+            if (super.isRouter(msg.sender)) {
+                // router encoded
+                (_from, extraData) = GatewayMessageHandler.parseFromRouterToGateway(_data);
+            } else {
+                _from = msg.sender;
+                extraData = _data;
+            }
+            // user encoded
+            (_maxSubmissionCost, extraData) = abi.decode(extraData, (uint256, bytes));
+            // the inboundEscrowAndCall functionality has been disabled, so no data is allowed
+            require(extraData.length == 0, "EXTRA_DATA_DISABLED");
+
+            require(_l1Token.isContract(), "L1_NOT_CONTRACT");
+            address l2Token = calculateL2TokenAddress(_l1Token);
+            require(l2Token != address(0), "NO_L2_TOKEN_SET");
+
+            _amount = outboundEscrowTransfer(_l1Token, _from, _amount);
+
+            // we override the res field to save on the stack
+            res = getOutboundCalldata(_l1Token, _from, _to, _amount, extraData);
+
+            seqNum = createOutboundTx(
+                _refundTo,
+                _amount,
+                _maxGas,
+                _gasPriceBid,
+                _maxSubmissionCost,
+                res
+            );
+        }
+        emit DepositInitiated(_l1Token, _from, _to, seqNum, _amount);
+        return abi.encode(seqNum);
+    }
+
     function outboundEscrowTransfer(
         address _l1Token,
         address _from,
--- a/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
+++ b/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
@@ -142,7 +142,8 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
         IERC20(_l1Token).safeTransfer(_dest, _amount);
     }
 
-    function createOutboundTx(
+    function createOutboundTxCustomRefund(
+        address _refundTo,
         address _from,
         uint256, /* _tokenAmount */
         uint256 _maxGas,
@@ -156,9 +157,10 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
         // msg.value is sent, but 0 is set to the L2 call value
         // the eth sent is used to pay for the tx's gas
         return
-            sendTxToL2(
+            sendTxToL2CustomRefund(
                 inbox,
                 counterpartGateway,
+                _refundTo,
                 _from,
                 msg.value, // we forward the L1 call value to the inbox
                 0, // l2 call value 0 by default
@@ -169,6 +171,17 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
                 }),
                 _outboundCalldata
             );
+    }
+
+    function createOutboundTx(
+        address _from,
+        uint256 _tokenAmount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        uint256 _maxSubmissionCost,
+        bytes memory _outboundCalldata
+    ) internal virtual returns (uint256) {
+        return createOutboundTxCustomRefund(_from, _from, _tokenAmount, _maxGas, _gasPriceBid, _maxSubmissionCost, _outboundCalldata);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


gzeoneth:
I kept `outboundTransfer` and `outboundTransferCustomRefund` separated instead of defaulting `_refundTo` in  `outboundTransfer` to allow compatibility with older gateways that did not implement the `outboundTransferCustomRefund` method. It will still fail but the caller can opt to the other method.

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
```diff
@@ -220,7 +252,7 @@ abstract contract L1ArbitrumGateway is L1ArbitrumMessenger, TokenGateway {
             res = getOutboundCalldata(_l1Token, _from, _to, _amount, extraData);
 
             seqNum = createOutboundTx(
-                _from,
+                _refundTo,
```
fredlacs, line 255: this will mess up the event that is emitted.
We could instead add the "refund" address field separate from "user" to [L1ArbitrumMessenger](https://github.com/OffchainLabs/arbitrum/blob/c8ab7a69ce1176aaca551607b8dbc1153d31581f/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/L1ArbitrumMessenger.sol#L74-L75) that is used.

This might create stack size issues, which you can get around by using structs

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -241,6 +241,32 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
         return super.outboundTransfer(_token, _to, _amount, _maxGas, _gasPriceBid, _data);
     }
 
+    function outboundTransferCustomRefund(
```
fredlacs, line 244: seems to have a lot of duplicate logic with `outboundTransfer`, any reason we can't reuse the logic?

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/gateway/GatewayRouter.sol
```diff
@@ -101,6 +101,34 @@ abstract contract GatewayRouter is TokenGateway {
             );
     }
 
+    function outboundTransferCustomRefund(
```
fredlacs, line 104: seems to have repeated logic with `outboundTransfer`

fredlacs (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
fredlacs: Ah just saw your comment above

fredlacs: would be good if we could a comment in the code explaining this

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
gzeoneth: resolved in commit 0f1c0b3637bb8d4dc4de4adf9742060459258fe7

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
gzeoneth: added some comment to [GatewayRouter.sol](https://github.com/OffchainLabs/arbitrum/commit/0f1c0b3637bb8d4dc4de4adf9742060459258fe7#diff-7a73936444469339b89260b9eea33c240dbc90164fc1b18f1acb3bf5c9f2c33b) in commit 0f1c0b3637bb8d4dc4de4adf9742060459258fe7

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol
gzeoneth: stack size issue is avoided by reordering `_refundTo` before `_user` since that is the order we will consume in L1ArbitrumMessenger

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -241,6 +241,32 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
...
+        // when sending a L1 to L2 transaction, we expect the user to send
+        // eth in flight in order to pay for L2 gas costs
+        // this check prevents users from misconfiguring the msg.value
+        (uint256 _maxSubmissionCost, ) = abi.decode(_data, (uint256, bytes));
+
+        // here we don't use SafeMath since this validation is to prevent users
+        // from shooting themselves on the foot.
+        uint256 expectedEth = _maxSubmissionCost + (_maxGas * _gasPriceBid);
+        require(_maxSubmissionCost > 0, "NO_SUBMISSION_COST");
+        require(msg.value == expectedEth, "WRONG_ETH_VALUE");
```
hkalodner, line 264: As opposed to the changes in #2128, I think maybe refactoring this chunk into its own function might make sense since the logic is relatively complex. It's significantly more likely we're going to want to mess with this logic later than the stuff from GatewayRouter.sol, eg. for nitro

hkalodner (changes requested):
Miner refactor suggestion, but otherwise looks good

gzeoneth:
was wrong about the gas comparison, the additional gas was caused by log emitted in InboxMock, the correct numbers are
140776 (before PR) -> 140969 (before optimization) -> 140738 (after optimization)

fredlacs (approved):
LGTM

gzeoneth:
Removed `virtual` keyword from old entrypoint [c68ad80](https://github.com/OffchainLabs/arbitrum/pull/2107/commits/c68ad8093c6586b154f221582b64706c33b545b3) to prevent  inherited contract only override the old entrypoint but not the new one, leading to different bridge logic implemented

gzeoneth:
in retrospect the missing type is due to missing `yarn build` in ci config, will leave it as-is for now

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1WethGateway.sol
```diff
@@ -45,7 +45,8 @@ contract L1WethGateway is L1ArbitrumExtendedGateway {
         l2Weth = _l2Weth;
     }
 
-    function createOutboundTx(
+    function createOutboundTxCustomRefund(
+        address _refundTo,
```
fredlacs, line 49: we added the `refundTo` option but don't use it in `sendTxToL2(`

fredlacs (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1WethGateway.sol
fredlacs: also seems like we don't expose a `function outboundTransferCustomRefund(` here

gzeoneth: > we added the `refundTo` option but don't use it in `sendTxToL2(`

will fix

> also seems like we don't expose a `function outboundTransferCustomRefund(` here like we do in the others, any reason for that?

L1WethGateway doesn't override `outboundTransfer` currently too

fredlacs: yea, we don't override that one since we don't need the reentrancy guard for that one (we know the weth9 code doesn't have a reentrancy)

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -218,29 +218,58 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
...
-        address _token,
-        address _to,
-        uint256 _amount,
+    function _outboundTransferChecks(
         uint256 _maxGas,
         uint256 _gasPriceBid,
-        bytes calldata _data
-    ) public payable override onlyWhitelisted returns (bytes memory) {
+        bytes memory _data
+    ) internal {
```
fredlacs, line 225:
```suggestion
    ) internal view {
```

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1CustomGateway.sol
```diff
@@ -57,15 +57,16 @@ contract L1CustomGateway is L1ArbitrumExtendedGateway, ICustomGateway {
 
     // end of inline reentrancy guard
 
-    function outboundTransfer(
+    function outboundTransferCustomRefund(
```
fredlacs, line 60: we want to add this in, but also keep the `function outboundTransfer(` override here to add in the `nonReentrant` modifier

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ERC20Gateway.sol
```diff
@@ -54,15 +54,16 @@ contract L1ERC20Gateway is L1ArbitrumExtendedGateway {
 
     // end of inline reentrancy guard
 
-    function outboundTransfer(
+    function outboundTransferCustomRefund(
```
fredlacs, line 57: Same as the custom gateway, we want to override both the `outboundTransfer` and the `outboundTransferCustomRefund` to add in `nonReentrant`

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -218,29 +218,58 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
...
 
+    function outboundTransferCustomRefund(
+        address _token,
+        address _refundTo,
+        address _to,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) public payable override onlyWhitelisted returns (bytes memory) {
```
fredlacs, line 264:
```suggestion
    ) public payable override returns (bytes memory) {
```
we don't have a whitelist anymore, we can remove this

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -218,29 +218,58 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
...
+    }
+
+    function outboundTransfer(
+        address _token,
+        address _to,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) public payable override onlyWhitelisted returns (bytes memory) {
```
fredlacs, line 249:
```suggestion
    ) public payable override returns (bytes memory) {
```

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
```diff
@@ -218,29 +218,58 @@ contract L1GatewayRouter is WhitelistConsumer, L1ArbitrumMessenger, GatewayRoute
...
+    function outboundTransferCustomRefund(
+        address _token,
+        address _refundTo,
+        address _to,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) public payable override onlyWhitelisted returns (bytes memory) {
+        // _refundTo is subject to L2 alias rewrite
```
fredlacs, line 265: this is important, we should be add it as a natspec comment to the function

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/gateway/GatewayRouter.sol
```diff
@@ -101,6 +103,34 @@ abstract contract GatewayRouter is TokenGateway {
...
```
fredlacs, line 106: we could add in a natspec comment saying that not all gateways support `outboundTransferCustomRefund` so this method may revert

fredlacs: I don't think this solidity version allows us to inherit natspec comments, but we should also add in to the L1GatewayRouter function natspec

fredlacs (changes requested):

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ERC20Gateway.sol
gzeoneth: nice catch will add

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
gzeoneth: should we drop the `WhitelistConsumer` base contract too? will change storage layout tho we need a `postUpgradeInit` to fix the owner and inbox

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1CustomGateway.sol
gzeoneth: let's consolidate the discussion in [the other thread](https://github.com/OffchainLabs/arbitrum/pull/2107/files#r850202260)

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ERC20Gateway.sol
gzeoneth: On second thought we don't need to have `nonReentrant` in `outboundTransfer` because I already marked it as non-virtual and `outboundTransfer` default to `outboundTransferCustomRefund` [here](https://github.com/OffchainLabs/arbitrum/blob/63cb2f38db044d92a81bafdf1b3a933af8d7a892/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol#L201)

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
fredlacs: hmmm I don't think we should mess with the storage layout, but removing the modifier will reduce gas costs a bit

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ERC20Gateway.sol
fredlacs: yea true, nice

packages/arb-bridge-peripherals/contracts/tokenbridge/libraries/gateway/GatewayRouter.sol
gzeoneth: added some natspec comments

packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1GatewayRouter.sol
gzeoneth: added some natspec comments

PR #2110, opened 2022-03-09
Tracing fixes

hkalodner:
- Correct "callCode" to "callcode" for call type
- Fix "from" address in delegatecall
- Show empty input field in calls with no calldata instead of skipping field
- Properly sort callframes by trace address
- Add cross comparison testing of trace output with openethereum devnet


joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

PR #2114, opened 2022-03-10
Remove unneeded unaliased check for crosschain message

fredlacs:
Also flipped it to apply the alias on the gateway address, instead of undoing it on the msg sender.
Feels more natural and clean (also if we were using immutables this would be better optimised)


gzeoneth (approved):
looks good

PR #2140, opened 2022-03-28
Refactor core errors

joshuacolvin0:
* Properly treat logcursor and inboxmessage errors as permanent
* Fix race conditions

### DIFF
--- a/packages/arb-avm-cpp/cmachine/arbcore.go
+++ b/packages/arb-avm-cpp/cmachine/arbcore.go
@@ -82,11 +82,12 @@ func (ac *ArbCore) MachineMessagesRead() *big.Int {
 
 func (ac *ArbCore) MessagesStatus() (core.MessageStatus, error) {
 	defer runtime.KeepAlive(ac)
+	statusRaw := C.arbCoreMessagesStatus(ac.c)
+	status := core.MessageStatus(int(statusRaw))
+
 	if err := ac.CheckError(); err != nil {
 		return 0, err
 	}
-	statusRaw := C.arbCoreMessagesStatus(ac.c)
-	status := core.MessageStatus(int(statusRaw))
 	return status, nil
 }
 
--- a/packages/arb-avm-cpp/cmachine/arbcore.go
+++ b/packages/arb-avm-cpp/cmachine/arbcore.go
@@ -82,13 +82,11 @@ func (ac *ArbCore) MachineMessagesRead() *big.Int {
 
 func (ac *ArbCore) MessagesStatus() (core.MessageStatus, error) {
 	defer runtime.KeepAlive(ac)
+	if err := ac.CheckError(); err != nil {
+		return 0, err
+	}
 	statusRaw := C.arbCoreMessagesStatus(ac.c)
 	status := core.MessageStatus(int(statusRaw))
-	if status == core.MessagesError {
-		cStr := C.arbCoreMessagesClearError(ac.c)
-		defer C.free(unsafe.Pointer(cStr))
-		return core.MessagesError, errors.New(C.GoString(cStr))
-	}
 	return status, nil
 }
 
@@ -511,8 +509,7 @@ func (ac *ArbCore) LogsCursorRequest(cursorIndex *big.Int, count *big.Int) error
 
 	status := C.arbCoreLogsCursorRequest(ac.c, unsafeDataPointer(cursorIndexData), unsafeDataPointer(countData))
 	if status == 0 {
-		err := ac.LogsCursorCheckError(cursorIndex)
-		if err != nil {
+		if err := ac.CheckError(); err != nil {
 			return err
 		}
 
@@ -527,8 +524,7 @@ func (ac *ArbCore) LogsCursorGetLogs(cursorIndex *big.Int) (*big.Int, []core.Val
 	cursorIndexData := math.U256Bytes(cursorIndex)
 	result := C.arbCoreLogsCursorGetLogs(ac.c, unsafeDataPointer(cursorIndexData))
 	if result.found == 0 {
-		err := ac.LogsCursorCheckError(cursorIndex)
-		if err != nil {
+		if err := ac.CheckError(); err != nil {
 			return nil, nil, nil, err
 		}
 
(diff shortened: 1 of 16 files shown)


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -375,25 +367,21 @@ InitializeResult ArbCore::applyConfig() {
...
-    return {status, false};
+    return {rocksdb::Status::OK(), false};
 }
 
 InitializeResult ArbCore::initialize(const LoadedExecutable& executable) {
     // Use latest existing checkpoint
 
     auto config_result = applyConfig();
-    if (config_result.finished) {
+    if (config_result.finished || config_result.status.ok()) {
```
hkalodner, line 384: Should this be `!config_result.status.ok()`?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1578,6 +1572,7 @@ void ArbCore::operator()() {
                     core_machine->machine_state.output.arb_gas_used);
                 if (!status.ok()) {
                     core_error_string = status.ToString();
+                    core_error = true;
```
hkalodner, line 1575: I think it'd be cleaner to have a `setCoreError()` function that sets both these fields to make sure we always set them together

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1490,9 +1481,10 @@ void ArbCore::operator()() {
                     auto add_status = addMessages(message_data, cache);
                     if (!add_status.status.ok()) {
                         core_error_string = add_status.status.ToString();
-                        message_data_status = MESSAGES_ERROR;
+                        core_error = true;
                         std::cerr << "ArbCore addMessages error: "
                                   << core_error_string << "\n";
+                        break;
```
hkalodner, line 1487: The breaks make me a bit nervous since we could accidentally have them in a nested for loop and not exit. Might be worth switching over to `return` with a scope guard approach. Something like
```c++
    class CoreShutdown {
        ArbCore &core;
        CoreShutdown(ArbCore& core): core(core_) {}
        ~CoreShutdown() { core.abortThread(); }
    };
    CoreShutdown shutdown{*this};
```

One place where a golang `defer` would be simpler

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1801,11 +1797,17 @@ void ArbCore::operator()() {
...
+                    if (!status) {
+                        // error already logged
+                        arbcore_abort = true;
+                        break;
+                    }
                 }
             }
 
-            if (machine_idle && message_data_status != MESSAGES_READY) {
+            if (!arbcore_abort && machine_idle &&
```
hkalodner, line 1809: Could we just check for `arbcore_error` here? I don't see anywhere we set `arbcore_abort` that isn't an error. Also if we switch to just returning on error as suggested above, we might not need to check either

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -3714,11 +3715,9 @@ bool ArbCore::logsCursorConfirmReceived(size_t cursor_index) {
         logs_cursors[cursor_index].reorg_mutex);
 
     if (logs_cursors[cursor_index].status != DataCursor::DELIVERED) {
-        logs_cursors[cursor_index].error_string =
-            "logsCursorConfirmReceived called at wrong state";
         std::cerr << "logsCursorConfirmReceived called at wrong state: "
                   << logs_cursors[cursor_index].status << "\n";
-        logs_cursors[cursor_index].status = DataCursor::ERROR;
+        abortThread();
```
hkalodner, line 3720: Isn't this just an API usage error that should be recoverable? Not sure we want to abort here vs just returning an error in the API

packages/arb-avm-cpp/cmachine/arbcore.go
```diff
@@ -82,13 +82,11 @@ func (ac *ArbCore) MachineMessagesRead() *big.Int {
 
 func (ac *ArbCore) MessagesStatus() (core.MessageStatus, error) {
 	defer runtime.KeepAlive(ac)
+	if err := ac.CheckError(); err != nil {
+		return 0, err
+	}
 	statusRaw := C.arbCoreMessagesStatus(ac.c)
```
hkalodner, line 88: Is there a possible race condition here where there could be no error before the `arbCoreMessagesStatus` call, but then an error occurs after that check? It might make sense to check for an error again after

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: This is actually the intended.  We only need to continue with initialize if applyConfig returns an error which means that the database isn't properly setup.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Will handle refactor in separate PR

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: seems reasonable

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: will handle in separate PR with above.

packages/arb-avm-cpp/cmachine/arbcore.go
joshuacolvin0: I figured if that happened, error would just be caught next time.  Moving `CheckError` after the message status check covers all eventualities.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Leaving it as is.  This should never happen as long as the API is followed, and it isn't always clear what the safest way to clear the error would be.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #2143, opened 2022-03-29
Refactor arbcore thread

joshuacolvin0:
Move body of arbcore thread loop into separate function for easier state management

### DIFF
--- a/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
+++ b/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
@@ -130,8 +130,10 @@ std::string ArbCore::getErrorString() {
 }
 
 void ArbCore::setCoreError(const std::string& message) {
-    core_error_string = message;
-    core_error = true;
+    if (!core_error) {
+        core_error_string = message;
+        core_error = true;
+    }
 }
 
 bool ArbCore::startThread() {
@@ -151,11 +153,10 @@ void ArbCore::abortThread() {
 #ifdef __linux__
         core_pthread = std::nullopt;
 #endif
-        arbcore_abort = true;
+        setCoreError("aborting thread");
         core_thread->join();
         core_thread = nullptr;
     }
-    arbcore_abort = false;
 }
 
 // deliverMessages sends messages to core thread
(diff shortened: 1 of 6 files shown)


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1409,404 +1409,379 @@ void ArbCore::printCoreThreadBacktrace() {
...
+                        tx, std::nullopt, target_gas, true);
+                    if (std::holds_alternative<rocksdb::Status>(
+                            current_execution)) {
+                        std::cerr
+                            << "Error loading profile machine number " << i
+                            << ": "
+                            << std::get<rocksdb::Status>(current_execution)
+                                   .ToString()
+                            << std::endl;
+                        return false;
```
hkalodner, line 1646: This changes the behavior since before it would just break out of the inner loop

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1409,404 +1409,379 @@ void ArbCore::printCoreThreadBacktrace() {
...
+
+    if (coreConfig.database_save_interval > 0) {
+        thread_data.next_rocksdb_save_timepoint =
+            std::chrono::steady_clock::now() +
+            std::chrono::seconds(coreConfig.database_save_interval);
+        std::filesystem::create_directories(thread_data.save_rocksdb_path);
+    }
+
+    try {
+        while (!arbcore_abort && !core_error) {
```
hkalodner, line 1782: Can we remove `arbcore_abort` now?

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: that's fair, when testing possible that only a few of the machines fail

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #2145, opened 2022-03-30
Configuration cleanups

joshuacolvin0:
* Don't use wallet if in watchtower strategy
* Parse node type and validator strategy as enum for more consistent type checking.

### DIFF
--- a/packages/arb-node-core/staker/staker.go
+++ b/packages/arb-node-core/staker/staker.go
@@ -39,15 +39,6 @@ import (
 
 var logger = arblog.Logger.With().Str("component", "staker").Logger()
 
-type Strategy uint8
-
-const (
-	WatchtowerStrategy Strategy = iota
-	DefensiveStrategy
-	StakeLatestStrategy
-	MakeNodesStrategy
-)
-
 type nodeAndHash struct {
 	id   core.NodeID
 	hash common.Hash
@@ -56,7 +47,7 @@ type nodeAndHash struct {
 type Staker struct {
 	*Validator
 	activeChallenge         *challenge.Challenger
-	strategy                Strategy
+	strategy                configuration.ValidatorStrategy
 	fromBlock               int64
 	baseCallOpts            bind.CallOpts
 	auth                    transactauth.TransactAuth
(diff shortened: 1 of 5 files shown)


packages/arb-rpc-node/cmd/arb-node/arb-node.go
```diff
@@ -181,20 +181,22 @@ func startup() error {
...
 			badConfig = true
 			fmt.Println("Aggregator node needs --node.aggregator.inbox-address")
 		}
-	} else if strings.EqualFold(config.Node.Type, "sequencer") {
+	} else if config.Node.Type() == configuration.SequencerNodeType {
 		// Sequencer always waits
 		config.WaitToCatchUp = true
-	} else if strings.EqualFold(config.Node.Type, "validator") {
-		// Validator always waits
-		config.WaitToCatchUp = true
```
PlasmaPower, line 194: Just noting that WaitToCatchUp was removed but I'm assuming it was intentional, since I'm not sure it makes sense for validators

packages/arb-node-core/staker/staker.go
```diff
@@ -247,40 +238,40 @@ func (s *Staker) Act(ctx context.Context) (*arbtransaction.ArbTransaction, error
...
+				effectiveStrategy = configuration.StakeLatestStrategy
 			}
 		} else {
 			logger.Info().Msg("defensive validator staked past incorrect node; waiting here")
 			s.bringActiveUntilNode = nil
 		}
 		s.inactiveLastCheckedNode = nil
 	}
-	if effectiveStrategy <= DefensiveStrategy && s.inactiveLastCheckedNode != nil {
+	if !effectiveStrategy.IsActive() && s.inactiveLastCheckedNode != nil {
```
PlasmaPower, line 257: I think this condition accidentally changed from including DefensiveStrategy to not.

PlasmaPower (changes requested):

packages/arb-rpc-node/cmd/arb-node/arb-node.go
joshuacolvin0: Correct, doesn't make any sense for validators to wait

PR #2146, opened 2022-04-01
Treat error in addMessages as non-fatal

joshuacolvin0:


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -110,13 +110,28 @@ bool ArbCore::machineIdle() {
...
-    if (!core_error && current_status != MESSAGES_READY) {
+    if (current_status != MESSAGES_ERROR && current_status != MESSAGES_READY) {
         message_data_status = MESSAGES_EMPTY;
     }
+
     return current_status;
 }
 
+std::string ArbCore::messagesClearError() {
+    auto lock = std::shared_lock<std::shared_mutex>(message_data_error_mutex);
```
PlasmaPower, line 123: Doesn't this need to be a unique lock as it mutates the `message_data_error_string`? If so, and we remove the lock in `messagesStatus`, we should probably make this a regular non-shared mutex.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -110,13 +110,28 @@ bool ArbCore::machineIdle() {
 }
 
 ArbCore::message_status_enum ArbCore::messagesStatus() {
+    auto lock = std::shared_lock<std::shared_mutex>(message_data_error_mutex);
```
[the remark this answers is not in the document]
joshuacolvin0: as discussed, mutex can be removed

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: as discussed, mutex can be removed

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
```diff
@@ -166,6 +166,11 @@ class ArbCore {
     // Core thread inbox status input/output. Core thread will update if and
     // only if set to MESSAGES_READY
     std::atomic<message_status_enum> message_data_status{MESSAGES_EMPTY};
+    std::string message_data_error_string;
+    // Message_data_error_mutex non-arbcore threads should acquire read mutex
+    // when checking for MESSAGES_ERROR, arbcore thread needs to acquire write
+    // mutex only when setting MESSAGES_ERROR
+    std::shared_mutex message_data_error_mutex;
```
PlasmaPower, line 173: I think this field can be removed now.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -111,12 +111,25 @@ bool ArbCore::machineIdle() {
...
 }
 
+std::string ArbCore::messagesClearError() {
+    if (message_data_status != ArbCore::MESSAGES_ERROR) {
+        return "";
+    }
+
+    message_data_status = MESSAGES_EMPTY;
+    auto str = message_data_error_string;
+    message_data_error_string.clear();
```
PlasmaPower, line 128: We should clear `message_data_error_string` before setting `message_data_status` back to empty, to ensure that another error isn't set inbetween the status being set and the error being cleared.

PlasmaPower (changes requested):

packages/arb-avm-cpp/data_storage/include/data_storage/arbcore.hpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

PlasmaPower (approved):
LGTM

PR #2148, opened 2022-04-04
Refactor build script to use bash and have shebang

fredlacs:


DZGoldman (approved):
should we bump/publish the 2 packages here too?

DZGoldman (approved):
lgtm

PR #2166, opened 2022-04-11
Add moved notice to arb-ts readme and delete source files

fredlacs:
Should we also delete the source files here since they are available in the sdk repo?

I think we should still keep the `packages/arb-ts/README.md` so people can find this and know to look into the sdk repo instead


gzeoneth:
or we can add some log spam and deprecate later just in case someone is using it?

gzeoneth (approved):

fredlacs:
> or we can add some log spam and deprecate later just in case someone is using it?

npm should give a deprecation warning when users install the package

gzeoneth (approved):

PR #2179, opened 2022-04-20
Fix parameter handling

joshuacolvin0:
* Remove unnecessary and broken check from pruning mode
* Simplify private key and validator smart contract wallet generation.  Now only does one at a time, but simpler and less error prone
* Fix `go vet` issue with unit test


hkalodner (approved):
LGTM

PR #2188, opened 2022-04-22
Remove default strategy, improve messaging on error

joshuacolvin0:


packages/arb-rpc-node/cmd/arb-node/arb-node.go
```diff
@@ -179,11 +177,12 @@ func startup() error {
 		fmt.Println("Missing --rollup.machine.filename")
 	}
 
+	var message string
 	var rpcMode web3.RpcMode
 	if config.Node.Type() == configuration.ForwarderNodeType {
 		if config.Node.Forwarder.Target == "" {
 			badConfig = true
-			fmt.Println("Forwarder node needs --node.forwarder.target")
+			message = "Forwarder node needs --node.forwarder.target"
```
PlasmaPower, line 185: Could we just return an error here instead of setting badConfig and message? Or alternatively, unify them into a badConfigError variable or such?

joshuacolvin0: Excellent

PlasmaPower (approved):
LGTM

PR #2190, opened 2022-04-24
Fix nitro documentation

joshuacolvin0:
Fix typo.  classic node used `--node.fowarder.target`, nitro uses `--node.forwarding-target`


PlasmaPower:
One other flag doesn't exist, `--node.type`. I think we should just remove that one from this document.

PlasmaPower (approved):
LGTM

PR #2193, opened 2022-04-26
Handle core machine thread errors

joshuacolvin0:
* Handle exceptions in machinethread
* Retry with progressively older database checkpoint when machine error or invalid machine
* To avoid any race conditions, ensure that database changes are committed before saving core thread to cache

### DIFF
--- a/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
+++ b/packages/arb-avm-cpp/data_storage/src/arbcore.cpp
@@ -381,7 +381,7 @@ InitializeResult ArbCore::applyConfig() {
     }
 
     if (!reorg_done) {
-        return {reorgToLastMessage(cache), false};
+        return {reorgToLastCheckpoint(cache), false};
     }
 
     return {rocksdb::Status::OK(), false};
@@ -655,12 +655,28 @@ rocksdb::Status ArbCore::saveAssertion(ReadWriteTransaction& tx,
     return rocksdb::Status::OK();
 }
 
-rocksdb::Status ArbCore::reorgToLastMessage(ValueCache& cache) {
-    std::cerr << "Reloading chain to the last message saved"
+rocksdb::Status ArbCore::reorgToLastCheckpoint(ValueCache& cache) {
+    std::cerr << "Reloading chain to the last checkpoint saved"
               << "\n";
 
     return reorgCheckpoints([&](const MachineOutput&) { return true; }, true,
                             cache);
+}
+
+rocksdb::Status ArbCore::reorgToPenultimateCheckpoint(ValueCache& cache) {
+    std::cerr << "Reloading chain to the penultimate checkpoint saved"
+              << "\n";
+
+    return reorgCheckpoints([&](const MachineOutput&) { return true; }, true,
+                            cache);
+    return reorgCheckpoints(
+        [&](const MachineOutput& output) {
+            static int count = 0;
+            count++;
+            // Skip first entry
+            return count > 1;
+        },
+        false, cache);
 }
 
 rocksdb::Status ArbCore::reorgToL1Block(const uint256_t& l1_block_number,
@@ -1410,15 +1426,34 @@ void ArbCore::printCoreThreadBacktrace() {
 
 bool ArbCore::threadBody(ThreadDataStruct& thread_data) {
     bool isMachineValid;
-    {
+    if (core_machine->status() == MachineThread::MACHINE_ERROR) {
+        isMachineValid = false;
+        core_machine->clearError();
+    } else {
         ReadTransaction tx(data_storage);
         isMachineValid = isValid(tx, core_machine->getReorgData());
     }
     if (!isMachineValid) {
-        std::cerr << "Core thread operating on invalid machine. Rolling back."
-                  << std::endl;
-        assert(false);
-        auto status = reorgToMessageCountOrBefore(0, false, thread_data.cache);
+        if (coreConfig.thread_max_failure_count != 0 &&
+            thread_data.thread_failure_count >
+                coreConfig.thread_max_failure_count) {
+            std::cerr << "Core thread had too many errors, aborting"
+                      << std::endl;
+            return false;
+        }
+        rocksdb::Status status;
+        if (thread_data.thread_failure_count == 0) {
+            std::cerr << "Core thread operating on invalid machine, "
+                         "loading last saved checkpoint"
+                      << std::endl;
+            status = reorgToLastCheckpoint(thread_data.cache);
+        } else {
+            std::cerr << "Core thread operating on invalid machine, "
+                         "loading penultimate saved checkpoint, failure_count: "
+                      << thread_data.thread_failure_count << std::endl;
+            status = reorgToPenultimateCheckpoint(thread_data.cache);
+        }
+        thread_data.thread_failure_count++;
         if (!status.ok()) {
             std::cerr << "Error in core thread calling "
                          "reorgToMessageCountOrBefore: "
@@ -1489,13 +1524,15 @@ bool ArbCore::threadBody(ThreadDataStruct& thread_data) {
 
     // Check machine thread
     if (core_machine->status() == MachineThread::MACHINE_ERROR) {
-        setCoreError(core_machine->getErrorString());
         std::cerr << "AVM machine stopped with error: " << core_error_string
                   << "\n";
-        return false;
+
+        // Retry with fresh machine from database
+        return true;
     }
 
     if (core_machine->status() == MachineThread::MACHINE_SUCCESS) {
+        thread_data.thread_failure_count = 0;
         std::chrono::time_point<std::chrono::steady_clock>
             begin_machine_success_timepoint;
         std::chrono::time_point<std::chrono::steady_clock> output_timepoint;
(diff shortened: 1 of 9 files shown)


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1410,15 +1426,34 @@ void ArbCore::printCoreThreadBacktrace() {
...
+        if (thread_data.thread_failure_count == 0) {
+            std::cerr << "Core thread operating on invalid machine, "
+                         "loading last saved checkpoint"
+                      << std::endl;
+            status = reorgToLastCheckpoint(thread_data.cache);
+        } else {
+            std::cerr << "Core thread operating on invalid machine, "
+                         "loading penultimate saved checkpoint, failure_count: "
+                      << thread_data.thread_failure_count << std::endl;
+            status = reorgToPenultimateCheckpoint(thread_data.cache);
```
PlasmaPower, line 1454: This might be overkill, but you could have a reorgToNthCheckpoint method which would also remove the branching here

PlasmaPower (changes requested):

packages/arb-util/configuration/configuration.go
```diff
@@ -850,6 +851,7 @@ func AddCore(f *flag.FlagSet, maxExecutionGas int) {
 	f.Duration("core.cache.timed-expire", 20*time.Minute, "length of time to hold L2 blocks in arbcore timed memory cache")
 
 	f.Int("core.add-messages-max-failure-count", 10, "number of add messages failures before exiting program")
+	f.Int("core.thread-max-failure-count", 0, "number of core thread failures before exiting program, 0 to ignore errors")
```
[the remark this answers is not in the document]
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: The it would have the potential of skipping more than a single checkpoint at a time.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1620,6 +1692,12 @@ bool ArbCore::threadBody(ThreadDataStruct& thread_data) {
                       << "\n";
             return false;
         }
+
+        if (machine_to_cache) {
+            // Add to cache now that database changes are committed
+            combined_machine_cache.timedAdd(std::move(machine_to_cache));
+        }
```
hkalodner, line 1699: Flagging this since I'm totally sure if we still want this change with the other change in @PlasmaPower's PR, but looks reasonable either way

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1199,9 +1239,8 @@ rocksdb::Status ArbCore::reorgCheckpoints(
         last_machine = std::make_unique<Machine>(*core_machine);
     }
 
-    // Checkpoint was saved at sideload, attempt to continue running
+    // Remove any extra messages left in machine from reorg
     core_machine->machine_state.context.clearInboxMessages();
-    core_machine->continueRunningMachine(true);
```
hkalodner, line 1204: Not sure why this was removed?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -655,14 +655,30 @@ rocksdb::Status ArbCore::saveAssertion(ReadWriteTransaction& tx,
...
                             cache);
 }
 
+rocksdb::Status ArbCore::reorgToPenultimateCheckpoint(ValueCache& cache) {
+    std::cerr << "Reloading chain to the penultimate checkpoint saved"
+              << "\n";
+
+    return reorgCheckpoints([&](const MachineOutput&) { return true; }, true,
+                            cache);
+    return reorgCheckpoints(
```
PlasmaPower, line 672: I'm guessing the first return should be removed

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -655,14 +655,30 @@ rocksdb::Status ArbCore::saveAssertion(ReadWriteTransaction& tx,
...
+        [&](const MachineOutput& output) {
+            static int count = 0;
+            count++;
+            // Skip first entry
+            return count > 1;
```
PlasmaPower, line 677: I think `count` will be kept across multiple invocations of `reorgToPenultimateCheckpoint`, though I'm not entirely sure. A simpler solution might be to declare `int count = 0;` outside the lambda.

PlasmaPower (changes requested):

hkalodner (changes requested):

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
PlasmaPower: It shouldn't be necessary, but I also don't think it can hurt anything

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: advanceCoreToTarget already called continueRunningMachine() in a loop until there were no sideloads, so this call is not needed

PlasmaPower: I might be misremembering that code, but I think it sets a gas limit to ensure we don't run into reorg'd out messages. I think we have the continueRunningMachine after clearing inbox messages to finish up any extra processing.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1408,25 +1431,52 @@ void ArbCore::printCoreThreadBacktrace() {
...
     bool isMachineValid;
-    {
+    if (core_machine->status() == MachineThread::MACHINE_ERROR) {
+        std::cerr << "Attempting to recover from AVM machine error: "
+                  << core_error_string << "\n";
+        isMachineValid = false;
+        core_machine->clearError();
+    } else {
         ReadTransaction tx(data_storage);
         isMachineValid = isValid(tx, core_machine->getReorgData());
```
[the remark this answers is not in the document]
joshuacolvin0: sounds good to me

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1408,25 +1431,52 @@ void ArbCore::printCoreThreadBacktrace() {
...
     }
     if (!isMachineValid) {
-        std::cerr << "Core thread operating on invalid machine. Rolling back."
+        if (coreConfig.thread_max_failure_count != 0 &&
```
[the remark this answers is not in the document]
joshuacolvin0: That won't work, I'll just remove support for allowing zero to mean unlimited errors

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: probably doesn't matter, but I'll leave it

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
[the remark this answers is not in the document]
hkalodner: Ah didn't realize that was the intention. Yeah I think avoiding special casing is nice here

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: interesting, I'll put it back in with comment to that effect

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1408,25 +1431,52 @@ void ArbCore::printCoreThreadBacktrace() {
...
+                      << std::endl;
+            return false;
+        }
+        rocksdb::Status status;
+        std::cerr << "Core thread operating on invalid machine, "
+                     "loading last saved checkpoint"
                   << std::endl;
-        assert(false);
-        auto status = reorgToMessageCountOrBefore(0, false, thread_data.cache);
+        status = reorgToLastCheckpoint(cache);
```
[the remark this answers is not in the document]
joshuacolvin0: no

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -655,14 +655,30 @@ rocksdb::Status ArbCore::saveAssertion(ReadWriteTransaction& tx,
...
 
     return reorgCheckpoints([&](const MachineOutput&) { return true; }, true,
                             cache);
 }
 
+rocksdb::Status ArbCore::reorgToPenultimateCheckpoint(ValueCache& cache) {
+    std::cerr << "Reloading chain to the penultimate checkpoint saved"
+              << "\n";
+
+    return reorgCheckpoints([&](const MachineOutput&) { return true; }, true,
```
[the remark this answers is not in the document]
joshuacolvin0: :+1:

hkalodner (approved):
LGTM

PR #2202, opened 2022-04-27
Ensure sane validator settings

joshuacolvin0:
Terminate program If validator watchtower mode and only-create-wallet-contract set.  This is because watchtower does not need wallet, but can't create wallet contract without wallet


PlasmaPower (approved):
LGTM

PR #2203, opened 2022-04-29
Log some broadcast client errors only as warnings

joshuacolvin0:


packages/arb-util/broadcastclient/broadcastclient.go
```diff
@@ -160,6 +160,8 @@ func (bc *BroadcastClient) startBackgroundReader(ctx context.Context, messageRec
 				}
 				if strings.Contains(err.Error(), "i/o timeout") {
 					logger.Error().Str("feed", bc.websocketUrl).Msg("Server connection timed out without receiving data")
+				} else if strings.Contains(err.Error(), "EOF") {
```
PlasmaPower, line 163: Can we use `errors.Is(err, io.EOF) || errors.Is(err, io.UnexpectedEOF)`? That should be a bit more stable. This also applies to the other places where we check for specific errors.

joshuacolvin0: Sounds good.  All the other `strings.Contains(err` code is either checking errors from third party libraries that don't provide the error object for comparing, or the string is set in C++ code.

PlasmaPower: Ah, that's too bad. Thanks for changing this one though!

PlasmaPower (approved):
LGTM

PR #2208, opened 2022-04-30
While executing machine, periodically `sleep(0)`

joshuacolvin0:
When handling a large number of concurrent executionCursor requests, the tight machine loops can hog all the processors, preventing other things from being handled, like keeping up with the chain.  Add a periodic `sleep(0)` which will allow more of a fair scheduling across all of the threads.

### DIFF
--- a/packages/arb-avm-cpp/avm/src/machine.cpp
+++ b/packages/arb-avm-cpp/avm/src/machine.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <iostream>
+#include <thread>
 
 #include <avm/inboxmessage.hpp>
 #include <avm/machine.hpp>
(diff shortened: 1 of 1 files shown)


packages/arb-avm-cpp/avm/src/machine.cpp
```diff
@@ -77,11 +78,14 @@ Assertion Machine::run() {
...
     while (true) {
-        if (delayAbortCheckCounter >= 100) {
+        if (delayAbortCheckCounter >= 1000000) {
             if (is_aborted.load(std::memory_order_relaxed)) {
                 break;
             }
             delayAbortCheckCounter = 0;
+
+            // Allow other threads to run
+            std::this_thread::sleep_for(std::chrono::seconds(0));
```
PlasmaPower, line 88: I believe the standard way to do this is `std::this_thread::yield();`

PlasmaPower (changes requested):

packages/arb-avm-cpp/avm/src/machine.cpp
joshuacolvin0: :+1:

PlasmaPower (approved):
LGTM

PR #2210, opened 2022-05-02
export state to nitro

PlasmaPower:


packages/arb-rpc-node/nitroexport/rpcserver.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return &ExportRpcServer{
+		db:         db,
+		arbcore:    arbcore,
+		pathPrefix: pathPrefix,
+	}, nil
+}
+
+func (r *ExportRpcServer) ExportBlocks(ctx context.Context, blockNumber rpc.BlockNumber) error {
+	return r.db.FillerUp(ctx, uint64(blockNumber))
```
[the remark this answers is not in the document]
tsahee: actually, replace block target with whatever came from last rpc call - so it'd be possible to cancel

packages/arb-rpc-node/nitroexport/rpcserver.go
```diff
@@ -0,0 +1,54 @@
...
+	blocks, err := r.db.BlocksExported()
+	if err != nil {
+		return blocks, err
+	}
+	return blocks, r.db.CurrentError()
+}
+
+func (r *ExportRPCServer) ExportState(ctx context.Context, blockNumber rpc.BlockNumber) error {
+	// TODO: stop state export on context cancel
+	return ExportState(r.arbcore, uint64(blockNumber), path.Join(r.pathPrefix, "state", hexutil.EncodeUint64(uint64(blockNumber))))
```
PlasmaPower, line 53: I don't think this should stop the export on context cancel as it'll take multiple minutes and there's risk of TCP connections timing out for instance

packages/arb-rpc-node/nitroexport/rpcserver.go
```diff
@@ -0,0 +1,54 @@
...
+	db.Start(ctx)
+	return &ExportRPCServer{
+		db:         db,
+		arbcore:    arbcore,
+		pathPrefix: pathPrefix,
+	}, nil
+}
+
+func (r *ExportRPCServer) ExportHistory(blockNumber rpc.BlockNumber) error {
+	r.db.UpdateTarget(uint64(blockNumber))
```
PlasmaPower, line 39: This (and ExportState) should accept "latest" as the block number (I think it's translated into rpc.LatestBlockNumber or something)

packages/arb-rpc-node/nitroexport/rpcserver.go
PlasmaPower: We should emit a log line when the export is done too

packages/arb-rpc-node/nitroexport/crossdb.go
```diff
@@ -124,6 +137,66 @@ func (c *CrossDB) importBlock(ctx context.Context, blockNumber uint64) error {
 	return err
 }
 
+func (c *CrossDB) OutboxBatchesExported() (uint64, error) {
+	return c.ethDB.Ancients()
```
PlasmaPower, line 141: This function seems to be never called but I think it returns the number of blocks exported, not outbox batches. Looks like it can be removed in favor of `BatchesExported`

PlasmaPower (commented):
LGTM though I can't approve my own PR

hkalodner (approved):
LGTM

joshuacolvin0 (approved):
LGTM

hkalodner (approved):
LGTM

hkalodner (approved):
LGTM

yspk:
How to use

PR #2216, opened 2022-05-03
Fix a couple typos

nodemadic:
Unsure if function should be _keccak() or keccak() but it should be consistent


gzeoneth:
@cla-bot check

PR #2233, opened 2022-05-13
Extra checks for machine thread abort

joshuacolvin0:
* Add extra checks to detect when machine thread was aborted
* Parameterize yield-instruction-count


hkalodner (approved):
LGTM

PR #2235, opened 2022-05-16
Add “depositAndCall” functionality

gzeoneth:
- [x] Add “depositAndCall” functionality
- [x] Enable affordance for gateway to signal "depositAndCall" support
- [ ] Construct test cases for the new feature

### DIFF
--- a/packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ArbitrumGateway.sol
+++ b/packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ArbitrumGateway.sol
@@ -225,6 +225,16 @@ abstract contract L2ArbitrumGateway is L2ArbitrumMessenger, TokenGateway, Escrow
         IArbToken(_l2Address).bridgeMint(_dest, _amount);
     }
 
+    function finalizeInboundTransfer(
+        address _token,
+        address _from,
+        address _to,
+        uint256 _amount,
+        bytes calldata _data
+    ) external payable override {
+        finalizeInboundTransferAndCall(_token, _from, _to, _amount, address(0), 0, _data);
+    }
+
     /**
      * @notice Mint on L2 upon L1 deposit.
      * If token not yet deployed and symbol/name/decimal data is included, deploys StandardArbERC20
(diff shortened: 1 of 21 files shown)


packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ArbitrumGateway.sol
```diff
@@ -243,14 +251,10 @@ abstract contract L2ArbitrumGateway is L2ArbitrumMessenger, TokenGateway {
...
+        (bytes memory gatewayData, bytes memory extraData) = GatewayMessageHandler
             .parseFromL1GatewayMsg(_data);
-
-        if (callHookData.length != 0) {
-            // callHookData should always be 0 since inboundEscrowAndCall is disabled
-            callHookData = bytes("");
-        }
-
+        // user encoded
+        (uint256 callHookGas, bytes memory callHookData) = abi.decode(extraData, (uint256, bytes));
```
fredlacs, line 257: Let's instead handle this in the GatewayMessageHandler

fredlacs (changes requested):
I think we should also add a pure function in the L1/L2 that signals that the gateway pair can handle the depositAndCall functionality (could be a pure function or erc165 style).
With this we can then make deposits revert on the L1 router if they have depositAndCall data, but are trying to use a token without that functionality

packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ArbitrumGateway.sol
gzeoneth: done in d8c24ac

gzeoneth:
I think now the code should eliminated the griefing factors by
1. Ensure at some pt of execution we have at least callHookData.gas
2. Ensure the refund can be triggered by increasing gas limit

Actually we don't really need to check `gasleft() > callHookData.gas` if we assume there will not be gas dependent (not oog) code in the call hook, if someone execute with a low gas limit it will oog and revert and the retryable ticket can be executed again.

We might need something to prevent user setting callHookData.gas too large that they cannot pay for it on L2, but I don't want to do decode on L1 due to gas cost. We can use uint24 so they cannot set it to more than 16m gas and with encodePacked we can save some calldata gas packing with the address field, or we can hard code the limit L2. 

Also note that currently `callHookData.gas` is not changed on L1 tradable exit.

fredlacs:
the above avoid a potential error of the abi encoding of the user data failing.
we instead pass the refundAddress and maxGas as parameters to a new entrypoint function

Jason-Wanxt:
We should encode token's address to callHookData because in standard transferAndCall, the target contract can call msg.sender to get the token's address, but for our implement, it's hard for the target contract to get that address.

gzeoneth:
> We should encode token's address to callHookData because in standard transferAndCall, the target contract can call msg.sender to get the token's address, but for our implement, it's hard for the target contract to get that address.

That's true but I think it might be better to do in tooling?

fredlacs:
I think it's best if we add it to the onTokenTransfer function signature instead

Jason-Wanxt:
Yea, but if it will be better if we add a check on inboundEscrowAndCall to avoid user encode a wrong l2 token address for any reasons?

fredlacs:
should we forward msg.value on the `onTokenTransfer`

gzeoneth:
> should we forward msg.value on the `onTokenTransfer`

there shouldn't be any callvalue as is
https://github.com/OffchainLabs/arbitrum/blob/03bb55b5c1936d0838baa639b020d291230beb9c/packages/arb-bridge-peripherals/contracts/tokenbridge/ethereum/gateway/L1ArbitrumGateway.sol#L206

fredlacs:
we should also assess issues brought up in https://github.com/OffchainLabs/arbitrum/issues/2255

PR #2241, opened 2022-05-18
Log L1 error after 10 warnings printed

joshuacolvin0:
* First 10 log messages are warning level when L1 unable to be contacted.  Log messages after the first 10 are logged as error
* Normalize a few log calls that were not using per-project custom loggers

### DIFF
--- a/packages/arb-util/ethutils/client.go
+++ b/packages/arb-util/ethutils/client.go
@@ -19,7 +19,6 @@ package ethutils
 import (
 	"context"
 	"encoding/json"
-	"github.com/rs/zerolog/log"
 	"math/big"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


packages/arb-util/ethutils/client.go
```diff
@@ -108,7 +113,18 @@ func (r *RPCEthClient) handleCallErr(err error) error {
 
 	// If we've had above a threshold number of errors, reinitialize the connection
 	if totalErrCount >= maxErrCount {
-		log.Warn().Err(err).Str("url", r.url).Msg("Reconnecting to client endpoint after repeated errors")
+		var partialLog *zerolog.Event
+		if totalErrCount < errorCountError {
```
hkalodner, line 117: I don't think `totalErrCount` should ever reach this size because `r.reconnect` resets it. Alternatively I think we could just always make this an error. I'd guess the false positive rate would be pretty low

hkalodner (changes requested):

packages/arb-util/ethutils/client.go
joshuacolvin0: removed.  Fixed actual problem instead

joshuacolvin0: It is too noisy to always have it as an error.

hkalodner (approved):
LGTM

PR #2243, opened 2022-05-21
A couple minor issues found by semgrep

joshuacolvin0:
* Unlock semaphore when exiting on error
* Do not use output when error flagged
* Properly check for nil value before converting with `Hex()`


hkalodner (approved):
LGTM

PR #2135, opened 2022-03-24
Add callvalue check to inbox

gzeoneth:

PR #2262, opened 2022-05-30
feat: add callvalue check to Inbox.depositEth

gzeoneth:
Add callvalue check to Inbox.depositEth on top of https://github.com/OffchainLabs/arbitrum/pull/2135


gzeoneth:
Got a user hitting this issue https://etherscan.io/tx/0xd84d57af2f3811a4fb8b1790759b090bde6cdad1e36eedc05cdb58bd009bc8a0.
Not sure it worth to add this check because the user would have to deposit a very small and insignificant amount to be affected, and this check add little gas cost to every deposit.

fredlacs:
we can handle this in the UI instead

PR #2266, opened 2022-06-02
Bump go-ethereum, bump broken deps, and fix api changes

hkalodner:
- Bump go-ethereum from v1.10.13 to v1.10.18
- Update some deps to use forks with go-ethereum dependency bumped in order to fix compatibility issue
- Change code that depended on removed go-ethereum compiler API


joshuacolvin0 (approved):
LGTM

PR #2268, opened 2022-06-02
Fix block tracing

hkalodner:
A bug was discovered where block tracing was broken in blocks containing transactions that had no traces. The update will fix the results of the `arbtrace_block` and `arbtrace_replayBlockTransactions` RPC calls in that case


joshuacolvin0 (approved):
LGTM

PR #2288, opened 2022-06-09
Always ignore the last `delay blocks` number of L1 blocks

joshuacolvin0:
This update fixes a corner case where latest L1 blocks were not ignored


PlasmaPower (approved):
LGTM

PR #2314, opened 2022-06-24
Fix race conditions when adding messages to arbcore

joshuacolvin0:
Sequencer could be adding messages at the same time as inboxreader, so needed to account for that.


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -115,7 +115,13 @@ bool ArbCore::machineIdle() {
 ArbCore::message_status_enum ArbCore::messagesStatus() {
     auto current_status = message_data_status.load();
     if (current_status == MESSAGES_SUCCESS) {
-        message_data_status = MESSAGES_EMPTY;
+        auto success = message_data_status.compare_exchange_strong(
+            current_status, MESSAGES_EMPTY);
+        if (!success) {
+            // Status was not updated because value changed, so return current
+            // value
+            current_status = message_data_status.load();
```
PlasmaPower, line 123: In theory, maybe the outer if statement should be a while loop? In practice it's unlikely to happen multiple times though

packages/arb-util/core/lookup.go
```diff
@@ -163,7 +164,7 @@ func waitForMessages(ctx context.Context, db ArbCoreInbox) (MessageStatus, error
 		if status == MessagesEmpty {
 			return 0, errors.New("should have messages")
 		}
-		if status != MessagesReady {
```
PlasmaPower, line 166: Do we need to handle the MessagesError case here?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -181,11 +187,47 @@ bool ArbCore::deliverMessages(
...
-    auto status = message_data_status.load();
-    if (status != MESSAGES_EMPTY) {
-        std::cerr << "unable to deliver messages, status: " << status
-                  << std::endl;
-        return false;
+    uint32_t error_count = 0;
+    auto status_before = MESSAGES_ERROR;
+    while (true) {
+        status_before = message_data_status.load();
+        if (status_before != MESSAGES_EMPTY) {
```
PlasmaPower, line 194: I'd recommend also allowing `MESSAGES_SUCCESS` here

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -181,11 +187,47 @@ bool ArbCore::deliverMessages(
...
+            auto success = message_data_status.compare_exchange_weak(
+                status_before, MESSAGES_LOADING);
+            if (success) {
+                // Successfully changed status
+                status_before = message_data_status.load();
+                break;
+            }
+
+            // Compare_exchange_weak may give false negative
+            if (message_data_status.load() != MESSAGES_EMPTY) {
```
PlasmaPower, line 207: If `status_before` could be multiple things, we should be comparing `message_data_status.load() != status_before` instead.

packages/arb-util/core/lookup.go
joshuacolvin0: `MessagesStatus` always returns an `error` object when there is an error, so it is unneccesary.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: If we did that, then then `messagesStatus` would never return success.  Each call to `DeliverMessages` needs to be followed with a call to `messagesStatus` to make sure no threads get stuck.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: good call, missed that in one of the refactors.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: It is already in a loop on the Go side

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Added MESSAGES_SUCCESS here as well as add code to `messagesStatus` to assuming MESSAGES_SUCCESS if MESSAGES_EMPTY is seen.  That breaks the requirement of DeliverMessages/messageStatus pairing, and makes things more asynchronous.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -181,11 +176,47 @@ bool ArbCore::deliverMessages(
...
+    uint32_t error_count = 0;
+    auto status_before = MESSAGES_ERROR;
+    while (true) {
+        status_before = message_data_status.load();
+        if (status_before != MESSAGES_EMPTY) {
+            std::cerr << "unable to deliver messages, status: " << status_before
+                      << ", error count: " << error_count << std::endl;
+        } else {
+            auto success = message_data_status.compare_exchange_weak(
+                status_before, MESSAGES_LOADING);
```
PlasmaPower, line 188: Let's use `compare_exchange_strong` here. This isn't performance critical in the larger scope of things, and I wouldn't want to spuriously increment `error_count`.

packages/arb-avm-cpp/tests/arbcore.cpp
```diff
@@ -58,16 +58,15 @@ void waitForDelivery(std::shared_ptr<ArbCore>& arbCore) {
     ArbCore::message_status_enum status;
     while (true) {
         status = arbCore->messagesStatus();
-        if (status != ArbCore::MESSAGES_EMPTY &&
-            status != ArbCore::MESSAGES_READY) {
+        if (status != ArbCore::MESSAGES_READY) {
             break;
```
PlasmaPower, line 62: Won't this return too early on `MESSAGES_LOADING`? I think we should specifically check that `status == ArbCore::MESSAGES_EMPTY`.

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: :+1:

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1532,26 +1572,21 @@ bool ArbCore::threadBody(ThreadDataStruct& thread_data) {
...
-                        coreConfig.add_messages_max_failure_count) {
-                    message_data_error_string = error_string;
-                    message_data_status = MESSAGES_ERROR;
-                    std::cerr << "Exiting after arbCore addMessages failed "
-                              << thread_data.add_messages_failure_count
-                              << " times: " << message_data_error_string
-                              << "\n";
+                message_data_error_string = error_string;
+                message_data_status.store(MESSAGES_ERROR);
+                std::cerr << "Exiting after arbCore addMessages failed "
```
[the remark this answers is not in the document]
joshuacolvin0: MESSAGES_ERROR should always be treated as fatal.

packages/arb-avm-cpp/tests/arbcore.cpp
joshuacolvin0: :+1: 
Also check for MESSAGES_ERROR

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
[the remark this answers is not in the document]
PlasmaPower: Previously we retried addMessages a couple times before setting MESSAGES_ERROR though. That might be the right way to do it, I'm just checking that we can't recover from an addMessages failure.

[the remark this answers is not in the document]
joshuacolvin0: previously the retry was only in the arbcore thread.  The retry was moved further down in the stack to ensure retry applied to arbcore as well as the sequencer.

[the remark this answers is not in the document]
joshuacolvin0: I was mistaken, retry loop should not have been removed.  I have restored retry loop

PlasmaPower (approved):
LGTM

PR #2315, opened 2022-06-24
Fix message delivery context cancellation

PlasmaPower:
- Don't cancel message delivery from the sequencer if a single user's transaction submission got canceled
- Allow for message delivery to recover from `waitForMessages` returning early due to a context cancellation (though this circumstance still isn't ideal as it might temporarily error)


joshuacolvin0 (approved):
LGTM

PlasmaPower:
Replaced by #2314

PR #2336, opened 2022-07-05
Update contract interfaces

gzeoneth:
This PR
- [x] Add unsafeCreateRetryableTicket to Inbox interface
- [x] Change interfaces pragma to `>=0.6.9 <0.9.0`
- [x] Bump arb-bridge-eth to 0.7.8
- [x] Bump arb-bridge-peripherals to 1.0.11
- [x] Add CI test for solc `0.6.9` `0.7.0` `0.8.15`

There shouldn't be any production code change except the metadata hash.


packages/arb-bridge-eth/contracts/bridge/interfaces/IBridge.sol
```diff
@@ -16,7 +16,8 @@
  * limitations under the License.
  */
 
-pragma solidity ^0.6.11;
+// solhint-disable-next-line compiler-version
+pragma solidity >=0.4.21 <0.9.0;
```
fredlacs, line 20: won't work on `0.4.x` since the `calldata` location wasn't available in function params

as in 
```solidity
function executeCall(
        address destAddr,
        uint256 amount,
        bytes calldata data
    ) external returns (bool success, bytes memory returnData);``` 
    
    ```suggestion
pragma solidity >=0.5.0 <0.9.0;
```

fredlacs (changes requested):
would be nice to add a test to be sure the interfaces compile in the intended versions as done in https://github.com/OffchainLabs/nitro/pull/758

packages/arb-bridge-eth/contracts/bridge/interfaces/IBridge.sol
gzeoneth: Actually it's only available since 0.6.9
https://github.com/ethereum/solidity/releases/tag/v0.6.9

fredlacs: for all variables, but before that you had data location for specific variables (I think only variable length ones)
https://docs.soliditylang.org/en/v0.5.3/types.html#data-location-and-assignment-behaviour

packages/arb-bridge-eth/contracts/test_only/Solc8Tester.sol
```diff
@@ -0,0 +1,42 @@
...
+import "../rollup/INodeFactory.sol";
+import "../rollup/IRollupCore.sol";
+import "../rollup/facets/IRollupFacets.sol";
+import "../validator/IGasRefunder.sol";
+
+import "arb-bridge-peripherals/contracts/tokenbridge/arbitrum/IArbToken.sol";
+import "arb-bridge-peripherals/contracts/tokenbridge/ethereum/ICustomToken.sol";
+import "arb-bridge-peripherals/contracts/tokenbridge/libraries/IWETH9.sol";
+import "arb-bridge-peripherals/contracts/tokenbridge/libraries/gateway/ICustomGateway.sol";
+import "arb-bridge-peripherals/contracts/tokenbridge/libraries/gateway/ITokenGateway.sol";
```
fredlacs, line 42: `arb-bridge-eth` doesn't have peripherals as a dependency.
we could have this file live in `arb-bridge-peripherals/contracts/test` instead

fredlacs (changes requested):
lets add a `build:all` or `test:compatibility` function to the package.json similar to the nitro PR. with that it should be easy to have CI check version compatibility for us

fredlacs (dismissed):
LGTM

PR #2338, opened 2022-07-06
Based on required condition, "This quantity will be 7/8 when G = 720S…

luxebeng:
… (the maximum gas pool size) and it will be 9/8 when G = 0", Then the formula should be "(3240S - G) / (2880S)".


luxebeng:
the CLA document is updated.

PR #2344, opened 2022-07-10
Add a new --core.test.reorg-messages to reorg messages on startup

PlasmaPower:
This should be used in conjunction with the `--core.reorg-to.*` family of flags to specify where to reorg to


joshuacolvin0 (approved):
LGTM

PR #2345, opened 2022-07-10
Allow inbox reader to reorg when it's past the current height

PlasmaPower:
This matter when the chain reorgs to a much lower block height.


joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

PR #2346, opened 2022-07-10
Don't reconnect to L1 because a receipt wasn't found

PlasmaPower:
This is expected behavior of WaitForRecipt's usage of the TransactionReceipt method


joshuacolvin0 (approved):
LGTM

PR #2347, opened 2022-07-11
don't bother reconnecting if item just not found

joshuacolvin0:


PlasmaPower:
Where else does this error come from?

PR #2351, opened 2022-07-13
Fix and Allowlist audit-ci advisories

gzeoneth:
This PR:

- [x] Pinned `hardhat` to 2.9.9
- [x] Allowlisted the following advisory related to OZ initializer functions.

1. [1067409](https://www.npmjs.com/advisories/1067409): Improper Initialization in OpenZeppelin
2. [1067487](https://www.npmjs.com/advisories/1067487): Initializer reentrancy may lead to double initialization
3. [1067486](https://www.npmjs.com/advisories/1067486): Initializer reentrancy may lead to double initialization

We should be good as long as our contracts does not make external call to unknown contracts in the initializer.


fredlacs (approved):
LGTM

PR #2356, opened 2022-07-15
revert: L1 custom gateway lazy mint (PR #1713)

gzeoneth:
#1713

### DIFF
--- a/packages/arb-bridge-peripherals/test/customGateway.e2e.ts
+++ b/packages/arb-bridge-peripherals/test/customGateway.e2e.ts
@@ -307,7 +307,7 @@ describe('Bridge peripherals end-to-end custom gateway', () => {
     )
   })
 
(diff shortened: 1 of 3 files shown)


packages/arb-bridge-peripherals/test/customGateway.e2e.ts
```diff
@@ -307,7 +307,7 @@ describe('Bridge peripherals end-to-end custom gateway', () => {
     )
   })
 
-  it('should withdraw tokens when minted in L2', async function () {
+  it.skip('should withdraw tokens when minted in L2', async function () {
```
fredlacs, line 310: lets update this test so it expects that the L1 side reverts when trying to withdraw tokens minted in L2

fredlacs (changes requested):

fredlacs (approved):
LGTM

PR #2375, opened 2022-07-22
export to nitro: fix offbyone

tsahee:


PlasmaPower (approved):
LGTM. Note that I pushed up a commit to subtract one from the ExportHistoryStatus return value.

PR #2400, opened 2022-07-27
Nitro init docs

tsahee:


joshuacolvin0:
out of date, current version in nitro docs

PR #2411, opened 2022-07-29
Default rinkeby to not connect to sequencer

joshuacolvin0:
The following settings have new defaults for rinkeby testnet because classic does not connect to nitro sequencer
* `--feed.input.url= --node.forwarder.rpc-mode=non-mutating`

Setting `feed.input.url` to nothing disables connecting to sequencer feed. Setting `node.forwarder.rpc-mode` to `non-mutating` disables connecting to the sequencer RPC.


packages/arb-rpc-node/aggregator/aggregator.go
```diff
@@ -161,22 +165,35 @@ func (m *Server) LatestSnapshot(ctx context.Context) (*snapshot.Snapshot, error)
...
+			return nil, err
+		}
+		if pending == nil {
+			return m.LatestSnapshot(ctx)
+		}
+		return pending, nil
 	}
-	return pending, nil
+
+	return nil, errors.New("no batcher defined, cannot generate pending snapshot")
```
PlasmaPower, line 179: We should return `m.LatestSnapshot(ctx)` in this case

PlasmaPower (commented):
Small comment but generally LGTM

packages/arb-rpc-node/aggregator/aggregator.go
joshuacolvin0: :+1:

PlasmaPower (approved):
LGTM

PR #2420, opened 2022-08-01
Mark TokenGateway.onlyCounterpartGateway as abstract

gzeoneth:
The current virtual function doesn’t make much sense because the lack of aliasing (for L2) or outbox checking (for L1), making it abstract and expect implementation in child contract. 

Also removed a deprecated check in L2GatewayRouter.


fredlacs (approved):
LGTM

PR #2424, opened 2022-08-01
harden nitro migrator

fredlacs:


packages/arb-bridge-eth/contracts/bridge/NitroMigrator.sol
```diff
@@ -162,6 +119,9 @@ contract NitroMigrator is OwnableUpgradeable, IMessageProvider {
         // this returns a different magic value so we can differentiate the user and admin facets
         require(_rollup.isNitroReady() == 0xa4b2, "ADMIN_ROLLUP_NOT_NITRO_READY");
 
+        UpgradeableBeacon beacon = _rollup.nodeFactory().beacon();
+        require(Node(address(beacon)).isNitroReady() == 0xa4b3, "NODE_BEACON_OLD");
```
PlasmaPower, line 123: This (and checks below) should be using the `NitroReadyMagicNums` constants

fredlacs: oops, pushed

packages/arb-bridge-eth/contracts/rollup/NodeFactory.sol
```diff
@@ -25,12 +25,13 @@ import "@openzeppelin/contracts/proxy/BeaconProxy.sol";
...
-    UpgradeableBeacon public beacon;
+    address public override beacon;
 
     constructor() public {
         address templateContract = address(new Node());
-        beacon = new UpgradeableBeacon(templateContract);
-        beacon.transferOwnership(msg.sender);
+        UpgradeableBeacon _beacon = new UpgradeableBeacon(templateContract);
+        _beacon.transferOwnership(msg.sender);
+        beacon = address(_beacon);
```
PlasmaPower, line 34: To avoid this (we'd need a postUpgradeInit too) we could just check if the first node is upgraded instead of explicitly checking if the beacon is upgraded

PlasmaPower (approved):
LGTM

PR #2466, opened 2022-08-19
DIsable feed if empty URL

joshuacolvin0:
Setting `--feed.input.url=""` to an empty string on command line creates an empty list.  If set using environment variable, creates a list with a single entry of an empty string, so treat that as disabling feed as well.


hkalodner (approved):
LGTM

PR #2495, opened 2022-09-05
Arb legacy sender

tsahee:
Removed arb-export executable because it's not used and not working.

Added some re-reading tests while exporting.


tsahee:
This uses geth PR https://github.com/OffchainLabs/go-ethereum/pull/160
It does not use up-to-date geth/master because that seems like an unnecessary upgrade right now.

PlasmaPower (approved):
LGTM

PR #2506, opened 2022-09-27
Better defaults when reached end of classic blocks

joshuacolvin0:
* Force snapshot save to database when final classic block is reached
* Default to no pruning on arbitrum one and Rinkeby testnet since non-archive classic node isn't useful
* Cleanup some log messages


packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -131,6 +106,37 @@ bool ArbCore::checkError() {
     return core_error;
 }
 
+void ArbCore::printMachineOutputInfo(const std::string& msg,
+                                     MachineOutput& machine_output) {
+    std::string l2_block_string;
+    if (machine_output.l2_block_number > coreConfig.final_block) {
+        l2_block_string = "TERMINAL_BLOCK";
```
PlasmaPower, line 113: Won't this always print TERMINAL_BLOCK if final_block is the default of zero?

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
```diff
@@ -1717,7 +1743,8 @@ bool ArbCore::threadBody(ThreadDataStruct& thread_data) {
                     std::chrono::seconds(coreConfig.database_save_interval);
             }
 
-            if (trigger_save_rocksdb_checkpoint ||
+            if (output.l2_block_number >= coreConfig.final_block ||
```
PlasmaPower, line 1746: Same here with the default being zero

PlasmaPower (commented):
Generally LGTM but I think the final_block default of zero causes some issues

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Fixed

packages/arb-avm-cpp/data_storage/src/arbcore.cpp
joshuacolvin0: Fixed

PlasmaPower (approved):
LGTM

PR #2514, opened 2022-10-07
Only update prune point when maybe pruning

hkalodner:
This removes the need for a large distance log query on boot which delays the classic node coming up. It also fixes what I think was a minor bug, not updating the prune point when the pruning mode was defaulted


joshuacolvin0 (approved):
LGTM

PR #2526, opened 2022-10-24
Updating Alchemy's Node Provider description.

deric-alchemy:
Improving Alchemy's node provider description to be updated and accurate.


joshuacolvin0:
docs now in Nitro repo

PR #2551, opened 2022-12-14
Disable classic relay and prevent node from attempting to connect to feed

joshuacolvin0:


packages/arb-node-core/cmd/arb-relay/arb-relay.go
```diff
@@ -70,9 +70,18 @@ func main() {
 	// Print line number that log was created on
 	logger = arblog.Logger.With().Str("component", "arb-relay").Logger()
 
-	if err := startup(); err != nil {
-		logger.Error().Err(err).Msg("Error running relay")
+	logger.Info().Msg("Classic Arbitrum sequencer no longer running, so feed is nonexistent")
+	logger.Info().Msg("Sleeping forever")
+	for {
+		// Sleep forever
+		select {}
```
PlasmaPower, line 77: I think we'll want not to hot loop here, so maybe throw in a sleep.

PlasmaPower: Nevermind, I'm wrong :) `select {}` halts.

PlasmaPower (approved):

